PR #5, opened 2023-03-07
DEV: Preparation work for multiple inference providers

xfalcox:


plugin.rb
```diff
@@ -16,7 +16,13 @@ module ::DiscourseAI
     PLUGIN_NAME = "discourse-ai"
   end
 
-  require_relative "lib/shared/inference_manager"
+  require_relative "app/models/classification_result"
```
romanrizzi, line 19: This one can be removed. It's automatically loaded by the engine.

romanrizzi (approved):

PR #8, opened 2023-03-14
FEATURE: Composer AI helper

romanrizzi:
This change introduces a new composer button for the group members listed in the `ai_helper_allowed_groups` site setting.

Users can use chatGPT to review, improve, or translate their posts to English.

### DIFF
--- a/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
+++ b/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module AiHelper
+    class AssistantController < ::ApplicationController
+      requires_plugin ::DiscourseAi::PLUGIN_NAME
+      requires_login
+      before_action :ensure_can_request_suggestions
+
+      def suggest
+        raise Discourse::InvalidParameters.new(:text) if params[:text].blank?
+
+        if !DiscourseAi::AiHelper::OpenAiPrompt::VALID_TYPES.include?(params[:mode])
+          raise Discourse::InvalidParameters.new(:mode)
+        end
+
+        RateLimiter.new(current_user, "ai_assistant", 6, 3.minutes).performed!
+
+        hijack do
+          response = {
+            suggestions:
+              DiscourseAi::AiHelper::OpenAiPrompt.new.generate_and_send_prompt(
+                params[:mode],
+                params[:text],
+              ),
+          }
+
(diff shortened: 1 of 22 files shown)


app/controllers/discourse_ai/ai_helper/assistant_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      before_action :ensure_can_request_suggestions
+
+      def suggest
+        raise Discourse::InvalidParameters.new(:text) if params[:text].blank?
+
+        if !DiscourseAi::AiHelper::OpenAiPrompt::VALID_TYPES.include?(params[:mode])
+          raise Discourse::InvalidParameters.new(:mode)
+        end
+
+        RateLimiter.new(current_user, "ai_assistant", 6, 3.minutes).performed!
```
xfalcox, line 17: Maybe bypass for staff? Specially now that everyone will be playing with it

app/controllers/discourse_ai/ai_helper/assistant_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        hijack do
+          response = {
+            suggestions:
+              DiscourseAi::AiHelper::OpenAiPrompt.new.generate_and_send_prompt(
+                params[:mode],
+                params[:text],
+              ),
+          }
+
+          if params[:mode] === DiscourseAi::AiHelper::OpenAiPrompt::PROOFREAD
```
xfalcox, line 28: We should move this away from the controller and into lib next.

xfalcox:
Looks great! Let's merge and let people test.

Next, I want to make the list of prompts completely data driven, so we can add / remove prompts from the database instead of hard coding it across various files. That will allow us to iterate quickly on the obscene amount of prompts suggestions we will receive.

config/settings.yml
```diff
@@ -87,3 +87,14 @@ plugins:
 
   ai_openai_api_key:
     default: ""
+
+  composer_ai_helper_enabled:
```
xfalcox, line 91: We need a `composer_ai_allowed_in_pm: false` and the logic for it before merging.

xfalcox (changes requested):
Only disabling it from PMs is a blocker, others are follow ups

app/controllers/discourse_ai/ai_helper/assistant_controller.rb
romanrizzi: Staff is bypassed by default unless specified otherwise

xfalcox (approved):

PR #17, opened 2023-03-20
FIX: Remove old seeded prompts.

romanrizzi:
We switched to negative integers for seeded prompts, but we also need to delete the old ones using IDs 1/2/3 if they exist,
 or deploys will fail due to the unique index on the name column.


db/post_migrate/20230320122645_delete_duplicated_seeded_prompts.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class DeleteDuplicatedSeededPrompts < ActiveRecord::Migration[7.0]
```
ZogStriP, line 3: Is a post migration the right place to do this? 🤔 Aren't the seeds done _way_ earlier?

romanrizzi: Ah, didn't know that. Will move it into a migration then.

xfalcox (approved):

PR #21, opened 2023-03-20
FEATURE: Prompts can consist of multiple messages.

romanrizzi:
A prompt with multiple messages leads to better results, as the AI can learn for given examples. Alongside this change, we provide a better default proofreading prompt.

### DIFF
--- a/app/models/completion_prompt.rb
+++ b/app/models/completion_prompt.rb
@@ -1,7 +1,38 @@
 # frozen_string_literal: true
 
 class CompletionPrompt < ActiveRecord::Base
+  # TODO(roman): Remove sept 2023.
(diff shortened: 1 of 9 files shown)


xfalcox:
Not sold on the new model. Why can't it just be an array on the original model to store the prompt? Is there a world where those will be used in a separate way to we to need a whole new table?  If we need multiple attributes per message, using a jsonb column would be enough.

romanrizzi:
Fair enough, will switch to a jsonb 👍

romanrizzi:
@xfalcox - Switched to jsonb 👍

xfalcox (approved):

PR #28, opened 2023-03-27
FEATURE: attempt to include related topics above suggested

SamSaffron:
This is an attempt at including related above suggested for logged on users

TODO

 - tests
 - ensure this is zero work if feature is not enabled
 - perhaps get rid of ai_embeddings_semantic_suggested_topics_anon_enabled and use a diff setting for both
 - have this disable random suggested, so we don't double up when logged in


SamSaffron:
tests will fail till https://github.com/discourse/discourse/pull/20893 is merged...

oblakeerickson (approved):

PR #34, opened 2023-03-31
FEATURE: Semantic asymetric full-page search

romanrizzi:
Depends on discourse/discourse#20915

Hooks to the full-page-search component using an experimental API and performs an assymetric similarity search using our embeddings database.


lib/modules/embeddings/topic.rb
```diff
@@ -3,54 +3,80 @@
...
-            <<~SQL,
-              INSERT INTO topic_embeddings_#{model.underscore} (topic_id, embedding)
-              VALUES (:topic_id, '[:embedding]')
-              ON CONFLICT (topic_id)
-              DO UPDATE SET embedding = '[:embedding]'
-            SQL
-            topic_id: @topic.id,
-            embedding: model_embedding,
-          )
+      def symmetric_semantic_search(model, topic)
```
xfalcox, line 19: Where is this used?

romanrizzi: It's used by the `SemanticSuggested` service. I moved the queries there since they only work for topic embeddings.

PR #35, opened 2023-03-31
FIX: Use correct variable and method for embeddings

romanrizzi:


assets/javascripts/initializers/semantic-full-page-search.js
```diff
@@ -32,7 +31,6 @@ function initializeSemanticSearch(api) {
               );
             }
           } else {
-            setTransient("lastSearch", { searchKey, model }, 5);
```
romanrizzi, line 35: It was adding an empty term to the recent searches list. Removed it for now.

xfalcox (approved):

PR #41, opened 2023-04-11
FEATURE: Topic summarization

romanrizzi:
Summarize topics using the TopicView's "summary" filter. The UI is similar to what we do for chat, but we don't allow the user to select a timeframe.


xfalcox (approved):

SamSaffron:
nice one!

PR #59, opened 2023-05-12
FEATURE: hide various api keys in site settings

SamSaffron:
Also... ensures that plugin has a link to more info and proper description


Drenmi (approved):
300px-All_Right_Then,_Keep_Your_Secrets

PR #60, opened 2023-05-12
Work in progress... post embeddings and version

SamSaffron:
Feel free to continue any work here, will pick it up again next week


SamSaffron:
My GPU is melting .... embedding every post TWICE, is pretty darn expensive... we will need to prove value here if we head down this path... 

Plus we could do asymetric embeddings for all posts (for search) and symetric for topic summaries as a possible option 

anyway ... please don't merge this draft....

PR #61, opened 2023-05-13
Refinements to embeddings and tokenizers

xfalcox:


SamSaffron:
Amazing, I was reading and instructor can handle really long  topics, but  perf gets bad at 10k tokens

romanrizzi (approved):

PR #67, opened 2023-05-17
Fixes for embeddings and truncate

xfalcox:


SamSaffron (approved):

SamSaffron:
oh that args thing got me as well when I was testing, thanks for fixing

PR #68, opened 2023-05-18
UX: improve drop down menu for enabled bots

SamSaffron:
Previously we were not using using HeaderPanel for drop down, which caused
it not to properly act like a header panel.

- Not styled right
- Not hidden when other buttons clicked

Etc...

Header is sadly full of legacy so this is somewhat hacky weaving widgets.


SamSaffron:
Before:

image


After:

image

assets/javascripts/discourse/components/ai-bot-header-panel.hbs
```diff
@@ -0,0 +1,18 @@
...
+  <div class="menu-panel drop-down">
+    <div class="panel-body">
+      <div class="panel-body-contents">
+        <div class="sidebar-hamburger-dropdown">
+          {{#each this.botNames as |bot|}}
+            <DButton
+              @class="btn-flat ai-bot-available-bot-content"
+              @translatedTitle={{bot.humanized}}
+              @translatedLabel={{bot.humanized}}
+              @action={{action "composeMessageWithTargetBot" bot.modelName}}
```
jjaffeux, line 11: Just a note that here you can now just do:

```
@action={{this.composeMessageWithTargetBot}}
```

assets/javascripts/discourse/components/ai-bot-header-panel.js
```diff
@@ -0,0 +1,45 @@
+import { action } from "@ember/object";
+import { inject as service } from "@ember/service";
+import Component from "@ember/component";
```
jjaffeux, line 3: would be nice to just use "@glimmer/component" here

assets/javascripts/discourse/components/ai-bot-header-panel.js
```diff
@@ -0,0 +1,45 @@
...
+
+  @action
+  async composeMessageWithTargetBot(target) {
+    this._composeAiBotMessage(target);
+  }
+
+  @action
+  async singleComposeAiBotMessage() {
+    this._composeAiBotMessage(
+      this.siteSettings.ai_bot_enabled_chat_bots.split("|")[0]
```
jjaffeux, line 21: is there a possibility `this.siteSettings.ai_bot_enabled_chat_bots` would be an empty string?

jjaffeux: Also it could probably use `this.enabledBotOptions` you defined later

assets/javascripts/discourse/lib/ai-bot-helper.js
```diff
@@ -0,0 +1,27 @@
+import { ajax } from "discourse/lib/ajax";
+import Composer from "discourse/models/composer";
+import I18n from "I18n";
+
+export async function ComposeAiBotMessage(targetBot, composer, appEvents) {
```
jjaffeux, line 5: Up to you, but we generally don't use this pattern of capitalized names for single usage function, and would more likely go with: `composeAiBot
Message`

assets/javascripts/discourse/components/ai-bot-header-panel.js
```diff
@@ -0,0 +1,45 @@
...
+
+  get enabledBotOptions() {
+    return this.siteSettings.ai_bot_enabled_chat_bots.split("|");
+  }
+
+  get singleBotEnabled() {
+    return this.enabledBotOptions.length === 1;
+  }
+
+  async _composeAiBotMessage(targetBot) {
```
jjaffeux, line 42: If you want a private method you can go with : `#composeAiBotMessage`

jjaffeux (approved):

assets/javascripts/discourse/components/ai-bot-header-panel.js
SamSaffron: would not be rendered, but yeah we should pass that in to the panel ... this specific code though was dead code... removed it

assets/javascripts/discourse/components/ai-bot-header-panel.js
SamSaffron: done :)

assets/javascripts/discourse/components/ai-bot-header-panel.js
SamSaffron: done :)

PR #71, opened 2023-05-19
add magic

SamSaffron:
- FEATURE: do not run ai reply job for small actions
- FEATURE: Teaching our bot about current Discourse instance

### DIFF
--- a/app/models/post_custom_prompt.rb
+++ b/app/models/post_custom_prompt.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class PostCustomPrompt < ActiveRecord::Base
+  belongs_to :post
+end
+
+class ::Post
+  has_one :post_custom_prompt, dependent: :destroy
+end
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -63,6 +63,10 @@ en:
     ai_bot_enabled_chat_bots: "Available models to act as an AI Bot"
     ai_helper_add_ai_pm_to_header: "Display a button in the header to start a PM with a AI Bot"
 
+    ai_stability_api_key: "API key for the stability.ai API"
+    ai_stability_engine: "Image generation ngine to use for the stability.ai API"
+    ai_stability_api_url: "URL for the stability.ai API"
+
 
   reviewables:
       reasons:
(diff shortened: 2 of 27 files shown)


xfalcox (approved):

app/models/post_custom_prompt.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class PostCustomPrompt < ActiveRecord::Base
+  belongs_to :post
+end
+
+class ::Post
+  has_one :post_custom_prompt, dependent: :destroy
```
romanrizzi, line 8: Core extensions should live inside the `EntryPoint`

app/models/post_custom_prompt.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class PostCustomPrompt < ActiveRecord::Base
```
romanrizzi, line 3: I think this is really similar to the existing `CompletionPrompt`model. What do you think about changing that instead of creating a second one?

romanrizzi:
Amazing PR, @SamSaffron! I took the liberty of pushing a refactor to better separate the command from the reply generation logic, hope you don't mind.

Failures are legit because same specs are also failing on main. Will took a look on Monday.

romanrizzi (dismissed):

app/models/post_custom_prompt.rb
SamSaffron: sure will have a look!

SamSaffron:
thanks heaps for the review and fix Roman! looks great

app/models/post_custom_prompt.rb
SamSaffron: I think we should possibly attack this later, it just is much leaner than CompletionPrompt... if we were to add this there it can get messy - we are going to want a polymorphic association and so on... I think the post specific one is a bit easier to reason about.

app/models/post_custom_prompt.rb
SamSaffron: tricky cause I don't think we have a "common" entry point, maybe in `plugin.rb` ?

PR #74, opened 2023-05-22
FIX: results will be nil for invalid queries

SamSaffron:
Previous to this change invalid searches would break the command.


nattsw (approved):
Curious, what's an example of an "invalid search"?

SamSaffron:
GPT likes to invent search terms like: `order:first_posted` and so on, these are stripped out of the search term leading to a blank one. 

Our core api just return nil for any blank search...

PR #75, opened 2023-05-22
FEATURE: improve search prompt

SamSaffron:
- We only support searching public topics - make it clear
- Stop using bug/feature, cause is poisons system - these may not exist
- Add after: and before: which are very handy for bounding search results


xfalcox (approved):

SamSaffron:
test if a flake :( will see if I can fix it later.

PR #76, opened 2023-05-23
FIX: ensure we only attempt embedding once every 15 minutes

SamSaffron:
This also heavily reduced log noise and ensures our exception handling is
more surgical.


romanrizzi (approved):

SamSaffron:
another heisentest... we need to rethink our strategy around mocking claude/openai

PR #80, opened 2023-05-23
FIX: Do not allow PG exceptions to bubble as it can trigger a failover on the main db

xfalcox:


lib/modules/embeddings/topic.rb
```diff
@@ -33,55 +33,73 @@ def symmetric_semantic_search(model, topic)
...
+              INSERT INTO topic_embeddings_#{model.name.underscore} (topic_id, embedding)
+              VALUES (:topic_id, '[:embedding]')
+              ON CONFLICT (topic_id)
+              DO UPDATE SET embedding = '[:embedding]'
+            SQL
+            topic_id: topic.id,
+            embedding: embedding,
+          )
+        rescue PG::Error => e
+          Rails.logger.error(
```
davidtaylorhq, line 99: How will this recover? Do we have an automatic backfill job?

xfalcox: Yes, we retry on next pageview on the same topic automatically.

lib/modules/embeddings/topic.rb
```diff
@@ -33,55 +33,73 @@ def symmetric_semantic_search(model, topic)
...
+                FROM
+                  topic_embeddings_#{model.name.underscore}
+                ORDER BY
+                  embedding #{model.pg_function} '[:query_embedding]'
+                LIMIT :limit
+                OFFSET :offset
+              SQL
+              .map(&:topic_id)
+        rescue PG::Error => e
+          raise StandardError, "No embeddings found for topic #{topic.id}" if candidate_ids.empty?
```
davidtaylorhq, line 52: I think this needs to be a `MissingEmbeddingError` so that it's caught upstream correctly?

https://github.com/discourse/discourse-ai/blob/b213fe7f94000205496ea5243c4801d4b3ebb252/lib/modules/embeddings/semantic_related.rb#L44

Would be great to add a spec which stubs `Database::Connection.db.query` to return a `PG::ConnectionBad`, and then makes an HTTP request. Then we'd have a nice end-to-end test of this handling.

oblakeerickson (approved):

PR #81, opened 2023-05-24
FIX: missing localization

SamSaffron:
blog.start_gpt_chat -> was on my blog

This also slightly tunes the search prompt to support filtering by oldest
and try a tiny bit harder to guide GPT 3.5 which is a bit of a losing battle

### DIFF
--- a/lib/modules/ai_bot/commands/search_command.rb
+++ b/lib/modules/ai_bot/commands/search_command.rb
@@ -13,8 +13,9 @@ module DiscourseAi::AiBot::Commands
 
       def extra_context
         <<~TEXT
-          Discourse search supports, the following special commands:
+          Discourse search supports, the following special filters:
 
+          user:USERNAME: only posts created by a specific user
           in:tagged: has at least 1 tag
           in:untagged: has no tags
           in:title: has the search term in the title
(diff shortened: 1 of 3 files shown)


lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -39,11 +40,15 @@ def extra_context
           -tags:TAG1+TAG2: excluding topics tagged TAG1 and TAG2
           order:latest: order by post creation desc
           order:latest_topic: order by topic creation desc
+          order:oldest : order by post creation asc
```
lis2, line 43:
```suggestion
          order:oldest: order by post creation asc
```

spec/lib/modules/ai_bot/jobs/regular/create_ai_reply_spec.rb
```diff
@@ -20,6 +20,9 @@
       before do
         bot_user = User.find(DiscourseAi::AiBot::EntryPoint::GPT3_5_TURBO_ID)
 
+        # time nees to be frozen so time in prompt does not drift
```
lis2, line 23:
```suggestion
        # time needs to be frozen so time in prompt does not drift
```

lis2 (approved):

SamSaffron:
thanks heaps @lis2

PR #82, opened 2023-05-26
FEATURE: use triage command to attempt to ground models on GPT 3.5

SamSaffron:
Previous to this change our results were ungrounded and simpler models like
GPT 3.5 and claude had a real tough time figuring out how to run commands

This splits responses into 2 phases:

Phase 1: figure out if you need to run a command
Phase 2: respond to user

This seems to produce better results on both Claude and GPT 3.5 but still
needs a fair bit of tuning.

### DIFF
--- a/lib/modules/ai_bot/anthropic_bot.rb
+++ b/lib/modules/ai_bot/anthropic_bot.rb
@@ -49,12 +49,12 @@ module DiscourseAi
         ).dig(:completion)
       end
 
-      def submit_prompt(prompt, prefer_low_cost: false, &blk)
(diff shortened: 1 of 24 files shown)


SamSaffron:
making progress here @romanrizzi / @xfalcox but I think my history technique needs work. 

injecting "context" mid conversation just throws off the entire flow. going to borrow some of the techniques for langchain to see if they help. maybe moving context to the absolute top is the way to go. 

on the upside this approach seems much more doable for GPT 3.5 so that is nice...

SamSaffron:
given the new changes to open ai api this is totally dead in the water and not needed for now

PR #87, opened 2023-06-15
FEATURE: add full bot support for GPT 3.5

SamSaffron:
Given latest GPT 3.5 16k which is both better steered and supports functions
we can now support rich bot integration.

Clunky system message based steering is removed and instead we use the
function framework provided by Open AI

Tests still need fixing and system needs a bit more debugging, but early
results are very promising!

### DIFF
--- a/lib/modules/ai_bot/bot.rb
+++ b/lib/modules/ai_bot/bot.rb
@@ -185,11 +185,11 @@ module DiscourseAi
           conversation.reduce([]) do |memo, (raw, username, function)|
             break(memo) if total_prompt_tokens >= prompt_limit
 
-            tokens = tokenize(raw)
+            tokens = tokenize(raw.to_s)
 
             while !raw.blank? && tokens.length + total_prompt_tokens > prompt_limit
               raw = raw[0..-100] || ""
-              tokens = tokenize(raw)
+              tokens = tokenize(raw.to_s)
             end
 
             next(memo) if raw.blank?
--- a/lib/modules/ai_bot/bot.rb
+++ b/lib/modules/ai_bot/bot.rb
@@ -131,6 +131,12 @@ module DiscourseAi
             skip_validations: true,
             skip_revision: true,
           )
+
+          bot_reply_post.post_custom_prompt ||= post.build_post_custom_prompt(custom_prompt: [])
+          prompt = post.post_custom_prompt.custom_prompt || []
+
+          prompt << build_message(bot_user.username, reply)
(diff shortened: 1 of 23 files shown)


lib/modules/ai_bot/bot.rb
```diff
@@ -3,6 +3,25 @@
...
+        attr_reader :functions
+        attr_reader :current_function
+
+        def initialize
+          @functions = []
+          @current_function = nil
+        end
+
+        def add_function(name)
+          @current_function = { name: name, arguments: +"" }
```
jjaffeux, line 16: What is the `+""` achieving here? I would say nothing given it's an empty string but I might not know something here...

romanrizzi: He will later mutate the string using `<<`

jjaffeux: ok I see

```ruby
irb(main):009:0> +"".freeze << "foo"
=> "foo"
irb(main):010:0> "".freeze << "foo"
(irb):10:in `<main>': can't modify frozen String: "" (FrozenError)
	from /Users/joffreyjaffeux/.asdf/installs/ruby/3.2.1/lib/ruby/gems/3.2.0/gems/irb-1.6.3/exe/irb:9:in `<top (required)>'
	from /Users/joffreyjaffeux/.asdf/installs/ruby/3.2.1/bin/irb:25:in `load'
	from /Users/joffreyjaffeux/.asdf/installs/ruby/3.2.1/bin/irb:25:in `<main>'
```

Well first time I see that syntax

lib/modules/ai_bot/open_ai_bot.rb
```diff
@@ -18,7 +18,7 @@ def prompt_limit
         if bot_user.id == DiscourseAi::AiBot::EntryPoint::GPT4_ID
           8192 - 3500
```
xfalcox, line 19: There is also GPT-4 32k

SamSaffron: we don't have access yet afaik

SamSaffron: plus MEGA expensive.

romanrizzi (approved):
Looks good to merge. We should probably do some cleanup/refactoring in the future to make it easier to follow - it has become quite complex.

SamSaffron:
Thanks Roman, yeah I agree, this is getting quite complex and we are going to have to support functions with other LLMs which is going to introduce even more complexities.

PR #88, opened 2023-06-15
DEV: Better strategies for summarization

romanrizzi:
Depends on discourse/discourse#22230

The strategy responsibility needs to be "Given a collection of texts, I know how to summarize them most efficiently, using the minimum amount of requests and maximizing token usage".

There are different token limits for each model, so it all boils down to two different strategies:

Fold all these texts into a single one, doing the summarization in chunks, and then build a summary from those. Build it by combining texts in a single prompt, and truncate it according to your token limits.

While the latter is less than ideal, we need it for "bart-large-cnn-samsum" and "flan-t5-base-samsum", both with low limits. The rest will rely on folding.


plugin.rb
```diff
@@ -7,8 +7,8 @@
 # url: https://meta.discourse.org/t/discourse-ai/259214
 # required_version: 2.7.0
 
-gem "tokenizers", "0.3.2", platform: RUBY_PLATFORM
-gem "tiktoken_ruby", "0.0.5", platform: RUBY_PLATFORM
+gem "tokenizers", "0.3.2", platform: "arm64-darwin"
```
xfalcox, line 10: :eyes:

xfalcox (approved):

PR #91, opened 2023-06-16
FIX: OpenAI Tokenizer was failing to truncate mid emojis

xfalcox:


spec/shared/tokenizer.rb
```diff
@@ -76,5 +76,10 @@
       sentence = "foo bar baz qux quux corge grault garply waldo fred plugh xyzzy thud"
       expect(described_class.truncate(sentence, 3)).to eq("foo bar baz")
     end
+
+    it "truncates a sentence sucesfully at a multibyte unicode character" do
```
jjaffeux, line 80:
```suggestion
    it "truncates a sentence succesfully at a multibyte unicode character" do
```

jjaffeux (approved):

PR #93, opened 2023-06-21
FEATURE: Add Azure cognitive service support

SamSaffron:
The new site settings:

ai_openai_gpt35_url
ai_openai_gpt4_url
ai_openai_embeddings_url

Can be used to configure OpenAI support on Azure. If untouched we will simply
use OpenAI endpoints.

Note: Azure require 1 URL per model, OpenAI allows a single URL to serve
multiple models.


xfalcox (approved):

SamSaffron:
spec is weird ... not failing on local ... looking

PR #99, opened 2023-07-10
FEATURE: Embeddings to main db

xfalcox:
This commit moves our embeddings store from an external configurable PostgreSQL instance back into the main database. This is done to simplify the setup.

There is a migration that will try to import the external embeddings into the main DB if it is configured and there are rows.

It removes support from embeddings models that aren't all_mpnet_base_v2 or OpenAI text_embedding_ada_002. However it will now be easier to add new models.

It also now takes into account:
  - topic title
  - topic category
  - topic tags
  - replies (as much as the model allows)

We introduce an interface so we can eventually support multiple strategies for handling long topics.

This PR severely damages the semantic search performance, but this is a temporary until we can get adapt HyDE to make semantic search use the same embeddings we have for semantic related with good performance.

Here we also have some ground work to add post level embeddings, but this will be added in a future PR.


lib/modules/embeddings/manager.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+        @strategy = DiscourseAi::Embeddings::Strategies::Truncation.new(@target, @model)
+      end
+
+      def generate!
+        @strategy.process!
+
+        # bail here if we already have an embedding with matching version and digest
+
+        @embeddings = @model.generate_embeddings(processed_target)
```
romanrizzi, line 22: I assume this should be `@strategy.processed_target`?

romanrizzi (commented):
Looks good to me. I left a comment for a variable not defined. If it wasn’t working, we should probably add a test for it.

romanrizzi (approved):

PR #103, opened 2023-07-13
REFACTOR: Cohesive narrative for single-chunk summaries.

romanrizzi:
Single and multi-chunk summaries end up using different prompts for the last summary. This change detects when the summarized content fits in a single chunk and uses a slightly different prompt, which leads to more consistent summary formats.

This PR also moves the chunk-splitting step to the `FoldContent` strategy as preparation for implementing streamed summaries.

### DIFF
--- a/lib/modules/summarization/models/open_ai.rb
+++ b/lib/modules/summarization/models/open_ai.rb
@@ -37,7 +37,7 @@ module DiscourseAi
           messages = [{ role: "system", content: build_base_prompt(opts) }]
 
           text_to_summarize = contents.map { |c| format_content_item(c) }.join
-          truncated_content = tokenizer.truncate(text_to_summarize, max_tokens - reserved_tokens)
+          truncated_content = tokenizer.truncate(text_to_summarize, available_tokens)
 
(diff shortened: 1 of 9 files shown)


lib/modules/summarization/models/open_ai.rb
```diff
@@ -47,13 +47,24 @@ def summarize_with_truncation(contents, opts)
...
+        end
+
         private
 
         def summarize_chunk(chunk_text, opts)
+          summary_instruction =
+            if opts[:single_chunk]
+              "Summarize the following forum discussion, creating a cohesive narrative:"
+            else
+              "Summarize the following in 400 words:"
```
xfalcox, line 61: Do we need an "up to" or this only happening when there are >1 chunks means 400 is always reasonable?

xfalcox (approved):

lib/modules/summarization/models/open_ai.rb
romanrizzi: I don't think we need it because we re-summarize the 400-word summaries.

PR #105, opened 2023-07-13
FIX: Disable truncation and padding in all-mpnet-base-v2 tokenizer

xfalcox:
The tokenizer was truncating and padding to 128 tokens, and we try append
new post content until we hit 384 tokens. This was causing the tokenizer
to accept all posts in a topic, wasting CPU and memory.


jjaffeux (approved):
I have limited understanding of this so far, but I see the padding/truncation removed and nothing suspicious.

PR #114, opened 2023-07-25
FEATURE: port to use claude-2 for chat bot

SamSaffron:
Claude 1 costs the same and is less good than Claude 2. Make use of Claude
2 in all spots ...

This also fixes streaming so it uses the far more efficient streaming protocol.


SamSaffron:
going to take a bit to get the tests working, but overall it is a way better model to use. 

also thinking about how to enable commands on claude 2, model may be smart enough to handle it.

PR #116, opened 2023-07-26
FEATURE: Llama2 for summarization

xfalcox:
image


lib/shared/inference/hugging_face_text_generation.rb
```diff
@@ -0,0 +1,117 @@
...
+                        nil
+                      end
+                    end
+                  end
+              rescue IOError
+                raise if !cancelled
+              ensure
+                log.update!(
+                  raw_response_payload: response_raw,
+                  request_tokens: DiscourseAi::Tokenizer::AnthropicTokenizer.size(prompt),
```
SamSaffron, line 101: anthropic?

SamSaffron:
adding a chat bot after this is going to be super easy :hugs:

SamSaffron:
Probably also worth adding instructions for devs on how to run locally, would be very cool cause more people can just hack easily on summarization or play with local bots.

romanrizzi (approved):
❤️

config/settings.yml
```diff
@@ -108,6 +108,9 @@ plugins:
     choices:
       - "stable-diffusion-xl-beta-v2-2-2"
       - "stable-diffusion-v1-5"
+  ai_hugging_face_api_url:
```
romanrizzi, line 111: We're missing a translation for this setting.

PR #117, opened 2023-07-28
FEATURE: add support for an open LLM implementation

SamSaffron:
This is very much a work in progress, testing at the moment on:

TheBloke/Llama-2-7B-chat-fp16


xfalcox:
Why you testing on the quantized variant instead of the base one? Speed?

SamSaffron:
lol was waiting on facebook to give me access, but one nice thing now that I can run 13B locally, so that is a cool side effect.

SamSaffron:
given @romanrizzi is rewriting all this stuff might as well close this and revisit.

PR #121, opened 2023-08-01
DEV: Pin plugin for v3.1

romanrizzi:
Changes to the topic recommendations list added by discourse/discourse#22896 were reverted from stable.


.discourse-compatibility
```diff
@@ -1 +1,2 @@
+3.1.0: 3e7c99de894b03e9161d9481da118fe893991a1d
```
davidtaylorhq, line 1: We need to cover future 'minor' releases of stable here. So let's use the new `<` syntax:
```suggestion
< 3.2.0.beta1-dev: 3e7c99de894b03e9161d9481da118fe893991a1d
```

davidtaylorhq (approved):

PR #126, opened 2023-08-03
FEATURE: Add support for StableBeluga and Upstage Llama2 instruct

xfalcox:
This means we support all models in the top3 of the Open LLM Leaderboard

Since some of those models have RoPE, we now have a setting so you can customize the token limit depending which model you use.


lib/modules/summarization/entry_point.rb
```diff
@@ -22,7 +23,8 @@ def inject_into(plugin)
           Models::OpenAi.new("gpt-3.5-turbo-16k", max_tokens: 16_384),
           Models::Discourse.new("long-t5-tglobal-base-16384-book-summary", max_tokens: 16_384),
           Models::Anthropic.new("claude-2", max_tokens: 100_000),
-          Models::Llama2.new("Llama-2-7b-chat-hf", max_tokens: 4096),
+          Models::Llama2.new("Llama2-chat-hf"),
+          Models::Llama2FineTunedOrcaStyle.new("StableBeluga2"),
```
romanrizzi, line 27: Can we pass `ai_hugging_face_token_limit` here? I prefer we don't use `nil` as a default param in the inference class to make it clear it's mandatory

romanrizzi (approved):

lib/modules/summarization/entry_point.rb
xfalcox: done

PR #128, opened 2023-08-08
FIX: Google command was including full payload

SamSaffron:
Additionally there was no truncating happening meaning you could blow token
budget easily on a single search.

This made Google search mostly useless and it would mean that after using
Google we would revert to a clean slate which is very confusing.

### DIFF
--- a/lib/modules/ai_bot/commands/google_command.rb
+++ b/lib/modules/ai_bot/commands/google_command.rb
@@ -48,22 +48,28 @@ module DiscourseAi::AiBot::Commands
         URI("https://www.googleapis.com/customsearch/v1?key=#{api_key}&cx=#{cx}&q=#{query}&num=10")
       body = Net::HTTP.get(uri)
 
-      parse_search_json(body)
+      parse_search_json(body, query)
     end
 
-    def parse_search_json(json_data)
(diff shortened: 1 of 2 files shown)


lib/modules/ai_bot/commands/google_command.rb
```diff
@@ -48,22 +48,28 @@ def process(query:)
...
       body = Net::HTTP.get(uri)
 
-      parse_search_json(body)
+      parse_search_json(body, query)
     end
 
-    def parse_search_json(json_data)
+    def minimize_field(result, field, max_tokens: 100)
+      data = result[field].squish
+      data = ::DiscourseAi::Tokenizer::BertTokenizer.truncate(data, max_tokens).squish if max_tokens
```
martin-brennan, line 56: Why do `if max_tokens` here if `max_tokens` has a default value of 100?

martin-brennan (approved):
Just a minor comment otherwise LGTM :+1:

lib/modules/ai_bot/commands/google_command.rb
SamSaffron: oops leftover from when it was nil :)

PR #131, opened 2023-08-09
FIX: refine image and read command

SamSaffron:
- FIX: refine image and read command
- Add a test for image command


martin-brennan (approved):

SamSaffron:
thanks @martin-brennan :hugs:

PR #133, opened 2023-08-11
FEATURE: basic progress for image generation

SamSaffron:
previously you would have to wait quite a while to see the prompt this implements
a very basic implementation of progress so you can see the api is working.

### DIFF
--- a/lib/modules/ai_bot/commands/command.rb
+++ b/lib/modules/ai_bot/commands/command.rb
@@ -78,18 +78,41 @@ module DiscourseAi
           true
         end
 
+        CARET = "<!-- caret -->"
+        CARET2 = "<!-- caret2 -->"
+
+        def show_progress(text, caret2: false)
+          caret = caret2 ? CARET2 : CARET
+          new_placeholder = @placeholder.sub(caret, text + caret)
+          raw = @current_post.raw.sub(@placeholder, new_placeholder)
+          @placeholder = new_placeholder
+
+          @current_post.revise(bot_user, { raw: raw }, skip_validations: true, skip_revision: true)
+        end
+
+        def localized_description
+          I18n.t(
+            "discourse_ai.ai_bot.command_description.#{self.class.name}",
+            self.description_args,
+          )
+        end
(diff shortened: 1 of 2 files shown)


SamSaffron:
failure is unrelated ...

lib/modules/ai_bot/commands/command.rb
```diff
@@ -15,6 +15,9 @@ def initialize(name:, description:, type:, enum: nil, required: false)
       end
 
       class Command
+        CARET = "<!-- caret -->"
+        CARET2 = "<!-- caret2 -->"
```
martin-brennan, line 19: Maybe something more descriptive like `DETAILS_CARET` and `PROGRESS_CARET`?

lib/modules/ai_bot/commands/command.rb
```diff
@@ -36,11 +39,25 @@ def parameters
...
+          @parent_post = parent_post
+
+          @placeholder = +(<<~HTML).strip
+            <details>
+              <summary>#{I18n.t("discourse_ai.ai_bot.command_summary.#{self.class.name}")}</summary>
+              <p>
+                #{CARET}
+              </p>
+            </details>
+            #{CARET2}
```
martin-brennan, line 57: Maybe we just have some text before the dots like `Generating image, please wait...` then just keep adding dots to the end.

martin-brennan (approved):
Just a minor comment but tried locally and this is definitely an improvement over the status quo. Would be nice to have a fancy animation on this too!

lib/modules/ai_bot/commands/command.rb
SamSaffron: cool ... will rename a bit ... agree caret 2 is somewhat confusing

SamSaffron: generating image ... is a bit of a problem cause one big point of this is so you see the prompt early. 

"Generating: "Amazing robots flying" ... " could work though ... will give it a shot ...

PR #136, opened 2023-08-15
FEATURE: HyDE-powered semantic search.

xfalcox:
It relies on the new outlet added on https://github.com/discourse/discourse/pull/23390 to display semantic search results in an unobtrusive way.

We'll use a HyDE-backed approach for semantic search, which consists on generating an hypothetical document from a given keywords, which gets transformed into a vector and used in a asymmetric similarity topic search.

This PR also reorganizes the internals to have less moving parts, maintaining one hierarchy of DAOish classes for vector-related operations like transformations and querying.

Completions and vectors created by HyDE will remain cached on Redis for now, but we could later use Postgres instead.


config/settings.yml
```diff
@@ -177,6 +177,18 @@ discourse_ai:
   ai_embeddings_semantic_search_enabled:
     default: false
     client: true
+  ai_embeddings_semantic_search_hyde_model:
+    default: "gpt-3.5-turbo"
```
romanrizzi, line 181: I remember this being neccessary for things to work, but I don't know if it's still true after all the changes. Will take a look.

app/controllers/discourse_ai/embeddings/embeddings_controller.rb
```diff
@@ -19,12 +18,15 @@ def search
             use_pg_headlines_for_excerpt: false,
           )
 
-        DiscourseAi::Embeddings::SemanticSearch
-          .new(guardian)
-          .search_for_topics(query, page)
-          .each { |topic_post| grouped_results.add(topic_post) }
+        semantic_search = DiscourseAi::Embeddings::SemanticSearch.new(guardian)
```
xfalcox, line 21: Can we introduce a rate limit here? A shared one for all anons and a per user for logged in?

PR #137, opened 2023-08-16
FEATURE: optional warning attached to all AI bot conversations

SamSaffron:
This commit introduces `ai_bot_enable_chat_warning` which can be used
to warn people prior to starting a chat with the bot.

In particular this is useful if moderators are regularly reading chat
transcripts as it sets expectations early.

By default this is disabled.

Also:

- Stops making ajax call prior to opening composer
- Hides PM title when starting a bot PM

image

Design can be improved but at least this is a start.

### DIFF
--- a/assets/javascripts/discourse/connectors/composer-after-composer-editor/composer-open.hbs
+++ b/assets/javascripts/discourse/connectors/composer-after-composer-editor/composer-open.hbs
@@ -1,0 +1,6 @@
+{{#if this.isGptChat}}
+  <DSection @bodyClass="ai-bot-chat" />
+	{{#if this.renderChatWarning}}
+	<div class='ai-bot-chat-warning'>{{i18n "discourse_ai.ai_bot.pm_warning"}}</div>
+	{{/if}}
+{{/if}}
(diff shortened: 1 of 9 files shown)


assets/javascripts/discourse/connectors/composer-after-composer-editor/composer-open.hbs
```diff
@@ -0,0 +1,8 @@
+{{#if this.isGptChat}}
```
xfalcox, line 1:
```suggestion
{{#if this.isAiBotChat}}
```

assets/javascripts/discourse/connectors/composer-after-composer-editor/composer-open.js
```diff
@@ -0,0 +1,32 @@
...
+  get composerModel() {
+    return this.args.outletArgs.model;
+  }
+
+  get renderChatWarning() {
+    return this.siteSettings.ai_bot_enable_chat_warning;
+  }
+
+  @computed("composerModel.targetRecipients")
+  get isGptChat() {
```
xfalcox, line 22:
```suggestion
  get isAiBotChat() {
```

SamSaffron:
will add a test later today, but I really want to land the warning now, so merging.

PR #160, opened 2023-08-25
FEATURE: add setting_context experimental command

SamSaffron:
This command can be used to extract information about a discourse
site setting directly from source.

To operate it needs the rg binary in the container.

### DIFF
--- a/lib/modules/ai_bot/commands/setting_context_command.rb
+++ b/lib/modules/ai_bot/commands/setting_context_command.rb
@@ -1,0 +1,148 @@
+#frozen_string_literal: true
+
+module DiscourseAi::AiBot::Commands
+  MAX_CONTEXT_TOKENS = 2000
+
+  class SettingContextCommand < Command
+    def self.rg_installed?
+      @rg_installed ||=
+        begin
+          Discourse::Utils.execute_command("which", "rg")
+          true
+        rescue Discourse::Utils::CommandError
+          false
+        end
+    end
+
+    class << self
+      def name
+        "setting_context"
+      end
+
+      def desc
+        "Will provide you with full context regarding a particular site setting in Discourse"
+      end
+
+      def parameters
+        [
+          Parameter.new(
+            name: "setting_name",
+            description: "The name of the site setting we need context for",
+            type: "string",
+            required: true,
+          ),
+        ]
+      end
+    end
+
+    def result_name
+      "context"
+    end
+
+    def description_args
+      { setting_name: @setting_name }
+    end
+
+    def process(setting_name:)
+      if !self.class.rg_installed?
+        return(
+          {
+            setting_name: setting_name,
+            context: "This command requires the rg command line tool to be installed on the server",
+          }
+        )
+      end
+
+      @setting_name = setting_name
+      if !SiteSetting.has_setting?(setting_name)
+        { setting_name: setting_name, context: "This setting does not exist" }
+      else
+        description = SiteSetting.description(setting_name)
+        result = +"# #{setting_name}\n#{description}\n\n"
+
+        setting_info =
+          find_setting_info(setting_name, [Rails.root.join("config", "site_settings.yml").to_s])
+        if !setting_info
+          setting_info =
+            find_setting_info(setting_name, Dir[Rails.root.join("plugins/**/settings.yml")])
+        end
+
+        result << setting_info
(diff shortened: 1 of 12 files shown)


spec/lib/modules/ai_bot/commands/setting_context_command.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+RSpec.describe DiscourseAi::AiBot::Commands::SettingContextCommand do
+  let(:bot_user) { User.find(DiscourseAi::AiBot::EntryPoint::GPT3_5_TURBO_ID) }
+  let(:command) { described_class.new(bot_user: bot_user, args: nil) }
+
+  # tests require that rg is installed and are likely going to fail in CI
```
martin-brennan, line 7: If you want to skip only on CI, you can do something like this in each test instead of xit:

```ruby
skip("rg is needed for these tests") if ENV["CI"]
```

lib/modules/ai_bot/commands/setting_context_command.rb
```diff
@@ -0,0 +1,148 @@
...
+          result << Discourse::Utils.execute_command(
+            "rg",
+            setting_name,
+            path,
+            "-g",
+            "!**/spec/**",
+            "-g",
+            "!**/dist/**",
+            "-g",
+            "*.{rb,js,gjs}",
```
martin-brennan, line 84: Should probably include `.hbs` files here too?

lib/modules/ai_bot/commands/setting_context_command.rb
```diff
@@ -0,0 +1,148 @@
...
+      path, result = nil
+
+      paths.each do |search_path|
+        result =
+          Discourse::Utils.execute_command(
+            "rg",
+            name,
+            search_path,
+            "-g",
+            "*.{rb,js,gjs}",
```
martin-brennan, line 114: Same comment here with .hbs

lib/modules/ai_bot/commands/setting_context_command.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+    end
+
+    class << self
+      def name
+        "setting_context"
+      end
+
+      def desc
+        "Will provide you with full context regarding a particular site setting in Discourse"
```
martin-brennan, line 23: Should this (and description below) be localized with I18n?

lib/modules/ai_bot/commands/setting_context_command.rb
```diff
@@ -0,0 +1,148 @@
...
+          {
+            setting_name: setting_name,
+            context: "This command requires the rg command line tool to be installed on the server",
+          }
+        )
+      end
+
+      @setting_name = setting_name
+      if !SiteSetting.has_setting?(setting_name)
+        { setting_name: setting_name, context: "This setting does not exist" }
```
martin-brennan, line 58: Same comment around localization

lib/modules/ai_bot/commands/setting_context_command.rb
SamSaffron: can we have site settings in hbs files?

lib/modules/ai_bot/commands/setting_context_command.rb
SamSaffron: its a very very tough one, this is instructions for the LLM, not humans, interacting with LLMs not in English is not super strong at the moment when it comes to mega fancy stuff.

spec/lib/modules/ai_bot/commands/setting_context_command.rb
SamSaffron: hmmm that is a good idea

lib/modules/ai_bot/commands/setting_context_command.rb
martin-brennan: Yes we definitely can, client side ones anyway. Eg you can do this.siteSettings.blah in a component

lib/modules/ai_bot/commands/setting_context_command.rb
martin-brennan: Ahh right I didn’t realise it went to the LLM, nvm!

martin-brennan (approved):

PR #161, opened 2023-08-25
DEV: adding a concept of prompt builder

SamSaffron:
This is half baked and not working, but starts demonstrating the idea


SamSaffron:
nothing to merge here, this is all very wip atm.

PR #172, opened 2023-08-29
FEATURE: add initial support for personas

SamSaffron:
This completes a V0 of personas. Personas are ways of adjusting our bot framework
so it behaves differently based on user expectations. 

This ships with 3 personas: 

- General - which is the standard one we use today
- Artist - which is purpose built for creating pictures
- Sql Helper - which is there to help building data explorer queries

We display a picker now to help pick which persona you want and decorate the bot to print the persona

System is very much V0, we are anticipating allowing Discourse admins to create the personas they want.

### DIFF
--- a/assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
+++ b/assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
@@ -1,0 +1,53 @@
+import Component from "@glimmer/component";
+import { inject as service } from "@ember/service";
+
+function isBotMessage(composer, currentUser) {
+  if (
+    composer &&
+    composer.targetRecipients &&
+    currentUser.ai_enabled_chat_bots
+  ) {
+    let reciepients = composer.targetRecipients.split(",");
+
+    return currentUser.ai_enabled_chat_bots.any((bot) =>
+      reciepients.any((username) => username === bot.username)
+    );
+  }
+  return false;
+}
+
+export default class BotSelector extends Component {
+  static shouldRender(args, container) {
+    return (
+      container?.currentUser?.ai_enabled_personas &&
+      isBotMessage(args.model, container.currentUser)
+    );
+  }
+
+  @service currentUser;
+
+  get composer() {
+    return this.args?.outletArgs?.model;
+  }
+
+  get botOptions() {
+    if (this.currentUser.ai_enabled_personas) {
+      return this.currentUser.ai_enabled_personas.map((persona) => {
+        return {
+          id: persona.name,
+          name: persona.name,
+          description: persona.description,
+        };
+      });
+    }
+  }
+
+  get value() {
+    return this._value || this.botOptions[0].id;
+  }
+
+  set value(val) {
+    this._value = val;
+    this.composer.metaData = { ai_persona: val };
+  }
+}
--- a/assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
+++ b/assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
@@ -7,7 +7,7 @@ function isBotMessage(composer, currentUser) {
     composer.targetRecipients &&
     currentUser.ai_enabled_chat_bots
   ) {
-    let reciepients = composer.targetRecipients.split(",");
+    const reciepients = composer.targetRecipients.split(",");
 
     return currentUser.ai_enabled_chat_bots.any((bot) =>
       reciepients.any((username) => username === bot.username)
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -95,8 +95,8 @@ en:
     ai_bot:
       personas:
         general:
-          name: Default
-          description: "General purpose AI Bot capable of performing a flexible amount of tasks"
+          name: Forum Helper
+          description: "General purpose AI Bot capable of performing various tasks"
         artist:
           name: Artist
           description: "AI Bot specialized in generating images"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -93,6 +93,16 @@ en:
         markdown_table: Generate Markdown table
 
     ai_bot:
+      personas:
+        general:
+          name: Default
+          description: "General purpose AI Bot capable of performing a flexible amount of tasks"
+        artist:
+          name: Artist
+          description: "AI Bot specialized in generating images"
+        sql_helper:
+          name: SQL Helper
+          description: "AI Bot specialized in helping craft SQL queries on this Discourse instance"
       default_pm_prefix: "[Untitled AI bot PM]"
       topic_not_found: "Summary unavailable, topic not found!"
       command_summary:
@@ -105,6 +115,7 @@ en:
         google: "Search Google"
         read: "Read topic"
         setting_context: "Look up site setting context"
+        schema: "Look up database schema"
       command_description:
         read: "Reading: <a href='%{url}'>%{title}</a>"
         time: "Time in %{timezone} is %{time}"
@@ -123,6 +134,7 @@ en:
           one: "Found %{count} <a href='%{url}'>result</a> for '%{query}'"
           other: "Found %{count} <a href='%{url}'>results</a> for '%{query}'"
         setting_context: "Reading context for: %{setting_name}"
+        schema: "%{tables}"
 
     summarization:
       configuration_hint:
--- a/lib/modules/ai_bot/bot.rb
+++ b/lib/modules/ai_bot/bot.rb
@@ -93,9 +93,9 @@ module DiscourseAi
 
         @persona = DiscourseAi::AiBot::Personas::General.new
         if persona_name = post.topic.custom_fields["ai_persona"]
-          personaClass =
+          persona_class =
             DiscourseAi::AiBot::Personas.all.find { |current| current.name == persona_name }
-          @persona = personaClass.new if personaClass
+          @persona = persona_class.new if persona_class
         end
 
         prompt =
--- a/lib/modules/ai_bot/bot.rb
+++ b/lib/modules/ai_bot/bot.rb
@@ -58,6 +58,7 @@ module DiscourseAi
 
       def initialize(bot_user)
         @bot_user = bot_user
+        @persona = DiscourseAi::AiBot::Personas::General.new
       end
 
       def update_pm_title(post)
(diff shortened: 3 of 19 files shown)


SamSaffron:
FEATURE: add initial support for personas

Personas are ways of adjusting our bot framework so it behaves differently based on user expectations. 

This ships with 3 personas: 

- General - which is the standard one we use today
- Artist - which is purpose built for creating pictures
- Sql Helper - which is there to help building data explorer queries

We display a picker now to help pick which persona you want and decorate the bot to print the persona

System is very much V0, we are anticipating allowing Discourse admins to create the personas they want.

assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
```diff
@@ -0,0 +1,53 @@
+import Component from "@glimmer/component";
+import { inject as service } from "@ember/service";
+
+function isBotMessage(composer, currentUser) {
+  if (
+    composer &&
+    composer.targetRecipients &&
+    currentUser.ai_enabled_chat_bots
+  ) {
+    let reciepients = composer.targetRecipients.split(",");
```
martin-brennan, line 10: nit: should be `const` ... I swear we had an eslint for this at some point, now it's gone. Maybe I am nuts

assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @service currentUser;
+
+  get composer() {
+    return this.args?.outletArgs?.model;
+  }
+
+  get botOptions() {
+    if (this.currentUser.ai_enabled_personas) {
```
martin-brennan, line 34: Easier just to do:


```javascript
return (this.currentUser.ai_enabled_personas || []).map() ...
```

config/locales/server.en.yml
```diff
@@ -93,6 +93,16 @@ en:
         markdown_table: Generate Markdown table
 
     ai_bot:
+      personas:
+        general:
+          name: Default
```
martin-brennan, line 98: With the dropdown, having "Default" showing at first seems a little uninspiring and doesn't tell that this is for personas (is personas an end-user term?):

image

Maybe the default one could be called "Flexible" or "Flexible Persona"? Or maybe "Versatile"?

lib/modules/ai_bot/bot.rb
```diff
@@ -90,6 +91,13 @@ def reply_to(
       )
         return if total_completions > MAX_COMPLETIONS
 
+        @persona = DiscourseAi::AiBot::Personas::General.new
+        if persona_name = post.topic.custom_fields["ai_persona"]
+          personaClass =
```
martin-brennan, line 96:
```suggestion
          persona_class =
```

lib/modules/ai_bot/bot.rb
```diff
@@ -90,6 +91,13 @@ def reply_to(
...
+            DiscourseAi::AiBot::Personas.all.find { |current| current.name == persona_name }
+          @persona = personaClass.new if personaClass
```
martin-brennan, line 98:
```suggestion
          @persona = personaClass.new if persona_class
```

lib/modules/ai_bot/entry_point.rb
```diff
@@ -39,13 +39,29 @@ def load_files
...
 
+        plugin.add_to_serializer(
+          :current_user,
+          :ai_enabled_personas,
+          include_condition: -> do
+            SiteSetting.ai_bot_enabled && scope.authenticated? &&
+              scope.user.in_any_groups?(SiteSetting.ai_bot_allowed_groups_map)
+          end,
+        ) do
+          Personas.all.map { |persona| { name: persona.name, description: persona.description } }
```
martin-brennan, line 62: Could maybe just add a class method in personas for this?

```ruby
def self.to_json
  all.map { |persona| { name: persona.name, description: persona.description } }
end
```

lib/modules/ai_bot/entry_point.rb
```diff
@@ -75,6 +91,12 @@ def inject_into(plugin)
 
         plugin.register_svg_icon("robot")
 
+        plugin.add_to_serializer(
+          :topic_view,
+          :persona_name,
```
martin-brennan, line 96: IMO would be better to keep consistent with `ai_persona`...then at least there is indication that this is added by the AI plugin

lib/modules/ai_bot/personas/persona.rb
```diff
@@ -0,0 +1,119 @@
...
+          substitutions[:participants] = topic.allowed_users.map(&:username).join(", ") if topic
+
+          system_prompt.gsub(/\{(\w+)\}/) do |match|
+            found = substitutions[match[1..-2].to_sym]
+            found.nil? ? match : found.to_s
+          end
+        end
+
+        def available_commands
+          return @available_commands if @available_commands
```
martin-brennan, line 54: Can't you just do this?

```ruby
def available_commands
  @available_commands ||= all_available_commands.filter { |cmd| commands.include?(cmd) }
end

File: lib/modules/ai_bot/personas/persona.rb [hunk 9]
```diff
@@ -0,0 +1,119 @@
...
+          return @available_functions if @available_functions
+
+          functions = []
+
+          functions =
+            available_commands.map do |command|
+              function =
+                DiscourseAi::Inference::Function.new(name: command.name, description: command.desc)
+              command.parameters.each do |parameter|
+                function.add_parameter(
```
martin-brennan, line 71: I wonder if the method signature for `add_parameter` could be changed to instead just accept a `Parameter` class instance? (Could do both if ruby had method overloading!)

martin-brennan: Would make things cleaner here:

```ruby
command.parameters.each { |parameter| function.add_parameter(parameter) }
```

lib/modules/ai_bot/personas/persona.rb
martin-brennan: Hmmm I read the rest of the class and maybe yes it's better to keep the way it is for consistency :+1: Disregard!

lib/modules/ai_bot/personas/persona.rb
```diff
@@ -0,0 +1,119 @@
...
+            ].tap do |cmds|
+              cmds << Commands::TagsCommand if SiteSetting.tagging_enabled
+              cmds << Commands::ImageCommand if SiteSetting.ai_stability_api_key.present?
+              if SiteSetting.ai_google_custom_search_api_key.present? &&
+                   SiteSetting.ai_google_custom_search_cx.present?
+                cmds << Commands::GoogleCommand
+              end
+            end
+
+          allowed_commands = SiteSetting.ai_bot_enabled_chat_commands.split("|")
```
martin-brennan, line 113: For this you can also use the `_map` magic, e.g. `SiteSetting.ai_bot_enabled_chat_commands_map`

lib/modules/ai_bot/personas/persona.rb
```diff
@@ -0,0 +1,119 @@
...
+
+          all_commands =
+            [
+              Commands::CategoriesCommand,
+              Commands::TimeCommand,
+              Commands::SearchCommand,
+              Commands::SummarizeCommand,
+              Commands::ReadCommand,
+              Commands::SettingContextCommand,
+            ].tap do |cmds|
```
martin-brennan, line 104: Personal preference but I find the use of `tap` with the array here makes it a bit harder to read (whereas I wouldn't think twice about it for class construction(

martin-brennan (commented):
I wonder if I did something wrong with the SQL persona? It's returning me wrong information about the bookmarks table for example (it no longer has post_id and topic_id, rather bookmarkable_id and bookmarkable_type):

image

assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
SamSaffron: kind of changes behavior though (no longer returns an undefined)

config/locales/server.en.yml
SamSaffron: Yeah tricky, not sure ... will try Forum Helper for now ... though Jammy wants it to be Bert and Lilly wants it to be Lola...

lib/modules/ai_bot/entry_point.rb
SamSaffron: will do

lib/modules/ai_bot/personas/persona.rb
SamSaffron: yeah not a fan of tap, this did not come from me :) will change

SamSaffron:
> I wonder if I did something wrong with the SQL persona? It's returning me wrong information about the bookmarks table for example (it no longer has post_id and topic_id, rather bookmarkable_id and bookmarkable_type):

Try again adjusted the prompt a bit, may need to decrease temp as well.

lib/modules/ai_bot/personas/persona.rb
SamSaffron: undefined method for that, I guess something not right about the site setting... technically though this is a list not a map?

lib/modules/ai_bot/personas/persona.rb
SamSaffron: tough one... looking

lib/modules/ai_bot/personas/persona.rb
martin-brennan: Ah @SamSaffron it's because if you don't define a type for the `list` site settings it returns nil from the type supervisor, and I only add the `_map` method for types `simple` and `compact:

https://github.com/discourse/discourse/blob/d3a5156e6635d6c33c42d12f4dc205646907cddd/lib/site_setting_extension.rb#L530-L533

Also I chose `_map` for the suffix because of ruby + js `.map()` returning an array. Maybe could have been `_list` in retrospect! :)

lib/shared/inference/function.rb
```diff
@@ -12,7 +12,21 @@ def initialize(name:, description:, type: nil)
         @parameters = []
       end
 
-      def add_parameter(name:, type:, description:, enum: nil, required: false)
+      def add_parameter(parameter = nil, **kwargs)
```
martin-brennan, line 15: Heh this is a nice way of "overloading" ;)

assets/javascripts/discourse/connectors/composer-fields/persona-selector.js
martin-brennan: Yeah fair enough :+1:

martin-brennan:
@SamSaffron the SQL helper worked much better this time :ok_hand: 

image

martin-brennan (approved):

lib/modules/ai_bot/personas/persona.rb
martin-brennan: Added here:

https://github.com/discourse/discourse/pull/23331

lib/shared/inference/function.rb
SamSaffron: lol you asked for it :)

PR #178, opened 2023-08-31
FEATURE: new site setting explorer persona

SamSaffron:
Also adds ai_bot_enabled_personas so admins can tweak which stock
personas are enabled.

The new persona has a full listing of all site settings and is
able to get context for each setting.

This means you can ask it to search through settings for something
relevant.

Security wise there is no access to actual configuration of settings
just to the names / description and implementation.

Previously this was part of the forum helper persona however it
just clashes too much with other behaviors, isolating it makes
it far more powerful.

### DIFF
--- a/lib/modules/ai_bot/personas/persona.rb
+++ b/lib/modules/ai_bot/personas/persona.rb
@@ -6,7 +6,10 @@ module DiscourseAi
       def self.all
         personas = [Personas::General, Personas::SqlHelper]
         personas << Personas::Artist if SiteSetting.ai_stability_api_key.present?
-        personas
+        personas << Personas::SettingsExplorer
+
+        personas_allowed = SiteSetting.ai_bot_enabled_personas.split("|")
+        personas.filter { |persona| personas_allowed.include?(persona.to_s.demodulize.underscore) }
       end
 
       class Persona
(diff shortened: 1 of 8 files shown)


lib/modules/ai_bot/personas/persona.rb
```diff
@@ -6,7 +6,10 @@ module Personas
       def self.all
         personas = [Personas::General, Personas::SqlHelper]
         personas << Personas::Artist if SiteSetting.ai_stability_api_key.present?
-        personas
+        personas << Personas::SettingsExplorer
+
+        personas_allowed = SiteSetting.ai_bot_enabled_personas.split("|")
```
martin-brennan, line 11: Remember you can do `SiteSetting.ai_bot_enabled_personas_map` now

lib/modules/ai_bot/personas/settings_explorer.rb
```diff
@@ -0,0 +1,37 @@
...
+
+            - You know the full list of all the site settings.
+            - You are able to request context for a specific setting.
+            - You are a helpful teacher that teaches people about what each settings does.
+
+            Current time is: {time}
+
+            Full list of all the site settings:
+            {{
+            #{SiteSetting.all_settings.map { |setting| setting[:setting].to_s }.join("\n")}
```
martin-brennan, line 27: This is really cool, but shouldn't we give it the description as well? Or is that unnecessary?

martin-brennan (approved):

lib/modules/ai_bot/personas/settings_explorer.rb
SamSaffron: Sadly too many tokens, for Claude though it would be ok

PR #179, opened 2023-09-01
fix settings explorer

SamSaffron:
- FIX: setting explorer was exceeding token budget
- In dev it is important to tell when calls are made to open ai this adds a console log to increase awareness around token usage
- PERF: stop counting tokens so often
- bug fix, commands may reach in for tokenizer
- add logging to console for anthropic calls as well

### DIFF
--- a/lib/modules/ai_bot/commands/search_settings_command.rb
+++ b/lib/modules/ai_bot/commands/search_settings_command.rb
@@ -1,0 +1,85 @@
+#frozen_string_literal: true
+
+module DiscourseAi::AiBot::Commands
+  class SearchSettingsCommand < Command
+    class << self
+      def name
+        "search_settings"
+      end
+
+      def desc
+        "Will search through site settings and return top 20 results"
+      end
+
+      def parameters
+        [
+          Parameter.new(
+            name: "query",
+            description:
+              "comma delimited list of settings to search for (e.g. 'setting_1,setting_2')",
+            type: "string",
+            required: true,
+          ),
+        ]
+      end
+    end
+
+    def result_name
+      "results"
+    end
+
+    def description_args
+      { count: @last_num_results || 0, query: @last_query || "" }
+    end
+
+    INCLUDE_DESCRIPTIONS_MAX_LENGTH = 10
+    MAX_RESULTS = 200
+
+    def process(query:)
+      @last_query = query
+      @last_num_results = 0
+
+      terms = query.split(",").map(&:strip).map(&:downcase).reject(&:blank?)
+
+      found =
+        SiteSetting.all_settings.filter do |setting|
+          name = setting[:setting].to_s.downcase
+          description = setting[:description].to_s.downcase
+          plugin = setting[:plugin].to_s.downcase
+
+          search_string = "#{name} #{description} #{plugin}"
+
+          terms.any? { |term| search_string.include?(term) }
+        end
+
+      if found.blank?
+        {
+          args: {
+            query: query,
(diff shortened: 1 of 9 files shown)


lib/shared/inference/anthropic_completions.rb
```diff
@@ -104,17 +105,22 @@ def self.perform!(
...
           end
+        ensure
+          if block_given?
+            log.update!(
+              raw_response_payload: response_raw,
+              request_tokens: DiscourseAi::Tokenizer::AnthropicTokenizer.size(prompt),
+              response_tokens: DiscourseAi::Tokenizer::AnthropicTokenizer.size(response_data),
+            )
+          end
+          if Rails.env == "development" && log
```
martin-brennan, line 121:
```suggestion
          if Rails.env.development? && log
```

lib/shared/inference/openai_completions.rb
```diff
@@ -125,19 +126,25 @@ def self.perform!(
...
+        if log && block_given?
+          request_tokens = DiscourseAi::Tokenizer::OpenAiTokenizer.size(extract_prompt(messages))
+          response_tokens = DiscourseAi::Tokenizer::OpenAiTokenizer.size(response_data)
+          log.update!(
+            raw_response_payload: response_raw,
+            request_tokens: request_tokens,
+            response_tokens: response_tokens,
+          )
+        end
+        if log && Rails.env == "development"
```
martin-brennan, line 145:
```suggestion
        if log && Rails.env.development?
```

lib/modules/ai_bot/commands/search_settings_command.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    INCLUDE_DESCRIPTIONS_MAX_LENGTH = 10
+    MAX_RESULTS = 200
+
+    def process(query:)
+      @last_query = query
+      @last_num_results = 0
+
+      terms = query.split(",").map(&:strip).map(&:downcase).reject(&:blank?)
```
martin-brennan, line 42: Maybe we can add  min length and early return? Pointless to allow people to use up CPU time searching for `a` :)

lib/shared/inference/anthropic_completions.rb
SamSaffron: yeah people like this incantation, but I really dislike it for some reason

lib/modules/ai_bot/commands/search_settings_command.rb
SamSaffron: I think a is OK, its not chewing up too much cpu... this is going to have to scan through everything no matter what you give it... no indexes.

lib/shared/inference/anthropic_completions.rb
SamSaffron: ok I dug into it... and yeah it is unconditionally better cause it is faster :open_mouth: 

https://github.com//blob/bb52c41dbfdf8d57b58f82319506446b70bca1b1/activesupport/lib/active_support/environment_inquirer.rb#L27-L33

martin-brennan (approved):
Works really nice locally :ok_hand:

lib/modules/ai_bot/commands/search_settings_command.rb
martin-brennan: Yeah I just mean it's kind of pointless to even try with anything < 3 characters long...but if you think it's fine just leave it :+1:

PR #204, opened 2023-09-05
UX: Show suggestion buttons only if sufficient content is present

keegangeorge:
This PR shows the suggest title button only when sufficient content is present in the composer
This PR also makes some improvements to `ai-suggestion-dropdown`'s code quality.

## Preview

https://github.com/discourse/discourse-ai/assets/30090424/879e7b72-1f6c-4a09-9229-f14411bc1712


xfalcox:
> This PR shows the suggest title button only when title input is focused.

Won't that hurt discoverability of the feature? I'd start hiding it only when the composer is empty for now, but this may be something better discussed in Discourse instead of here in a PR.

PR #210, opened 2023-09-07
FEATURE: add semantic search with hyde bot

SamSaffron:
- FEATURE: add semantic search to bot
- FIX: embedding backfill rake task not working

### DIFF
--- a/lib/modules/ai_bot/commands/search_command.rb
+++ b/lib/modules/ai_bot/commands/search_command.rb
@@ -15,7 +15,8 @@ module DiscourseAi::AiBot::Commands
         [
           Parameter.new(
             name: "search_query",
-            description: "Search query (correct bad spelling, remove connector words!)",
+            description:
+              "Specific keywords to search for, space seperated (correct bad spelling, remove connector words)",
             type: "string",
           ),
           Parameter.new(
@@ -93,6 +94,9 @@ module DiscourseAi::AiBot::Commands
       }
     end
 
+    MAX_RESULTS = 20
+    MIN_SEMANTIC_RESULTS = 5
+
     def process(**search_args)
       limit = nil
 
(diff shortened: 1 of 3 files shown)


lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -120,12 +124,35 @@ def process(**search_args)
         )
 
       # let's be frugal with tokens, 50 results is too much and stuff gets cut off
-      limit ||= 20
-      limit = 20 if limit > 20
+      limit ||= MAX_RESULTS
+      limit = MAX_RESULTS if limit > MAX_RESULTS
+
+      should_try_semantic_search = SiteSetting.ai_embeddings_semantic_search_enabled
+      should_try_semantic_search &&= (limit == MAX_RESULTS)
```
martin-brennan, line 131: Oh I never knew about &&=, nice!

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -120,12 +124,35 @@ def process(**search_args)
...
+      should_try_semantic_search &&= (search_args.keys - %i[search_query order]).length == 0
```
martin-brennan, line 132: nit, could be `(search_args.keys - %i[search_query order]).zero?`

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -120,12 +124,35 @@ def process(**search_args)
...
+      should_try_semantic_search &&= (search_args[:search_query].present?)
+
+      limit = limit - MIN_SEMANTIC_RESULTS if should_try_semantic_search
 
       posts = results&.posts || []
       posts = posts[0..limit - 1]
 
+      if should_try_semantic_search
+        semantic_search = DiscourseAi::Embeddings::SemanticSearch.new(Guardian.new())
```
martin-brennan, line 141: I know `Guardian.new()` is making an anon user, but it kind of looks like a weird mistake...we should make a `Guardian.anonymous` method in core to do this. Anyway, doesn't need changing in this PR really, just a thought.

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -120,12 +124,35 @@ def process(**search_args)
...
+        topic_ids = Set.new(posts.map(&:topic_id))
```
martin-brennan, line 142: This is a nice way of getting uniq, I am guessing it's faster than doing `.uniq` on the end?

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -120,12 +124,35 @@ def process(**search_args)
...
+
+        semantic_search
+          .search_for_topics(search_args[:search_query])
+          .each do |post|
+            next if topic_ids.include?(post.topic_id)
+
+            topic_ids << post.topic_id
```
martin-brennan, line 149: I'm guessing it does, just confirming `<<` works for a `Set`?

martin-brennan (approved):

lib/modules/ai_bot/commands/search_command.rb
SamSaffron: It is really splitting hairs cause there are just 20 elements here... but I wanted a Set ... not an Array so the `include?` is faster.

lib/modules/ai_bot/commands/search_command.rb
SamSaffron: true ... I so rarely use zero ActiveSupport is so magical

PR #214, opened 2023-09-08
FEATURE: Create custom prompts with composer AI helper

keegangeorge:
### ✨ FEATURE: Custom Prompts
This PR adds support for custom prompts to composer's AI helper. This PR also makes some minor UX tweaks such as adding icon support.

### 🔍 Preview
https://github.com/discourse/discourse-ai/assets/30090424/6b64975c-2231-435e-9361-a01a14315320


SamSaffron:
This is great, but I would love it if you can just type in the custom prompt direct from the menu (so "Custom Prompt" would just be placeholder text in an INPUT)

keegangeorge:
> This is great, but I would love it if you can just type in the custom prompt direct from the menu (so "Custom Prompt" would just be placeholder text in an INPUT)

Ah I like that idea, so something like this? I can do that
Screenshot 2023-09-11 at 09 53 11

nxuu:
How can the plugin disable automatic title modification?

PR #215, opened 2023-09-11
FEATURE: add tags and categories to read context

SamSaffron:
Note, we perform permission checks on tag list against anon
to ensure we do not disclose information about private tags
to the llm which could get extracted.

### DIFF
--- a/lib/modules/ai_bot/commands/read_command.rb
+++ b/lib/modules/ai_bot/commands/read_command.rb
@@ -52,8 +52,12 @@ module DiscourseAi::AiBot::Commands
 
       content = +<<~TEXT.strip
         title: #{topic.title}
-        category: #{topic.category&.name}
       TEXT
+
+      category_names = [topic.category&.parent_category&.name, topic.category&.name].compact.join(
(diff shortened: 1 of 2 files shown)


ZogStriP (approved):

lib/modules/ai_bot/commands/read_command.rb
```diff
@@ -50,7 +50,15 @@ def process(topic_id:, post_number: nil)
 
       posts = posts.where("post_number = ?", post_number) if post_number
 
-      content = +"title: #{topic.title}\n\n"
+      content = +<<~TEXT.strip
+        title: #{topic.title}
+        category: #{topic.category&.name}
```
vinothkannans, line 55: Can we include the parent category too?

vinothkannans (approved):

lib/modules/ai_bot/commands/read_command.rb
SamSaffron: interesting, yes we can and probably should! will change.

PR #217, opened 2023-09-12
FEATURE: include tag and category context in search

SamSaffron:
Previous to this we just included title/body.. tags and category
structure can be very critical for decision making.

### DIFF
--- a/lib/modules/ai_bot/commands/search_command.rb
+++ b/lib/modules/ai_bot/commands/search_command.rb
@@ -176,7 +176,8 @@ module DiscourseAi::AiBot::Commands
 
           if SiteSetting.tagging_enabled
             hidden_tags ||= DiscourseTagging.hidden_tag_names
-            tags = post.topic.tags.pluck(:name) - hidden_tags
+            # using map over pluck to avoid n+1 (assuming caller preloading)
+            tags = post.topic.tags.map(&:name) - hidden_tags
             row[:tags] = tags.join(", ") if tags.present?
           end
           row
--- a/lib/modules/ai_bot/commands/search_command.rb
+++ b/lib/modules/ai_bot/commands/search_command.rb
@@ -154,17 +154,32 @@ module DiscourseAi::AiBot::Commands
       end
 
       @last_num_results = posts.length
+      # this is the general pattern from core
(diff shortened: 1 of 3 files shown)


vinothkannans (changes requested):
The test failure looks legit.

vinothkannans (approved):
LGTM 👌

SamSaffron:
thanks @vinothkannans was thinking about this and there are some N+1s here ... not sure how easy it is to fix though ... its all in a background job and most of the time you are waiting for LLM, but maybe I should fix.

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -154,17 +154,33 @@ def process(**search_args)
...
             title: post.topic.title,
             url: Discourse.base_path + post.url,
             excerpt: post.excerpt,
             created: post.created_at,
+            category: category_names,
           }
+
+          if SiteSetting.tagging_enabled
+            hidden_tags ||= DiscourseTagging.hidden_tag_names
+            # using map over pluck to avoid n+1 (assuming caller preloading)
```
vinothkannans, line 179: Instead of assuming can we check whether the association is loaded or not? like `post.topic.association(:tags).loaded?`

lib/modules/ai_bot/commands/search_command.rb
```diff
@@ -154,17 +154,33 @@ def process(**search_args)
...
+      # if there are millions of hidden tags it may fail
+      hidden_tags = nil
 
       if posts.blank?
         { args: search_args, rows: [], instruction: "nothing was found, expand your search" }
       else
         format_results(posts, args: search_args) do |post|
-          {
+          category_names = [
+            post.topic.category&.parent_category&.name,
```
vinothkannans, line 166: I don't know how the search command is executed in the background. Will it cause n+1 in category and parent_category attributes too?

lib/modules/ai_bot/commands/search_command.rb
SamSaffron: I think it is the responsibility of Search or SemanticSearch ... but maybe we can raise a warning in case this happens.

PR #222, opened 2023-09-13
FEATURE: add filter support to ai bot semantic search

SamSaffron:
Previously we would bypass semantic search if any filters were
present

Also shows progress now.


tgxworld (approved):

SamSaffron:
thanks @tgxworld failure is legit, will sort it out

PR #226, opened 2023-09-14
FIX: Made bot more robust

SamSaffron:
This is a collection of small fixes

- Display "Searching for: ..." while searching instead of showing found 0 results.
- Only allow 5 commands in lang chain - 6 feels like too much
- On the 5th command stop informing the engine about functions, so it is forced to complete
- Add another 30 tokens of buffer and explain why
- Typo in command prompt

### DIFF
--- a/lib/modules/ai_bot/bot.rb
+++ b/lib/modules/ai_bot/bot.rb
@@ -52,7 +52,7 @@ module DiscourseAi
       attr_reader :bot_user
 
       BOT_NOT_FOUND = Class.new(StandardError)
-      MAX_COMPLETIONS = 6
+      MAX_COMPLETIONS = 5
 
       def self.as(bot_user)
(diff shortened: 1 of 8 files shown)


lib/modules/ai_bot/bot.rb
```diff
@@ -173,6 +168,8 @@ def reply_to(
           end
         end
 
+        puts reply
+
```
tgxworld, line 172:
```suggestion
```

I'm guessing this should be removed.

tgxworld (approved):
I don't have much context about this plugin so I'm just reviewing that the changes look safe. Aside from a small comment, the PR looks  good.

lib/modules/ai_bot/bot.rb
SamSaffron: ooops I knew I forgot something, great catch!

PR #234, opened 2023-09-28
FIX: improve token counting

SamSaffron:
We were running out of tokens under certain conditions (really long
chains)

Add more buffer.


tgxworld (approved):
Approving but I'm wondering if this should be tested? Would not testing it make it harder for someone in the future to work on this area of the code?

SamSaffron:
I wish there was a way to test this :cry:  we need to make API calls against Open AI to validate our algorithm

PR #245, opened 2023-10-05
FEATURE: support OpenAI-Organization header

SamSaffron:
Per: https://platform.openai.com/docs/api-reference/authentication

There is an organization option which is useful for large orgs

> For users who belong to multiple organizations, you can pass a header to specify which organization is used for an API request. Usage from these API requests will count against the specified organization's subscription quota.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -93,6 +93,7 @@ discourse_ai:
   ai_openai_gpt4_url: "https://api.openai.com/v1/chat/completions"
   ai_openai_gpt4_32k_url: "https://api.openai.com/v1/chat/completions"
   ai_openai_embeddings_url: "https://api.openai.com/v1/embeddings"
+  ai_openai_organization:
   ai_openai_api_key:
     default: ""
     secret: true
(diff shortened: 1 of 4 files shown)


SamSaffron:
failure completely unrelated

config/settings.yml
```diff
@@ -93,6 +93,7 @@ discourse_ai:
   ai_openai_gpt4_url: "https://api.openai.com/v1/chat/completions"
   ai_openai_gpt4_32k_url: "https://api.openai.com/v1/chat/completions"
   ai_openai_embeddings_url: "https://api.openai.com/v1/embeddings"
+  ai_openai_organization:
```
tgxworld, line 96: All site settings require a default value to be set. I think we have some checks for this in core but apparently not for plugins.

SamSaffron: weird... seems though like pointless ceremony?

SamSaffron: it return null now that seems to be correct according to code ... anyway ... will give all the blank ones an empty string...

tgxworld (approved):

PR #248, opened 2023-10-09
FIX: Include OP when building title suggestion prompt.

romanrizzi:
When the OP is the only post in the topic, we'll send a prompt without user content to the LLM, and the suggested title will make no sense.


xfalcox (approved):

SamSaffron:
this is fine, but we should follow up with a test here.

PR #251, opened 2023-10-12
DEV: Fix compatibility file

tgxworld:
Why this change?

We should be pinning to 3bced1c6f525553fe2b5b79d29bd66c1cf2a47d5 for
Discourse version smaller than `3.2.0.beta2-dev` which is the version
which the incompatibile change was introduced.


martin-brennan (approved):

tgxworld:
Thank you @martin-brennan :+1:

PR #257, opened 2023-10-19
UX: unify ai representing icon

tytekkit:
This PR addresses the push to unify the icon representing AI throughout Discourse, by using the discourse-sparkles icon.

From this top icon "magic" to only using the bottom icon "discourse-sparkles":

CleanShot 2023-10-19 at 12 43 51@2x

(the textarea overlay is already covered)

The covered areas include:
Related Topics, icon "magic" -> "discourse-sparkles"

If you see any other instances of other icons used throughout Discourse-ai, please commit to this PR.


romanrizzi (approved):

pmusaraj (commented):
We should remove `discourse-sparkles` from this register as well: 

https://github.com/discourse/discourse-ai/blob/main/lib/modules/ai_helper/entry_point.rb#L16 

Also, with changes like this, technically the plugin will have an empty icon if used with a version of core before the accompanying PR is merged. Usually we solve this with a `.discourse-compatibility` file, probably not a big deal here but for the next similar fix.

lib/modules/ai_helper/entry_point.rb
```diff
@@ -12,8 +12,7 @@ def inject_into(plugin)
         plugin.register_seedfu_fixtures(
           Rails.root.join("plugins", "discourse-ai", "db", "fixtures", "ai_helper"),
         )
-
-        additional_icons = %w[discourse-sparkles spell-check language]
+
```
pmusaraj, line 15:
```suggestion
        
additional_icons = %w[spell-check language]
```

This still needs to keep the "spell-check" and "language" icons. This `%w` syntax converts the contents inside the square brackets into an array, using space as the delimiter.

tytekkit: Oh I see, I was thinking ``discourse-sparkles(spell-check, language)``, as I don't speak Ruby.

pmusaraj (approved):

PR #258, opened 2023-10-23
FEATURE: defer creation of bot users

SamSaffron:
Also fixes it so users without bot in header can send it messages.

Previous to this change we would seed all bots with database seeds.

This lead to lots of confusion for people who do not enable ai bot.

Instead:

1. We do not seed any bots **until** user enables the ai_bot_enabled setting
2. If it is disabled we will
  a. If no messages were created by bot - delete it
  b. Otherwise we will deactivate account

### DIFF
--- a/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
+++ b/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
@@ -1,6 +1,7 @@
 import Component from "@glimmer/component";
 import { inject as service } from "@ember/service";
 import { computed } from "@ember/object";
+import I18n from "I18n";
 
 export default class extends Component {
   @service currentUser;
@@ -12,6 +13,18 @@ export default class extends Component {
 
   get renderChatWarning() {
     return this.siteSettings.ai_bot_enable_chat_warning;
+  }
+
+  @computed("composerModel.targetRecipients", "composerModel.title")
+  get aiBotClasses() {
+    if (
+      this.composerModel?.title ===
+      I18n.t("discourse_ai.ai_bot.default_pm_prefix")
+    ) {
+      return "ai-bot-chat";
+    } else {
+      return "ai-bot-pm";
+    }
   }
 
   @computed("composerModel.targetRecipients")
--- a/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
+++ b/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
@@ -1,7 +1,7 @@
 import Component from "@glimmer/component";
 import { inject as service } from "@ember/service";
 import { computed } from "@ember/object";
-import I18n from "I18n";
+import I18n from "discourse-I18n";
 
 export default class extends Component {
   @service currentUser;
--- a/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
+++ b/assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
@@ -1,7 +1,7 @@
 import Component from "@glimmer/component";
 import { inject as service } from "@ember/service";
 import { computed } from "@ember/object";
-import I18n from "discourse-I18n";
+import I18n from "discourse-i18n";
 
 export default class extends Component {
   @service currentUser;
--- a/lib/modules/ai_bot/site_settings_extension.rb
+++ b/lib/modules/ai_bot/site_settings_extension.rb
@@ -1,18 +1,6 @@
 # frozen_string_literal: true
 
 module DiscourseAi::AiBot::SiteSettingsExtension
-  module ClassMethods
-    def ai_bot_enabled=(val)
-      super(val)
-      DiscourseAi::AiBot::SiteSettingsExtension.enable_or_disable_ai_bots
-    end
-
-    def ai_bot_enabled_chat_bots=(val)
-      super(val)
-      DiscourseAi::AiBot::SiteSettingsExtension.enable_or_disable_ai_bots
-    end
-  end
-
   def self.enable_or_disable_ai_bots
     enabled_bots = SiteSetting.ai_bot_enabled_chat_bots_map
     enabled_bots = [] if !SiteSetting.ai_bot_enabled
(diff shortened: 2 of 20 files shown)


assets/javascripts/discourse/connectors/after-d-editor/composer-open.js
```diff
@@ -1,6 +1,7 @@
 import Component from "@glimmer/component";
 import { inject as service } from "@ember/service";
 import { computed } from "@ember/object";
+import I18n from "I18n";
```
martin-brennan, line 4: Lint will probably complain too, but this import is now:

```
import I18n from "discourse-i18n";
```

lib/modules/ai_bot/site_settings_extension.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module DiscourseAi::AiBot::SiteSettingsExtension
+  module ClassMethods
+    def ai_bot_enabled=(val)
```
martin-brennan, line 5: Rather than doing these overrides, you can rely on the plugin API `on(:site_setting_changed)` event:

https://github.com/discourse/discourse/blob/a2000a3559eef9ef17d83ee79cf9a3319eb5df24/plugins/chat/plugin.rb#L259-L269

martin-brennan: It gives you the setting name, old value, and new value.

martin-brennan: You could still keep a separate class though for `enable_or_disable_ai_bots` since you need to run that from the seed too

lib/modules/ai_bot/site_settings_extension.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def ai_bot_enabled_chat_bots=(val)
+      super(val)
+      DiscourseAi::AiBot::SiteSettingsExtension.enable_or_disable_ai_bots
+    end
+  end
+
+  def self.enable_or_disable_ai_bots
+    enabled_bots = SiteSetting.ai_bot_enabled_chat_bots.split("|")
```
martin-brennan, line 17: You can do `SiteSetting.ai_bot_enabled_chat_bots_map` here to get the array :)

SamSaffron:
some broken tests here due to bad assumptions ... will fix...

lib/modules/ai_bot/site_settings_extension.rb
martin-brennan: Could also just do `enabled_bots = SiteSetting.ai_bot_enabled ? SiteSetting.ai_bot_enabled_chat_bots_map : []`

lib/modules/ai_bot/site_settings_extension.rb
```diff
@@ -0,0 +1,59 @@
...
+              username: UserNameSuggester.suggest(bot_name),
+              active: true,
+              approved: true,
+              admin: true,
+              moderator: true,
+              trust_level: TrustLevel[4],
+            )
+          user.save!(validate: false)
+        else
+          user.update!(active: true)
```
martin-brennan, line 39: I guess the only potential issue is that these bang! methods may throw an error, does this stuff happen in a job, or would it prevent the app from booting if e.g. this errored when the seed is run?

lib/modules/ai_bot/site_settings_extension.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+      elsif !active && user
+        # will include deleted
+        has_posts = DB.query_single("SELECT 1 FROM posts WHERE user_id = #{id} LIMIT 1").present?
```
martin-brennan, line 43: This complicates things a bit, but does the AI bot also own things like uploads, or post chat messages? Can it own topics?

martin-brennan: Might be safer to always deactivate?

SamSaffron: very unlikely, I think it is practically impossible for it just to own stuff like uploads, that said longer term maybe if we add a chat modality it will own chat messages and no posts

I prefer to use the hammer here cause it lets us delete a few thousand accounts on our hosting

lib/modules/ai_bot/site_settings_extension.rb
SamSaffron: should never run on seed (it calls outside of it) but yeah it has potential to be an issue, though downside of eating this is that we will not pick up on the error

lib/modules/ai_bot/site_settings_extension.rb
SamSaffron: hmmm maybe, this is cleaner will read through it

lib/modules/ai_bot/site_settings_extension.rb
martin-brennan: Yep no problem, just a thought for future 👍

lib/modules/ai_bot/site_settings_extension.rb
martin-brennan: Yeah it's kind of tricky either way. If it won't prevent boot it's probably fine, though to be super duper safe it could be outsourced to a sidekiq job.

lib/modules/ai_bot/site_settings_extension.rb
martin-brennan: I think in general it's nicer to dogfood the plugin APIs rather than patch things in this way

SamSaffron: the trouble though with the API is that it is called for every setting change... I only want to call this on particular changes

martin-brennan: It's called on every setting change either way, you just need to skip if the setting name doesn't match the one you need. If we're reluctant to use that plugin API it probably indicates it's not a correct design for it, like plugins should be able to register which particular settings they want change events for

martin-brennan (approved):
Approving, I've said my piece about the plugin setting event and you already made the other changes/suggestions :)

SamSaffron:
FYI @martin-brennan moved to the official api as well as part of this.

PR #265, opened 2023-10-24
FIX: make stable diffusion multi site friendly

SamSaffron:
Previous to this change image generation did not work on multisite

There was a background thread generating the images and it was
getting site settings from the default site in the cluster

This also removes referer header which is not needed

### DIFF
--- a/lib/modules/ai_bot/commands/image_command.rb
+++ b/lib/modules/ai_bot/commands/image_command.rb
@@ -47,13 +47,25 @@ module DiscourseAi::AiBot::Commands
 
       results = nil
 
+      # this ensures multisite safety since background threads
+      # generate the images
+      api_key = SiteSetting.ai_stability_api_key
+      engine = SiteSetting.ai_stability_engine
+      api_url = SiteSetting.ai_stability_api_url
+
       # API is flaky, so try a few times
       3.times do
         begin
           thread =
             Thread.new do
               begin
-                results = DiscourseAi::Inference::StabilityGenerator.perform!(prompt)
+                results =
+                  DiscourseAi::Inference::StabilityGenerator.perform!(
+                    prompt,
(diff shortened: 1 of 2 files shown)


lib/shared/inference/stability_generator.rb
```diff
@@ -3,12 +3,15 @@
 module ::DiscourseAi
   module Inference
     class StabilityGenerator
-      def self.perform!(prompt, width: nil, height: nil)
+      def self.perform!(prompt, width: nil, height: nil, api_key: nil, engine: nil, api_url: nil)
+        api_key ||= SiteSetting.ai_stability_api_key
```
martin-brennan, line 7: Isn't this a bit of a footgun to keep these in here? Can we just make the three arguments required instead of default `nil`?

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -65,6 +77,8 @@ def process(prompt:)
         end
       end
 
+      return { prompt: prompt, error: "Something went wrong, could not generate image" } if !results
```
martin-brennan, line 80: Error maybe should be `I18n` if it's shown to users?

martin-brennan (approved):
Just some minor comments but approving to unblock

lib/modules/ai_bot/commands/image_command.rb
SamSaffron: not shown to user just llm

lib/shared/inference/stability_generator.rb
SamSaffron: other apis do support automatic lookup (eg open ai) so I think keeping this aligned is right ... using this from a background thread is a bit of an outlier way of acting.

PR #266, opened 2023-10-25
FEATURE: Make artist more creative

SamSaffron:
This allows for 2 big features:

1. Artist can ship up to 4 prompts for image generation
2. Artist can regenerate images cause it is aware of seed

This allows for iteration on images maintaining visual style

### DIFF
--- a/lib/modules/ai_bot/commands/image_command.rb
+++ b/lib/modules/ai_bot/commands/image_command.rb
@@ -14,11 +14,19 @@ module DiscourseAi::AiBot::Commands
       def parameters
         [
           Parameter.new(
-            name: "prompt",
+            name: "prompts",
             description:
-              "The prompt used to generate or create or draw the image (40 words or less, be creative)",
-            type: "string",
+              "The prompts used to generate or create or draw the image (40 words or less, be creative) up to 4 prompts",
+            type: "array",
+            item_type: "string",
             required: true,
+          ),
+          Parameter.new(
+            name: "seeds",
+            description:
+              "The seed used to generate the image (optional) - can be used to retain image style on amended prompts",
+            type: "array",
+            item_type: "integer",
           ),
         ]
       end
@@ -40,8 +48,12 @@ module DiscourseAi::AiBot::Commands
       @custom_raw
     end
 
-    def process(prompt:)
-      @last_prompt = prompt
+    def process(prompts:, seeds: nil)
+      # max 4 prompts
+      prompts = prompts[0..3]
+      seeds = seeds[0..3] if seeds
+
+      @last_prompt = prompts[0]
 
       show_progress(localized_description)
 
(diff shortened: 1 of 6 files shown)


SamSaffron:
work in progress, need to fix anthropic and tests...

SamSaffron:
Note, this works properly on GPT-3.5 and 4, but we need a followup with Claude. The command framework is way to fragile and needs to be re-implemented for Claude

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -53,55 +65,78 @@ def process(prompt:)
...
-                  DiscourseAi::Inference::StabilityGenerator.perform!(
-                    prompt,
-                    engine: engine,
-                    api_key: api_key,
-                    api_url: api_url,
-                  )
-              rescue => e
-                Rails.logger.warn("Failed to generate image for prompt #{prompt}: #{e}")
-              end
+      i = 0
```
martin-brennan, line 68: Can you just use `.map.with_index` below?

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -53,55 +65,78 @@ def process(prompt:)
...
+      threads =
+        prompts.map do |prompt|
+          seed = seeds ? seeds[i] : nil
+          i += 1
+          Thread.new(seed, prompt) do |inner_seed, inner_prompt|
```
martin-brennan, line 73: I didn't now you could pass args to `Thread` like this, nice

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -53,55 +65,78 @@ def process(prompt:)
...
-
-          show_progress(".", progress_caret: true) while !thread.join(2)
-
-          break if results
+          end
         end
+
+      while true
+        show_progress(".", progress_caret: true)
+        break if threads.all? { |t| t.join(2) }
```
martin-brennan, line 95: I also didn't know about the `limit` seconds for `join` :+1:

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -53,55 +65,78 @@ def process(prompt:)
...
 
-      results[:artifacts].each_with_index do |image, i|
-        f = Tempfile.new("v1_txt2img_#{i}.png")
-        f.binmode
-        f.write(Base64.decode64(image[:base64]))
-        f.rewind
-        uploads << UploadCreator.new(f, "image.png").create_for(bot_user.id)
-        f.unlink
+      i = 0
+      results.each do |result|
```
martin-brennan, line 107: You can use `each_with_index` here too

lib/modules/ai_bot/commands/image_command.rb
```diff
@@ -53,55 +65,78 @@ def process(prompt:)
...
+        result[:artifacts].each do |image|
+          f = Tempfile.new("v1_txt2img_#{i}.png")
```
martin-brennan, line 109: You can do:

```ruby
Tempfile.create("v1_txt2img_#{i}.png") do |file|
  file.binmode
  file.write(Base64.decode64(image[:base64]))
  file.rewind
  // upload stuff
end
```

Then it is automatically closed and unlinked at the end of the block. Ruby docs recommend doing the `close` and `unlink` in an `ensure` block if not using `create` https://devdocs.io/ruby~3.2/tempfile#class-Tempfile-label-Good+practices

martin-brennan (approved):
I don't have an API key for stability so can't try this locally, but the code looks good with some minor comments. Keen to try this on Meta :)

lib/modules/ai_bot/commands/image_command.rb
SamSaffron: yeah, if you dont do that you can get some terrible concurrency bugs when stuff changes underneath you.

lib/modules/ai_bot/commands/image_command.rb
SamSaffron: will change!

lib/modules/ai_bot/commands/image_command.rb
SamSaffron: will change!

SamSaffron:
thanks for the review :hugs:

PR #268, opened 2023-10-27
FIX: avoid semicolons at the end of queries for SQL Helper

SamSaffron:
This makes it harder to cut and paste

Also fine tune the prompt a bit.


Drenmi (approved):
👌

PR #270, opened 2023-10-29
gem adds Chinese mirror source for network acceleration

soberGou:
The existing gem image source is not available in China, causing the installation to fail


soberGou:
The existing gem image source is not available in China, causing the installation to fail

soberGou (commented):
pass

xfalcox:
I guess this was opened by mistake?

PR #272, opened 2023-10-30
FIX: Update bedrock endpoint

xfalcox:
AWS updated their endpoints per https://docs.aws.amazon.com/general/latest/gr/bedrock.html


keegangeorge (approved):

SamSaffron:
lint issues though ...

PR #274, opened 2023-10-31
FEATURE: support topic_id and post_id logging in ai audit log

SamSaffron:
This makes it easier to track who is responsible for a completion
in logs

Note: ai helper and summarization are not yet implemented


SamSaffron:
@romanrizzi / @xfalcox I think this is important - we also want to populate user_id

bot is easy here, but looks like we need lots of changes to support summarization and ai helper.

xfalcox (approved):

PR #280, opened 2023-11-03
DEV: Update linting config

davidtaylorhq:


CvX (approved):
Would you also like to drop our stree plugin, update the stree config, and run `bundle update`?

PR #284, opened 2023-11-07
FEATURE: Expose sentiment classifications via the admin dashboard.

romanrizzi:
This PR adds new reports for displaying information about post sentiments grouped by date and emotions group by TL.

Depends on discourse/discourse#24274

Screenshot 2023-11-07 at 14 45 06


lib/modules/sentiment/entry_point.rb
```diff
@@ -18,6 +18,100 @@ def inject_into(plugin)
...
+              AVG((cr.classification::jsonb->'joy')::integer) AS avg_joy,
+              AVG((cr.classification::jsonb->'disgust')::integer) AS avg_disgust
+            FROM
+              classification_results AS cr
+            INNER JOIN posts p ON p.id = cr.target_id AND cr.target_type = 'Post'
+            INNER JOIN users u ON p.user_id = u.id
+            INNER JOIN topics t ON t.id = p.topic_id
+            INNER JOIN categories c ON c.id = t.category_id
+            WHERE 
+              NOT c.read_restricted AND
```
xfalcox, line 82: Will this make the graph empty for sites where all categories are read restricted?  Since peeps seeing this graph will all be staff anyway, I wonder if we should consider all posts 🤔

xfalcox (approved):

lib/modules/sentiment/entry_point.rb
romanrizzi: Makes sense 👍

PR #288, opened 2023-11-09
FEATURE: basic infrastructure for custom personas

SamSaffron:
- New AiPersona model which can store custom personas
- Persona are restricted via group security
- They can contain custom system messages
- They can support a list of commands optionally

To avoid expensive DB calls in the serializer a Multisite
friendly Hash was introduced (which can be expired on transaction commit)

This means we can safely cache the persona classes and have very low
load time impact

### DIFF
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -42,16 +42,13 @@ class AiPersona < ActiveRecord::Base
       system_prompt = ai_persona.system_prompt
       allowed_group_ids = ai_persona.allowed_group_ids
       commands =
-        ai_persona
-          .commands
-          .map do |inner_name|
-            begin
-              ("DiscourseAi::AiBot::Commands::#{inner_name}").constantize
-            rescue StandardError
-              nil
-            end
+        ai_persona.commands.filter_map do |inner_name|
+          begin
+            ("DiscourseAi::AiBot::Commands::#{inner_name}").constantize
+          rescue StandardError
+            nil
           end
-          .compact
+        end
 
       Class.new(DiscourseAi::AiBot::Personas::Persona) do
         define_singleton_method :name do
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -41,6 +41,7 @@ class AiPersona < ActiveRecord::Base
   def self.all_personas
     persona_cache[:value] ||= AiPersona
       .order(:name)
+      .where(enabled: true)
       .all
       .limit(MAX_PERSONAS_PER_SITE)
       .map do |ai_persona|
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
 class AiPersona < ActiveRecord::Base
+  # places a hard limit, so per site we cache a maximum of 500 classes
+  MAX_PERSONAS_PER_SITE = 500
+
   class MultisiteHash
     def initialize(id)
-      @hash = Hash.new({})
+      @hash = Hash.new { |h, k| h[k] = {} }
       @id = id
 
       MessageBus.subscribe(channel_name) { |message| @hash[message.data] = {} }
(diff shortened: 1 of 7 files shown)


lib/modules/ai_bot/personas/persona.rb
```diff
@@ -3,15 +3,28 @@
...
-        personas.filter { |persona| personas_allowed.include?(persona.to_s.demodulize.underscore) }
+        personas =
+          personas.filter do |persona|
+            personas_allowed.include?(persona.to_s.demodulize.underscore)
+          end
+
+        if user
+          personas.concat(
+            AiPersona.all_personas.filter do |persona|
+              (persona.allowed_group_ids & user.group_ids).length > 0
```
martin-brennan, line 22: We have a shortcut method for this https://github.com/discourse/discourse/blob/e772bb9fce603d95bf23aca3d19576bbc22030f2/app/models/user.rb#L518-L520 

```ruby
user.in_any_groups?(persona.allowed_group_ids)
```

app/models/ai_persona.rb
```diff
@@ -0,0 +1,103 @@
...
+      MessageBus.publish(channel_name, current_db)
+    end
+  end
+
+  def self.persona_cache
+    @persona_cache ||= MultisiteHash.new("persona_cache")
+  end
+
+  def self.all_personas
+    persona_cache[:value] ||= AiPersona.all.map do |ai_persona|
```
martin-brennan, line 39: So this cache ends up having class definitions in it? :exploding_head: 

How does the output of `persona_cache[:value]` look?

db/migrate/20231109011155_create_ai_personas.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+#
+class CreateAiPersonas < ActiveRecord::Migration[7.0]
+  def change
+    create_table :ai_personas do |t|
```
martin-brennan, line 5: I think since admins will be creating these we should have a `created_by_id` linking to a User. Maybe even have a `status` or `enabled` bool, so people can turn them on/off easily without deleting them?

app/models/ai_persona.rb
martin-brennan: Ah I found it locally. With `Class.new` below, can you give it a name? Like `CustomPersona#{name}`? Otherwise the output looks like this and it's not immediately apparent which class is which:

```
AiPersona.all_personas.last
=> #<Class:0x00007f3753759f68>
```

Or maybe add a custom `.inspect` to the generated class to give different output?

app/models/ai_persona.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+class AiPersona < ActiveRecord::Base
+  class MultisiteHash
+    def initialize(id)
+      @hash = Hash.new({})
+      @id = id
+
+      MessageBus.subscribe(channel_name) { |message| @hash[message.data] = {} }
```
martin-brennan, line 9: What is the need behind messagebus?

martin-brennan (commented):
Working really nice locally!

app/models/ai_persona.rb
SamSaffron: we need to communicate cross process to notify everyone the in memory cache is out of date. 

distributedcache can be used here ... but we would need a "version" to cache and finding that is hard. this construct is easier to reason about.

db/migrate/20231109011155_create_ai_personas.rb
SamSaffron: agree totally ... enabled and created_by_id are very handy

app/models/ai_persona.rb
SamSaffron: hmmm let me investigate

SamSaffron: `Object.const_set(name, ...)` wow never knew about this one, just need to be careful not to overlap here and to clear the const... not sure this is worth it

SamSaffron: inspect I guess is easiest here at least we don't need to worry about overlap

app/models/ai_persona.rb
martin-brennan: OK cool makes sense :+1:

app/models/ai_persona.rb
SamSaffron: hmm inspect does not give us much :( 
> Class.new { def inspect; "test"; end  }
=> #<Class:0x00007f52adabfac0>

I think we might as well live with this for now

.const_set hackery can work, but we need to be very careful to be namespaced perfectly, it can possibly lead to terribly complicated bugs

martin-brennan: Yeah even just inspect would be good, just helps with devex when debugging

martin-brennan:
@SamSaffron I tried out inspect/to_s stuff locally, this is how you can define it:

```ruby
define_singleton_method :to_s do
  "#<DiscourseAi::AiBot::Personas::Persona::Custom @name=#{self.name} @allowed_group_ids=#{self.allowed_group_ids.join(',')}>"
end

define_singleton_method :inspect do
  "#<DiscourseAi::AiBot::Personas::Persona::Custom @name=#{self.name} @allowed_group_ids=#{self.allowed_group_ids.join(',')}>"
end
```

Then the result:

```
=> puts AiPersona.all_personas.last
  AiPersona Load (0.7ms)  SELECT "ai_personas".* FROM "ai_personas"
#<DiscourseAi::AiBot::Personas::Persona::Custom @name=Warhammer Nerd @allowed_group_ids=3,1>
```

Can probably get away with just `to_s`

app/models/ai_persona.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+class AiPersona < ActiveRecord::Base
+  class MultisiteHash
+    def initialize(id)
+      @hash = Hash.new({})
```
danielwaterworth, line 6: I think you want `Hash.new { |h, k| h[k] = {} }`, otherwise the same default hash aliases for every key.

app/models/ai_persona.rb
danielwaterworth, line 9: You can get the db from `message.site_id`, you don't need to include it in the message.

danielwaterworth (changes requested):
Since we're storing these in memory, it's important to have a limit on the number of personas and on all of the fields that we are storing.

danielwaterworth:
I also wonder if creating classes at runtime is such a good idea. I presume the idea is that admins will be able to create these personas, but it means that admins can force method cache flushes.

app/models/ai_persona.rb
SamSaffron: that is a bit risky, it would depend on an assumption here (which happens to be correct, but who knows)

app/models/ai_persona.rb
SamSaffron: good catch, thanks fixing

SamSaffron:
> I also wonder if creating classes at runtime is such a good idea. I presume the idea is that admins will be able to create these personas, but it means that admins can force method cache flushes.

I think this is OK for now, method cache flushing is far more efficient these days and this should be pretty rare in the big scheme.

I added some safeguards as well... this construct though (MultisiteHash), I think we should keep it in core as a platform feature? It is a sharp knife, but also is a very useful piece

I restricted the size of this cache by a ton, it will only have name,description,ids ... the bulk of the data which is system_prompt is loaded on demand

db/migrate/20231109011155_create_ai_personas.rb
SamSaffron: going to do this in a followup PR

danielwaterworth:
How about adding a DB limit to the size of the system prompt?

SamSaffron:
picking a limit is very hard for system prompt... we are already at 120k, when I build the UI will start at say 1 million.

SamSaffron:
ok added 1 million for now...

danielwaterworth:
Great! Even a huge limit is better than no limit.

spec/models/ai_persona_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      description: "you write puns",
+      system_prompt: "you are pun bot",
+      commands: ["ImageCommand"],
+      allowed_group_ids: [Group::AUTO_GROUPS[:trust_level_0]],
+    )
+
+    AiPersona.all_personas
+
+    expect(AiPersona.persona_cache[:value].length).to eq(1)
+    RailsMultisite::ConnectionManagement.stubs(:current_db) { "abc" }
```
martin-brennan, line 16: For this we have `test_multisite_connection`:

https://github.com/discourse/discourse/blob/813278717111a8a545d64330decdceb0be6a0006/spec/multisite/site_settings_spec.rb#L15-L19

Just note that the `"second"` is important, it only works with a few predefined values.

martin-brennan (approved):
> I added some safeguards as well... this construct though (MultisiteHash), I think we should keep it in core as a platform feature? It is a sharp knife, but also is a very useful piece

Agree with this, I think in a subsequent PR this could move to core.

Also I saw you are going to add the `created_by` and `enabled` in a followup PR; just curious why not do it here since it's the start of the table?

SamSaffron:
> Also I saw you are going to add the created_by and enabled in a followup PR; just curious why not do it here since it's the start of the table?

honestly, it was mainly laziness of fixing schema on discourse_development ... but I will figure this out and add it in now ... will also fix the multisite thing...

SamSaffron:
Oh 

    p RailsMultisite::ConnectionManagement.all_dbs
  ["default"]

So this does not work in plugin tests :open_mouth:

PR #290, opened 2023-11-10
FEATURE: UI to update ai personas on admin page

SamSaffron:
image

Introduces a UI to manage customizable personas (admin only feature)

Part of the change was some extensive internal refactoring: 

1. AIBot now has a persona set in the constructor, once set it never changes
2. Command now takes in bot as a constructor param, so it has the correct persona and is not generating AIBot objects on the fly
3. Added a `.prettierignore` file, due to the way ALE is configured in nvim it is a pre-req for prettier to work
4. Adds a bunch of validations on the AIPersona model, system personas (artist/creative etc...) are all seeded. We now ensure name uniqueness, and only allow certain properties to be touched for system personas. 
5. (JS note) the client side design takes advantage of nested routes, the parent route for personas gets all the personas via `this.store.findAll("ai-persona")` then child routes simply reach into this model to find a particular persona. 
6. (JS note) data is sideloaded into the ai-persona model the `meta` property supplied from the controller, `resultSetMeta` 
7. This removes ai_bot_enabled_personas and ai_bot_enabled_chat_commands, both should be controlled from the UI on a per persona basis
8. Fixes a long standing bug in token accounting ... we were doing `to_json.length` instead of `to_json.to_s.length`
9. Amended it so {commands} are always inserted at the end unconditionally, no need to add it to the template of the system message as it just confuses things 
10. Adds a concept of `required_commands` to stock personas, these are commands that must be configured for this stock persona to show up.
11. Refactored tests so we stop requiring inference_stubs, it was very confusing to need it, added to plugin.rb for now which at least is clearer 
12. Migrates the persona selector to gjs


assets/javascripts/discourse/components/ai-persona-list-editor.gjs
```diff
@@ -0,0 +1,70 @@
...
+      >
+        {{this.plusIcon}}
+        <span>{{I18n.t "discourse_ai.ai-persona.new"}}</span>
+      </LinkTo>
+    </div>
+    <div class="content-list ai-persona-list-editor">
+      <ul>
+        {{#each @personas as |persona|}}
+          <li
+            class="{{if persona.enabled '' 'disabled'}}
```
martin-brennan, line 47: You can use `concat-class` for this, e.g. 

```javascript
class={{concatClass
  @class
  (if @isLoading "is-loading")
  (if this.btnLink "btn-link" "btn")
  (if this.noText "no-text")
  this.btnType
}}
```

And `import concatClass from "discourse/helpers/concat-class";`

lib/modules/ai_bot/jobs/regular/create_ai_reply.rb
```diff
@@ -6,10 +6,25 @@ class CreateAiReply < ::Jobs::Base
...
       return unless bot_user = User.find_by(id: args[:bot_user_id])
-      return unless bot = DiscourseAi::AiBot::Bot.as(bot_user)
       return unless post = Post.includes(:topic).find_by(id: args[:post_id])
 
-      bot.reply_to(post)
+      kwargs = {}
+      kwargs[:user] = post.user
+      if persona_id = post.topic.custom_fields["ai_persona_id"]
+        persona_id = persona_id.to_i
+        kwargs[:persona_id] = persona_id
```
martin-brennan, line 15:
```suggestion
        kwargs[:persona_id] = persona_id.to_i
```

Could just do this and get rid of previous row

lib/modules/ai_bot/jobs/regular/create_ai_reply.rb
```diff
@@ -6,10 +6,25 @@ class CreateAiReply < ::Jobs::Base
...
+      else
+        kwargs[:persona_name] = post.topic.custom_fields["ai_persona"]
+      end
+
+      begin
+        bot = DiscourseAi::AiBot::Bot.as(bot_user, **kwargs)
+        bot.reply_to(post)
```
martin-brennan, line 22: Could just make this a oneliner too, will throw the BOT_NOT_FOUND either way

```suggestion
        DiscourseAi::AiBot::Bot.as(bot_user, **kwargs).reply_to(post)
```

lib/modules/ai_bot/personas/persona.rb
```diff
@@ -127,8 +150,12 @@ def all_available_commands
             all_commands << Commands::GoogleCommand
           end
 
-          allowed_commands = SiteSetting.ai_bot_enabled_chat_commands.split("|")
-          @cmds = all_commands.filter { |klass| allowed_commands.include?(klass.name) }
+          all_commands
+        end
+
+        def all_available_commands
+          return @cmds if @cmds
```
martin-brennan, line 157: Can't you do `@cmds ||= self.class.all_available_commands` here?

spec/system/ai_bot/persona_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  it "allows creation of a persona" do
+    visit "/admin/plugins/discourse-ai/ai_personas"
+    find(".ai-persona-list-editor__header .btn-primary").click()
+    find(".ai-persona-editor__name").set("Test Persona")
+    find(".ai-persona-editor__description").fill_in(with: "I am a test persona")
+    find(".ai-persona-editor__system_prompt").fill_in(with: "You are a helpful bot")
+    find(".ai-persona-editor__save").click()
+
+    wait_for(timeout: 2) { !page.current_path.include?("new") }
```
martin-brennan, line 19: You don't need this, you can do:

```
  expect(page).not_to have_current_path("/admin/plugins/discourse-ai/ai_personas/new")
```

spec/system/ai_bot/persona_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    persona_id = page.current_path.split("/").last.to_i
```
martin-brennan, line 21: If you know you are off the new page, then you can just look up the record with the data you already have -- can just do `AiPersona.find_by(name: "Test Persona")`.

spec/system/ai_bot/persona_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(persona.name).to eq("Test Persona")
+    expect(persona.description).to eq("I am a test persona")
+    expect(persona.system_prompt).to eq("You are a helpful bot")
+  end
+
+  it "will not allow deletion or editing of system personas" do
+    visit "/admin/plugins/discourse-ai/ai_personas/#{DiscourseAi::AiBot::Personas.system_personas.values.first}"
+    expect(page).not_to have_selector(".ai-persona-editor__delete")
+    prompt = find(".ai-persona-editor__system_prompt")
+    expect(prompt).to be_disabled
```
martin-brennan, line 33: You can do this in one line `expect(find(".ai-persona-editor__system_prompt")).to be_disabled`

spec/system/ai_bot/persona_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  it "will enable persona right away when you click on enable but does not save side effects" do
+    persona = Fabricate(:ai_persona, enabled: false)
+
+    visit "/admin/plugins/discourse-ai/ai_personas/#{persona.id}"
+
+    find(".ai-persona-editor__name").set("Test Persona 1")
+    find(".ai-persona-editor__enabled+span").click()
```
martin-brennan, line 42: You don't need `+span`, you can just click the main selector

spec/system/ai_bot/persona_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    wait_for(timeout: 2) { persona.reload.enabled }
```
martin-brennan, line 44: You should use the system spec specific `try_until_success`, because it takes into account capybara defaults. Then you can also use expectations inside like so:

```
try_until_success do
  expect(persona.reload.enabled).to eq(true)
end
expect(persona.name).not_to eq("Test Persona 1")
```

config/settings.yml
```diff
@@ -250,29 +250,6 @@ discourse_ai:
       - gpt-3.5-turbo
       - gpt-4
       - claude-2
-  ai_bot_enabled_chat_commands:
```
martin-brennan, line 253: We need a small migration to delete the setting values from the DB too for these two deleted settings

assets/javascripts/discourse/templates/admin-plugins/discourse-ai/ai-personas/index.hbs
```diff
@@ -0,0 +1 @@
+<AiPersonaListEditor @personas={{this.model}} />
```
martin-brennan, line 1: I am not sure how to structure it...but I think there is a way you can make these routes using `{{outlet}}` that means you won't have to make multiple templates and repeat this component. Might be worth asking someone like CvX about it, but it works as it is too.

assets/javascripts/discourse/admin/models/ai-persona.js
```diff
@@ -0,0 +1,24 @@
...
+  "description",
+  "commands",
+  "system_prompt",
+  "allowed_group_ids",
+  "enabled",
+  "system",
+  "priority",
+];
+
+export default class AiPersona extends RestModel {
```
martin-brennan, line 14: IIRC we are generally trying to move away from `RestModel` because it doesn't support ember `@tracked` properties... I could be wrong though, something to ask David or CvX about

app/models/ai_persona.rb
```diff
@@ -38,68 +45,115 @@ def self.persona_cache
...
-          define_method :commands do
-            commands
-          end
-
-          define_method :system_prompt do
-            @ai_persona&.system_prompt || "You are a helpful bot."
-          end
-        end
-      end
+      .map { |ai_persona| ai_persona.instance }
```
martin-brennan, line 56: nit

```suggestion
      .map(&:instance)
```

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      def show
+        render json: { ai_persona: @ai_persona }
+      end
+
+      def create
+        ai_persona = AiPersona.new(ai_persona_params)
+        if ai_persona.save
+          render json: { ai_persona: ai_persona }, status: :created
+        else
+          render json: ai_persona.errors, status: :unprocessable_entity
```
martin-brennan, line 27: For all the error ones, you can use `render_json_error(ai_persona)` and it uses `422` by default

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+        if @ai_persona.update(ai_persona_params)
+          render json: @ai_persona
+        else
+          render json: @ai_persona.errors, status: :unprocessable_entity
+        end
+      end
+
+      def destroy
+        @ai_persona.destroy
+        if @ai_persona.errors.present?
```
martin-brennan, line 41: This is flipped from the others, can you do

```ruby
if @ai_persona.destroy
  head :no_content
else
  # render errorrs
end
```

?

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  module Admin
+    class AiPersonasController < ::Admin::AdminController
+      before_action :find_ai_persona, only: %i[show update destroy]
+
+      def index
+        ai_personas =
+          AiPersona.ordered.map { |persona| AiPersonaSerializer.new(persona, root: false) }
+        commands =
+          DiscourseAi::AiBot::Personas::Persona.all_available_commands.map do |command|
+            { id: command.to_s.split("::").last, name: command.name }
```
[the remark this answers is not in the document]
SamSaffron: trying to stay light on serializers and only adding were totally needed, just to keep code simple, but yeah the change makes sense!

app/models/ai_persona.rb
```diff
@@ -4,6 +4,13 @@ class AiPersona < ActiveRecord::Base
   # places a hard limit, so per site we cache a maximum of 500 classes
   MAX_PERSONAS_PER_SITE = 500
 
+  validates :name, presence: true, uniqueness: true, length: { maximum: 100 }
+  validates :description, presence: true, length: { maximum: 2000 }
+  validates :system_prompt, presence: true
```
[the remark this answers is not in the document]
SamSaffron: will add... it is one of those "technically yes" things... but in practice it makes it so we have one more place to edit if we change the rules.

app/models/ai_persona.rb
```diff
@@ -38,68 +45,115 @@ def self.persona_cache
...
+  def system_persona_unchangeable
+    if system_prompt_changed? || commands_changed? || name_changed? || description_changed?
+      errors.add(:base, I18n.t("discourse_ai.ai_bot.personas.cannot_edit_system_persona"))
+    end
+  end
+
+  def ensure_not_system
+    if system
+      errors.add(:base, I18n.t("discourse_ai.ai_bot.personas.cannot_delete_system_persona"))
+      throw :abort
```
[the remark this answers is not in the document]
SamSaffron: Oh I think this is legit, throw :abort stops the chain of callbacks, so in theory it is a bit cleaner, especially for cases where there are more and more callbacks

in this specific case it probably will not make a diff, but it is more future proof.

app/models/ai_persona.rb
```diff
@@ -110,10 +164,14 @@ def bump_cache
 #  name              :string(100)      not null
 #  description       :string(2000)     not null
 #  commands          :string           default([]), not null, is an Array
-#  system_prompt     :string           not null
+#  system_prompt     :string(10000000) not null
```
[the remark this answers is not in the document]
SamSaffron: I picked 10 million cause GPT-4 these days can do 125k tokens which is well over 1 million, I think over the next few years we will get more and more situations where there are massive system prompts, in practice I don't think too many will go for prompts this huge. even all of discourse schema comes in at less than say 50_000

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+  @service store;
+  @service dialog;
+
+  @tracked allGroups = [];
+  @tracked isSaving = false;
+  @tracked justSaved = false;
+  @tracked model = null;
+  @tracked priority = false;
+
+  constructor() {
```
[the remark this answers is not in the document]
SamSaffron: loading groups is expensive, i only ever want to do it once, I guess didInsert works but the constructor does make sure it only happens once in a totally fail safe way.

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+
+  @action
+  toggleEnabled() {
+    this.args.model.set("enabled", !this.args.model.enabled);
+    if (!this.args.model.isNew) {
+      this.args.model.update({ enabled: this.args.model.enabled });
+    }
+  }
+
+  sortPersonas() {
```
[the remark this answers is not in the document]
SamSaffron: :+1:

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+
+    Group.findAll().then((groups) => {
+      this.allGroups = groups;
+    });
+  }
+
+  @action
+  updateModel() {
+    this.model = EmberObject.create(this.args.model.createProperties());
+    this.priority = this.model.priority;
```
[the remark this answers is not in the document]
SamSaffron: hmm, yeah this is not ... will change

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+
+export default class PersonaEditor extends Component {
+  @service router;
+  @service store;
+  @service dialog;
+
+  @tracked allGroups = [];
+  @tracked isSaving = false;
+  @tracked justSaved = false;
+  @tracked model = null;
```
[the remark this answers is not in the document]
SamSaffron: good suggestion will fix

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+    super(...arguments);
+
+    Group.findAll().then((groups) => {
+      this.allGroups = groups;
+    });
+  }
+
+  @action
+  updateModel() {
+    this.model = EmberObject.create(this.args.model.createProperties());
```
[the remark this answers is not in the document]
SamSaffron: hmm but I already have the rest model, I wonder if I should just use it

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+          disabled={{this.model.system}}
+        />
+      </div>
+      <div class="control-group ai-persona-editor__action_panel">
+        <DButton
+          class="btn-primary ai-persona-editor__save"
+          @action={{this.save}}
+          @disabled={{this.isSaving}}
+        >{{I18n.t "discourse_ai.ai-persona.save"}}</DButton>
+        {{#if this.justSaved}}
```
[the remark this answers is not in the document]
SamSaffron: interesting ... we should probably do this in badges as well... will give it a shot

assets/javascripts/discourse/admin/models/ai-persona.js
SamSaffron: it saves so much code though, we need to modernize this thing

assets/javascripts/discourse/templates/admin-plugins/discourse-ai/ai-personas/index.hbs
SamSaffron: oh but they all come with a slightly diff flavor, I kind of like this explicitness

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+  @action
+  updateAllowedGroups(ids) {
+    this.model.set("allowed_group_ids", ids);
+  }
+
+  @action
+  toggleEnabled() {
+    this.args.model.set("enabled", !this.args.model.enabled);
+    if (!this.args.model.isNew) {
+      this.args.model.update({ enabled: this.args.model.enabled });
```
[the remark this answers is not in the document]
SamSaffron: I don't think we do this in badges for example? it can get noisy, when you change it it amends the UI (light grey to dark)

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+    this.args.personas.setObjects(sorted);
+  }
+
+  @action
+  togglePriority() {
+    this.priority = !this.priority;
+
+    this.model.set("priority", this.priority);
+    this.args.model.set("priority", this.model.priority);
+    if (!this.args.model.isNew) {
```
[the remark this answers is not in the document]
SamSaffron: not sure about this for same reason ... ui updates

lib/modules/ai_bot/jobs/regular/create_ai_reply.rb
SamSaffron: mixed on this, a bit of a personal taste here, I don't mind this extra verbosity in this particular case

spec/system/ai_bot/persona_spec.rb
SamSaffron: oh this will time out for me, nice

spec/system/ai_bot/persona_spec.rb
SamSaffron: I tried it, the the __enabled is a disabled element :cry:

config/settings.yml
SamSaffron: oh i had it on local ... checking it in

SamSaffron:
raised internally about the grey on names... it is our default styling...

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
martin-brennan: @SamSaffron `did-insert` only happens once, you should use this and not the constructor. AFAIK there is no way for this to happen multiple times.

lib/modules/ai_bot/jobs/regular/create_ai_reply.rb
martin-brennan: Fair enough!

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
martin-brennan: Actually yeah fair enough for the toggles for enabled + priority, don't need toast for that :+1:

spec/system/ai_bot/persona_spec.rb
martin-brennan: @SamSaffron actually we already have a component page object for toggles, you can do this:

```
PageObjects::Components::DToggleSwitch.new(".ai-persona-editor__enabled").toggle
```

spec/system/ai_bot/persona_spec.rb
martin-brennan: Yep generally if you are manually waiting for something in capybara, you are probably doing the wrong thing, and there is probably a better way.

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
martin-brennan: Yeah not sure...I know for sure we are trying to move away from get/set things in ember and use tracked objects. Not sure totally, it's just something you will need to get David or CVX to weigh in on.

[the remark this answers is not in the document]
martin-brennan: Actually rereading the internal topic about RestModel and EmberData, you should be able to use `@tracked` on your `ApiPersona`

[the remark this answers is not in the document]
martin-brennan: So maybe give that a shot?

app/models/ai_persona.rb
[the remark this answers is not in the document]
martin-brennan: Interesting, I didn't realise you had to manually do this :+1:

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
martin-brennan: Tried it locally and it is indeed safe.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+          head :no_content
+        end
+      end
+
+      private
+
+      def find_ai_persona
+        @ai_persona = AiPersona.find(params[:id])
+      rescue ActiveRecord::RecordNotFound
+        render json: { error: "AiPersona not found" }, status: :not_found
```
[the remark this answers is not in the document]
SamSaffron: true ... will fix

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
SamSaffron: see my latest refactor I did clean this up so there is no longer an ember object involved

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+      this.args.model
+        .update({ priority: this.args.model.priority })
+        .then(() => {
+          this.sortPersonas();
+        });
```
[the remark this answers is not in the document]
SamSaffron: fixed this, also looks better as an async method

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+        this.sortPersonas();
+        if (!error) {
+          if (isNew) {
+            this.args.personas.addObject(this.args.model);
+            this.router.transitionTo(
+              "adminPlugins.discourse-ai.ai-personas.show",
+              this.args.model
+            );
+          } else {
+            later(() => {
```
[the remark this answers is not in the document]
SamSaffron: yeah much cleaner as an async method :+1:

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+          class="btn-primary ai-persona-editor__save"
+          @action={{this.save}}
+          @disabled={{this.isSaving}}
+        >{{I18n.t "discourse_ai.ai-persona.save"}}</DButton>
+        {{#if this.justSaved}}
+          <span class="ai-persona-editor__saved">
+            {{I18n.t "discourse_ai.ai-persona.saved"}}
+          </span>
+        {{/if}}
+        {{#unless @model.system}}
```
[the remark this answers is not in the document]
SamSaffron: cleaned this up

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+    this.args.model.setProperties(this.model);
+    this.args.model
+      .save()
+      .catch((e) => {
+        popupAjaxError(e);
+        error = true;
+      })
+      .finally(() => {
+        this.sortPersonas();
+        if (!error) {
```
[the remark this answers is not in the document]
SamSaffron: yes much cleaner as async

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
SamSaffron: cool done

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,226 @@
...
+  delete() {
+    return this.dialog.confirm({
+      message: I18n.t("discourse_ai.ai-persona.confirm_delete"),
+      didConfirm: () => {
+        return this.args.model.destroyRecord().then(() => {
+          this.args.personas.removeObject(this.args.model);
+          this.router.transitionTo(
+            "adminPlugins.discourse-ai.ai-personas.index"
+          );
+        });
```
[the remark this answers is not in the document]
SamSaffron: on transition? why would we need it here?

spec/system/ai_bot/persona_spec.rb
SamSaffron: I am struggling a bit with this whole page object extra abstraction, that said, sure changed it ..

lib/modules/ai_bot/personas/persona.rb
SamSaffron: sure will change

assets/javascripts/discourse/components/ai-persona-list-editor.gjs
```diff
@@ -0,0 +1,70 @@
...
+
+export default class AiPersonaListEditor extends Component {
+  @tracked _noPersonaText = null;
+
+  constructor() {
+    super(...arguments);
+  }
+
+  get noPersonaText() {
+    if (this._noPersonaText === null) {
```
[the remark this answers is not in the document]
SamSaffron: I guess cause translating markdown is easier than translating html. that said, not a hill I would die on.

spec/system/ai_bot/persona_spec.rb
SamSaffron: a bit unclear on this one?

SamSaffron:
I think we are all addressed now @martin-brennan except for 2 comments needing a tiny bit more clarification. 

I think the biggest issue atm is that once a model is "dirty" we don't have any protection for you navigating away, so its super easy to just forget you are editing and click a different place and lose your work. Any idea how I would wire that up? 

I worry about putting screenshots in cause they can not be translated, added help text for priority

overall this is getting very close!

assets/javascripts/discourse/components/ai-persona-list-editor.gjs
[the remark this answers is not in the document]
martin-brennan: I just think it's a strange pattern to introduce in one place, AFAICT there are no other cases where we use markdown in the translation files. I think in this case the end result is not too complex for translators, and it maintains consistency:

```
<h2>What are AI Personas?</h2>

<p>AI Personas are a powerful feature that allows you to customize the behavior of the AI engine in your Discourse forum. They act as a 'system message' that guides the AI's responses and interactions, helping to create a more personalized and engaging user experience.</p>

<h2>Why use AI Personas?</h2>

<p>With AI Personas, you can tailor the AI's behavior to better fit the context and tone of your forum. Whether you want the AI to be more formal for a professional setting, more casual for a community forum, or even embody a specific character for a role-playing game, AI Personas give you the flexibility to do so.</p>

<h2>Group-Specific Access to AI Personas</h2>

<p>Moreover, you can set it up so that certain user groups have access to specific personas. This means you can have different AI behaviors for different sections of your forum, further enhancing the diversity and richness of your community's interactions.</p>
```

spec/system/ai_bot/persona_spec.rb
martin-brennan: It is tricky at first, but once you get used to it it is very helpful, you aren't having to reinvent the wheel on every system test. In reality these page objects are rarely added or changed,  and they are very useful (e.g. especially the Modal one I find really good)

assets/javascripts/discourse/components/ai-persona-editor.gjs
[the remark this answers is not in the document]
martin-brennan: You need it for the `destroyRecord()` call, not the transition.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+          AiPersona.ordered.map { |persona| AiPersonaSerializer.new(persona, root: false) }
+        commands =
+          DiscourseAi::AiBot::Personas::Persona.all_available_commands.map do |command|
+            { id: command.to_s.split("::").last, name: command.name.humanize.titleize }
+          end
+        render json: { ai_personas: ai_personas, meta: { commands: commands } }
+      end
+
+      def show
+        render json: { ai_persona: @ai_persona }
```
martin-brennan, line 19: This should be using `AiPersonaSerializer`, right?

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def create
+        ai_persona = AiPersona.new(ai_persona_params)
+        if ai_persona.save
```
martin-brennan, line 24: Same here, should both be using `AiPersonaSerializer`

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+          render json: { ai_persona: ai_persona }, status: :created
+        else
+          render_json_error ai_persona
+        end
+      end
+
+      def update
+        if @ai_persona.update(ai_persona_params)
+          render json: @ai_persona
```
martin-brennan, line 33: Sorry...also needs `AiPersonaSerializer`

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,236 @@
...
+        this.toasts.show({
+          data: { message: I18n.t("discourse_ai.ai-persona.saved") },
+          duration: 2000,
+        });
+      }
+    } catch (e) {
+      this.args.model.setProperties(backupModel);
+      popupAjaxError(e);
+    } finally {
+      later(() => {
```
martin-brennan, line 69: Why wait a second to do this?

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,236 @@
...
+      await this.args.model.save();
+      this.#sortPersonas();
+      if (isNew) {
+        this.args.personas.addObject(this.args.model);
+        this.router.transitionTo(
+          "adminPlugins.discourse-ai.ai-personas.show",
+          this.args.model
+        );
+      } else {
+        this.toasts.show({
```
martin-brennan, line 60: Sorry minor, this should be `this.toasts.success`

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,236 @@
...
+    });
+    this.args.personas.clear();
+    this.args.personas.setObjects(sorted);
+  }
+
+  <template>
+    <form
+      class="form-horizontal ai-persona-editor"
+      {{didUpdate this.updateModel @model.id}}
+      {{didInsert this.updateModel @model.id}}
```
martin-brennan, line 146: I don't think you need `@model.id` on either of these `didInsert` calls; that is used to tell `didUpdate` which property triggers the update

config/locales/client.en.yml
```diff
@@ -39,9 +39,38 @@ en:
               description: "Either gpt-4 or gpt-3-5-turbo or claude-2"
 
     discourse_ai:
+      title: "AI"
       modals:
         select_option: "Select an option..."
 
+      ai-persona:
```
martin-brennan, line 46: Didn't catch this before, but this should be `ai_persona`

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -0,0 +1,236 @@
...
+        />
+      </div>
+      <div class="control-group ai-persona-editor__priority">
+        <DToggleSwitch
+          class="ai-persona-editor__priority"
+          @state={{@model.priority}}
+          @label="discourse_ai.ai-persona.priority"
+          {{on "click" this.togglePriority}}
+        />
+        <a
```
martin-brennan, line 164: We can use `DTooltip` for this stuff now...I will push up a commit to change it.

spec/system/ai_bot/persona_spec.rb
martin-brennan: I mean, you already know you are no longer on the New page which means the save worked, so you can just go straight to:

```ruby
persona = AiPersona.find_by(name: "Test Persona")
```

And skip having to get the ID from the URL.

martin-brennan:
@SamSaffron thanks for making all those changes -- I just sent a few more minor ones through, and sent comments for the ones that needed clarification.

> Some help text or documentation around the enabled commands and what they are supposed to do as well would be good; maybe we can show a two-tiered select kit list like the category one, with name and description?

I still think this would be cool, but not a blocker ^^ Also agree with you about the screenshot thing...it's a shame because it's hard to know where these personas are used, but I definitely appreciate the issue with translations.

Also the JS code is looking much nicer now with async added, thanks for doing that :100:

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: actually ... confusingly enough no ... I should probably name stuff "ai_person_class" vs "ai_persona" it makes the code a bit hard to follow ... also thinking of ways of simplifying

assets/javascripts/discourse/components/ai-persona-editor.gjs
SamSaffron: I think I tried this and the component is reused when you navigate between personas so we need to issue it by observing the model id

assets/javascripts/discourse/components/ai-persona-editor.gjs
SamSaffron: will change

config/locales/client.en.yml
SamSaffron: oh my :) I guess its a big editing spree ... will take care

assets/javascripts/discourse/components/ai-persona-editor.gjs
SamSaffron: oh cause it feels like the save function is too fast otherwise... the button flashes too fast.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: let me rename the serializer for clarity

SamSaffron: ok renamed serializer and fixed it

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: updates and creations do not need the localized serializer imo. should be clearer now

SamSaffron:
ok @martin-brennan I think I am done with round 2 :)

assets/javascripts/discourse/components/ai-persona-editor.gjs
martin-brennan: It's fine that the component is reused, it doesn't need to be destroyed + recreated every time, the model itself will update with `updateModel` because `@model.id` changed. Maybe `updateModel` isn't doing enough cleanup?

assets/javascripts/discourse/components/ai-persona-editor.gjs
martin-brennan: But there is a toast too to signify it saved; AFAIK we don't introduce these kind of delays on save elsewhere. Also in prod, usually stuff is a bit slower anyway, it's mostly local that has this issue.

martin-brennan (approved):
Nice one, LGTM now, thanks for bearing with me with the changes :shipit:

PR #292, opened 2023-11-13
FIX: ChatChannel no longer inherits from RestModel.

romanrizzi:
Trying to hydrate using the store will fail. We need to use the model create function instead.


jjaffeux:
It's been like this for months though 😅

jjaffeux (approved):
If it broke for so long without noticing, maybe it needs a test?

romanrizzi:
I believe only a few sites are currently using this module. Will add a test 👍

PR #295, opened 2023-11-14
FIX: start zeitwerk structure

lis2:
Zeitwerk heavily relies on folder structure.

In file `lib/modules/ai_bot/open_ai_bot.rb` it would expect to find `Modules::AiBot::OpenAi` class. Instead, we are defining `DiscourseAi::AiBot::OpenAiBot`


SamSaffron:
closing this in favor of https://github.com/discourse/discourse-ai/pull/319

PR #297, opened 2023-11-19
WIP: Abstract LLM communication

romanrizzi:


lib/completions/endpoints/base.rb
```diff
@@ -0,0 +1,167 @@
...
+
+module DiscourseAi
+  module Completions
+    module Endpoints
+      class Base
+        CompletionFailed = Class.new(StandardError)
+        TIMEOUT = 60
+
+        def self.endpoint_for(model_name)
+          # Order is important.
```
xfalcox, line 11: I'd love to get away from this. Making Azure / Bedrock magically have priority if creds are present work, but we have many people in our team confused about this.


Can we present this as a configuration for our users? Maybe we just duplicate models times inference services?
 Making the model picker look like:

- OpenAI's GPT-4
- Azure GPT-4
- Antrhopic Claude 2
- AWS Claude 2
-

and so on?

romanrizzi: Sure, I understand it can be too magical. I didn't give it much thought and just ported the current behavior.

Bedrock is a separate endpoint because the requests/responses differ from Anthropic's. On the other hand, Azure re-uses the OpenAI settings, so we don't treat it differently. Do you think it makes sense to list it as a separate option?

xfalcox: > Do you think it makes sense to list it as a separate option?

Yes, we had people on our team failing to configure because of this.


IMO we should make it explicit. Maybe we even add a global LLM config, and make the per-module LLM config optional, so it's easier to configure?

romanrizzi: I agree. Do you mind if I take care of that after I complete the transition to this new system? Will add a note so I don't forget.

xfalcox (approved):

PR #298, opened 2023-11-20
FIX: Un-clickable AI Results

keegangeorge:
**This PR fixes an issue where AI results are not clickable.** 

This was due to mapping the original `model.posts` object to add `generatedByAI` attribute which was overwriting the `EmberObject` with a plain JS object causing attributes such as `url` to be removed. Instead we update it to simply add the attribute to the `EmberObject`.

No specs added as specs are still `todo` for after LLM abstraction. However, I've made a personal note to include testing topic url click on spec.

**This PR also applies some smaller fixes related to AI Semantic search:**
- It removes the `background: var(--tertiary-very-low);` that is applied on all AI search results. This makes the UI look cleaner especially when there are only AI results, making the background feel unnecessarily repetitive.
- It disables the toggle switch if there are no AI results.

### DIFF
--- a/assets/javascripts/discourse/connectors/full-page-search-below-search-header/semantic-search.gjs
+++ b/assets/javascripts/discourse/connectors/full-page-search-below-search-header/semantic-search.gjs
@@ -133,12 +133,13 @@ export default class SemanticSearch extends Component {
         })
           .then(async (results) => {
             const model = (await translateResults(results)) || {};
-            const AIResults = model.posts.map(function (post) {
-              return Object.assign({}, post, { generatedByAI: true });
+
+            model.posts.forEach((post) => {
+              post.generatedByAI = true;
             });
 
-            this.args.outletArgs.addSearchResults(AIResults, "topic_id");
-            this.AIResults = AIResults;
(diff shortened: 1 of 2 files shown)


pmusaraj (approved):
Can you followup with a test as well?

assets/javascripts/discourse/connectors/full-page-search-below-search-header/semantic-search.gjs
```diff
@@ -65,6 +65,12 @@ export default class SemanticSearch extends Component {
   @tracked AIResults = [];
   @tracked showingAIResults = false;
 
+  get disableToggleSwitch() {
+    if (this.searching || this.AIResults.length === 0) {
+      return true;
+    }
+  }
```
tytekkit, line 72: Nice 👍

tytekkit (approved):
✅ lgtm

keegangeorge:
> Can you followup with a test as well?

Specs are still not working, but we have a `todo` to get them working once we do the LLM abstraction refactor. But I've made a personal note to ensure we are also testing a click on the AI result.

PR #306, opened 2023-11-22
DEV: Only allow semantic search on "Relevance" sort mode

keegangeorge:
AI semantic search does not support alternative sort modes. This PR updates the UI to make that clear. Semantic search will only work on the "Relevance" sort order. When selecting any other sort mode, semantic search will be disabled.

No tests yet until LLM abstraction PR is merged.

## Preview
https://github.com/discourse/discourse-ai/assets/30090424/fcf52058-a64a-4bcb-85f9-19a64d62f7ab


xfalcox:
Maybe we just make the AI component disabled instead of hiding, so we don't make the whole section jump around and can add a tooltip to why it's disabled?

romanrizzi (approved):

PR #307, opened 2023-11-23
FIX: implement tools framework for Anthropic

SamSaffron:
Previous to this changeset we used a custom system for tools/command
support for Anthropic.

We defined commands by using !command as a signal to execute it

Following Anthropic Claude 2.1, there is an official supported syntax (beta)
for tools execution.

eg:

```
+      <function_calls>
+      <invoke>
+      <tool_name>image</tool_name>
+      <parameters>
+      <prompts>
+      [
+      "an oil painting",
+      "a cute fluffy orange",
+      "3 apple's",
+      "a cat"
+      ]
+      </prompts>
+      </parameters>
+      </invoke>
+      </function_calls>
```

This implements the spec per Anthropic, it should be stable enough
to also work on other LLMs.

Keep in mind that OpenAI is not impacted here at all, as it has its
own custom system for function calls.

Additionally:

- Fixes the title system prompt so it works with latest Anthropic
- Uses new spec for "system" messages by Anthropic
- Tweak forum helper persona to guide Anthropic a tiny be better

Overall results are pretty awesome and Anthropic Claude performs
really well now on Discourse

### DIFF
--- a/lib/modules/ai_bot/anthropic_bot.rb
+++ b/lib/modules/ai_bot/anthropic_bot.rb
@@ -38,7 +38,11 @@ module DiscourseAi
       def build_message(poster_username, content, system: false, function: nil)
         role = poster_username == bot_user.username ? "Assistant" : "Human"
 
-        "#{role}: #{content}"
+        if system || function
+          content
+        else
+          "#{role}: #{content}"
+        end
       end
 
       def model_for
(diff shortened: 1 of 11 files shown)


lib/modules/ai_bot/anthropic_bot.rb
```diff
@@ -61,6 +65,7 @@ def submit_prompt(prompt, post: nil, prefer_low_cost: false, &blk)
           temperature: 0.4,
           max_tokens: 3000,
           post: post,
+          stop_sequences: ["\n\nHuman:", "</function_calls>"],
```
xfalcox, line 68: :chef_kiss:

xfalcox (approved):

romanrizzi (approved):

PR #311, opened 2023-11-24
FEATURE: DALL-E-3 persona for image generation

SamSaffron:
This persona generates images using DALL-E-3 API and is enabled
by default

Keep in mind that we are still waiting on seeds/gen_id so we can
not retain style consistently between turns.

This will change as soon as API params are added to Open AI. 

Also: fixes a minor glitch in the persona selector

### DIFF
--- a/assets/javascripts/discourse/connectors/composer-fields/persona-selector.gjs
+++ b/assets/javascripts/discourse/connectors/composer-fields/persona-selector.gjs
@@ -36,11 +36,14 @@ export default class BotSelector extends Component {
     super(...arguments);
 
     if (this.botOptions && this.composer) {
-      const personaId = this.preferredPersonaStore.getObject("id");
+      let personaId = this.preferredPersonaStore.getObject("id");
+
+      this._value = this.botOptions[0].id;
       if (personaId) {
-        this._value = parseInt(personaId, 10);
-      } else {
-        this._value = this.botOptions[0].personaId;
+        personaId = parseInt(personaId, 10);
+        if (this.botOptions.any((bot) => bot.id === personaId)) {
+          this._value = parseInt(personaId, 10);
+        }
       }
 
       this.composer.metaData = { ai_persona_id: this._value };
--- a/assets/javascripts/discourse/connectors/composer-fields/persona-selector.gjs
+++ b/assets/javascripts/discourse/connectors/composer-fields/persona-selector.gjs
@@ -42,7 +42,7 @@ export default class BotSelector extends Component {
       if (personaId) {
         personaId = parseInt(personaId, 10);
         if (this.botOptions.any((bot) => bot.id === personaId)) {
-          this._value = parseInt(personaId, 10);
+          this._value = personaId;
         }
       }
 
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -143,6 +143,9 @@ en:
         creative:
           name: Creative
           description: "AI Bot with no external integrations specialized in creative tasks"
+        dall_e3:
+          name: "DALL-E 3"
+          description: "AI Bot specialized in generating images using DALL-E 3"
       topic_not_found: "Summary unavailable, topic not found!"
       searching: "Searching for: '%{query}'"
       command_summary:
@@ -157,10 +160,12 @@ en:
         setting_context: "Look up site setting context"
         schema: "Look up database schema"
         search_settings: "Searching site settings"
+        dall_e: "Generate image"
       command_description:
         read: "Reading: <a href='%{url}'>%{title}</a>"
         time: "Time in %{timezone} is %{time}"
         summarize: "Summarized <a href='%{url}'>%{title}</a>"
+        dall_e: "%{prompt}"
         image: "%{prompt}"
         categories:
           one: "Found %{count} category"
--- a/lib/modules/ai_bot/commands/dall_e_command.rb
+++ b/lib/modules/ai_bot/commands/dall_e_command.rb
@@ -64,7 +64,7 @@ module DiscourseAi::AiBot::Commands
           rescue => e
             attempts += 1
             retry if attempts < 3
-            Rails.logger.warn("Failed to generate image for prompt #{prompt}: #{e}")
+            Discourse.warn_exception(e, message: "Failed to generate image for prompt #{prompt}")
             nil
           end
         end
@@ -75,7 +75,7 @@ module DiscourseAi::AiBot::Commands
         break if threads.all? { |t| t.join(2) }
       end
 
-      results = threads.map(&:value).compact
+      results = threads.filter_map(&:value)
 
       if results.blank?
         return { prompts: prompts, error: "Something went wrong, could not generate image" }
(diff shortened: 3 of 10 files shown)


assets/javascripts/discourse/connectors/composer-fields/persona-selector.gjs
```diff
@@ -36,11 +36,14 @@ export default class BotSelector extends Component {
...
+      let personaId = this.preferredPersonaStore.getObject("id");
+
+      this._value = this.botOptions[0].id;
       if (personaId) {
-        this._value = parseInt(personaId, 10);
-      } else {
-        this._value = this.botOptions[0].personaId;
+        personaId = parseInt(personaId, 10);
+        if (this.botOptions.any((bot) => bot.id === personaId)) {
+          this._value = parseInt(personaId, 10);
```
martin-brennan, line 45: No need for extra parse :) 

```suggestion
          this._value = personaId;
```

config/locales/server.en.yml
```diff
@@ -143,6 +143,9 @@ en:
         creative:
           name: Creative
           description: "AI Bot with no external integrations specialized in creative tasks"
+        dall_e3:
```
martin-brennan, line 146: Minor nit, but only here you have called it `dall_e3`, everywhere else is `dall_e`

lib/modules/ai_bot/commands/dall_e_command.rb
```diff
@@ -0,0 +1,116 @@
...
+      false
+    end
+
+    def custom_raw
+      @custom_raw
+    end
+
+    def process(prompts:)
+      # max 4 prompts
+      prompts = prompts[0..3]
```
martin-brennan, line 46:
```suggestion
      prompts = prompts.take(4)
```

lib/modules/ai_bot/commands/dall_e_command.rb
```diff
@@ -0,0 +1,116 @@
...
+      threads = []
+      prompts.each_with_index do |prompt, index|
+        threads << Thread.new(prompt) do |inner_prompt|
+          attempts = 0
+          begin
+            DiscourseAi::Inference::OpenAiImageGenerator.perform!(inner_prompt, api_key: api_key)
+          rescue => e
+            attempts += 1
+            retry if attempts < 3
+            Rails.logger.warn("Failed to generate image for prompt #{prompt}: #{e}")
```
martin-brennan, line 67: `Discourse.warn_exception` is better for this:

https://github.com/discourse/discourse/blob/d6b9c8c828eb48b0e4bc237563fb10aab1ad211b/lib/email/sender.rb#L378-L386

lib/modules/ai_bot/commands/dall_e_command.rb
```diff
@@ -0,0 +1,116 @@
...
+          end
+        end
+      end
+
+      while true
+        show_progress(".", progress_caret: true)
+        break if threads.all? { |t| t.join(2) }
+      end
+
+      results = threads.map(&:value).compact
```
martin-brennan, line 78: You can use `threads.filter_map(&:value)` and it will do the map + compact

lib/modules/ai_bot/commands/dall_e_command.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      if !results.present?
```
martin-brennan, line 80: nit, feel free to ignore

```suggestion
      if results.blank?
```

lib/shared/inference/openai_image_generator.rb
```diff
@@ -0,0 +1,41 @@
...
+module ::DiscourseAi
+  module Inference
+    class OpenAiImageGenerator
+      TIMEOUT = 60
+
+      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil)
+        api_key ||= SiteSetting.ai_openai_api_key
+
+        url = URI("https://api.openai.com/v1/images/generations")
+        headers = { "Content-Type" => "application/json" }
```
martin-brennan, line 12:
```suggestion
        headers = { "Content-Type" => "application/json", "Authorization" => "Bearer #{api_key}" }
```

lib/shared/inference/openai_image_generator.rb
```diff
@@ -0,0 +1,41 @@
...
+        headers["Authorization"] = "Bearer #{api_key}"
+
+        payload = { model: model, prompt: prompt, n: 1, size: size, response_format: "b64_json" }
```
martin-brennan, line 15: What is `n: 1` here?

lib/shared/inference/openai_image_generator.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        Net::HTTP.start(
```
martin-brennan, line 17: Should we maybe use our `FinalDestination::HTTP` library here instead, since it wraps `Net::HTTP`?

https://github.com/discourse/discourse/blob/d6b9c8c828eb48b0e4bc237563fb10aab1ad211b/lib/final_destination.rb#L130-L146

martin-brennan: It automatically follows redirects etc

config/locales/server.en.yml
martin-brennan: Not sure if this will affect translations etc.

SamSaffron: its a tough one, I think I like the command being called dall e cause it has access to 2/3 potentially but the persona is very much tuned for 3 ... not a hill

lib/modules/ai_bot/commands/dall_e_command.rb
SamSaffron: true ... will change

lib/modules/ai_bot/commands/dall_e_command.rb
SamSaffron: I always forget this, will change ... sure

lib/shared/inference/openai_image_generator.rb
SamSaffron: no need to follow redirects here, I kind of prefer to keep this code trivial and use net http

SamSaffron:
@martin-brennan I think I have the review addressed

lib/shared/inference/openai_image_generator.rb
SamSaffron: number of images... but it is meant to only be 1 for now ... I prefer not to omit in case they pick a diff default.

lib/shared/inference/openai_image_generator.rb
martin-brennan: I mostly mean that, nowhere in core do we use `Net::HTTP` directly, it's always `FinalDestination::HTTP`, so it would be good to be consistent.

config/locales/server.en.yml
martin-brennan: Oh right yes I see the distinction now, I thought it was just an inconsistency. Sounds fine :+1:

martin-brennan (approved):

PR #313, opened 2023-11-27
FEATURE: Azure OpenAI support for DALL*E 3

SamSaffron:
Previous to this there was no way to add an inference endpoint for
DALL*E on Azure cause it requires custom URLs

Also:

- On save, when editing a persona it would revert priority and enabled
- More forgiving parsing in command framework for array function calls
- By default generate HD images - they tend to be a bit better
- Improve DALL*E prompt which was getting very annoying and always echoing what it is about to do
- Add a bit of a sleep between retries on image generation
- Fix error handling in image_command
-

### DIFF
--- a/lib/shared/inference/function_list.rb
+++ b/lib/shared/inference/function_list.rb
@@ -50,7 +50,12 @@ module ::DiscourseAi
 
             type = parameter[:type]
             if type == "array"
-              arguments[name] = JSON.parse(value)
+              arguments[name] = begin
+                JSON.parse(value)
+              rescue StandardError
+              end
+              # TODO consider other heuristics as well
+              arguments[name] ||= value.split("\n").map(&:strip).reject(&:blank?)
             elsif type == "integer"
               arguments[name] = value.to_i
             elsif type == "float"
--- a/lib/shared/inference/openai_image_generator.rb
+++ b/lib/shared/inference/openai_image_generator.rb
@@ -5,23 +5,31 @@ module ::DiscourseAi
     class OpenAiImageGenerator
       TIMEOUT = 60
 
-      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil)
+      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, api_url: nil)
         api_key ||= SiteSetting.ai_openai_api_key
+        api_url ||= SiteSetting.ai_openai_dall_e_3_url
 
-        url = URI("https://api.openai.com/v1/images/generations")
-        headers = { "Content-Type" => "application/json", "Authorization" => "Bearer #{api_key}" }
+        uri = URI(api_url)
+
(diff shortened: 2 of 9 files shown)


lib/shared/inference/function_list.rb
```diff
@@ -50,7 +50,12 @@ def coerce_arguments!(parsed)
 
             type = parameter[:type]
             if type == "array"
-              arguments[name] = JSON.parse(value)
+              arguments[name] = begin
+                JSON.parse(value)
+              rescue StandardError
+              end
+              # TODO consider other heuristics as well
```
martin-brennan, line 57: IME a `# TODO` with no name never gets done :) Can you please change to `TODO (sam)` so it has an owner?

martin-brennan: Also what other heuristics are we talking about?

lib/shared/inference/openai_image_generator.rb
```diff
@@ -5,23 +5,31 @@ module Inference
...
-      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil)
+      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, api_url: nil)
         api_key ||= SiteSetting.ai_openai_api_key
+        api_url ||= SiteSetting.ai_openai_dall_e_3_url
 
-        url = URI("https://api.openai.com/v1/images/generations")
-        headers = { "Content-Type" => "application/json", "Authorization" => "Bearer #{api_key}" }
+        uri = URI(api_url)
+
+        headers = { "Content-Type" => "application/json", "quality" => "hd" }
```
martin-brennan, line 14: Interesting choice to have the `quality` as a header, or is supposed to be part of the body?

martin-brennan (approved):
Few comments but overall looks good :+1:

lib/shared/inference/openai_image_generator.rb
SamSaffron: oh ooops good catch will fix

lib/shared/inference/function_list.rb
SamSaffron: thanks, yeah will improve, problem is LLM returns an "array" but you have no idea how it made the array. 

trying json first is good, but then maybe splitting on new line , splitting on comma , deciding what to do based on length / regex... its a tough one.

lib/shared/inference/function_list.rb
```diff
@@ -50,7 +50,12 @@ def coerce_arguments!(parsed)
 
             type = parameter[:type]
             if type == "array"
-              arguments[name] = JSON.parse(value)
+              arguments[name] = begin
```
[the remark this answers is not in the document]
SamSaffron: yeah will tidy this up!

lib/shared/inference/function_list.rb
martin-brennan: Ugh yeah that's annoying that it doesn't have a consistent response...

PR #319, opened 2023-11-28
DEV: port directory structure to Zeitwerk

SamSaffron:
Previous to this change we relied on explicit loading for a files in Discourse AI. 

This had a few downsides: 

1. Busywork whenever you add a file (an extra require relative)
2. We were not keeping to conventions internally ... some places were `OpenAI` others are `OpenAi`
3. Autoloader did not work which lead to lots of full application broken reloads when developing. 

This moves all of DiscourseAI into a Zeitwerk compatible structure.

It also leaves some minimal amount of manual loading (jobs and automation - which is loading into an existing namespace) 

To avoid needing `/lib/discourse_ai/...` we mount a namespace thus we are able to keep `/lib` pointed at `::DiscourseAi` 

Various files were renamed to get around zeitwerk rules and minimize usage of custom inflections


SamSaffron:
inflections are a mess, no clean way to add overrides from what I can tell @lis2 ?

plugin.rb
```diff
@@ -27,59 +27,46 @@
...
-  require_relative "lib/discourse_automation/llm_triage"
+  Rails.autoloaders.each do |autoloader|
+    autoloader.inflector.inflect({ "llm" => "LLM", "chat_gpt" => "ChatGPT" })
+  end
+
+  # do not autoload this cause we may have no namespace
+  require_relative "discourse_automation/llm_triage"
+
+  # jobs are special, they live in a discourse ::Jobs
+  require_relative "jobs/regular/create_ai_reply"
```
martin-brennan, line 49: Chat doesn't need this -- you should be able to put the jobs into `app/jobs/regular/discourse_ai/` (or `app/jobs/scheduled/discourse_ai/`) and they will autoload. Then their class path will be `Jobs::DiscourseAi::CreateAiReply` for example, and you won't need all this require_relative stuff.

martin-brennan (approved):
Apart from the thing about how jobs are loaded this is looking good (and much better than before), though for ultimate confidence you should ask Kris K :)

lib/ai_bot/personas.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      def self.all(user:)
+        personas =
+          AiPersona.all_personas.filter { |persona| user.in_any_groups?(persona.allowed_group_ids) }
+
+        # this needs to be dynamic cause site settings may change
+        all_available_commands = Persona.all_available_commands
+
+        personas.filter do |persona|
```
lis2, line 29: Would it be ok to merge those two filters to remove one loop? Something like
```ruby 
AiPersona.all_personas.filter do |persona|
  return false if !user.in_any_groups?(persona.allowed_group_ids)
  
  if persona.system
    instance = persona.new
    (
      instance.required_commands == [] ||
      (instance.required_commands - all_available_commands).empty?
    )
  else
     true
  end
end
```

lis2 (approved):

lib/ai_bot/personas.rb
SamSaffron: sure ... will change!

PR #329, opened 2023-12-01
FEATURE: User sentiment on profile summary page

xfalcox:
This introduces a new user stat in a user profile summary page.

It will show either neutral/positive/negative according to the dominant
sentiment in the user last interactions.

The user-stat widget is only rendered for staff.

image

image

image
image

### DIFF
--- a/assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
+++ b/assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
@@ -1,16 +1,8 @@
-{{#if
-  (and
-    this.siteSettings.ai_sentiment_enabled
-    @outletArgs.model.sentiment
-    this.showSentiment
-  )
-}}
-  <li class="user-summary-stat-outlet sentiment">
-    <span class="value" title={{i18n "discourse_ai.sentiments.summary.title"}}>
-      {{d-icon this.icon}}
-    </span>
-    <span class="label">
-      {{html-safe (i18n "discourse_ai.sentiments.summary.label")}}
-    </span>
-  </li>
-{{/if}}
+<li class="user-summary-stat-outlet sentiment">
+  <span class="value" title={{i18n "discourse_ai.sentiments.summary.title"}}>
+    {{d-icon this.icon}}
+  </span>
+  <span class="label">
+    {{html-safe (i18n "discourse_ai.sentiments.summary.label")}}
+  </span>
+</li>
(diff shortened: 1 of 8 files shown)


assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
```diff
@@ -0,0 +1,16 @@
+{{#if
+  (and
+    this.siteSettings.ai_sentiment_enabled
+    @outletArgs.model.sentiment
+    this.showSentiment
+  )
+}}
+  <li class="user-summary-stat-outlet sentiment">
```
keegangeorge, line 8: The `user-summary-stat` plugin outlet has:
```
@connectorTagName="li"
```

So this would nest, `<li>` inside an `<li>`. Maybe we should use `<div>` here instead?

assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
```diff
@@ -0,0 +1,16 @@
+{{#if
+  (and
+    this.siteSettings.ai_sentiment_enabled
+    @outletArgs.model.sentiment
+    this.showSentiment
+  )
+}}
```
keegangeorge, line 7: We can probably get rid of this wrapping `{{#if}}` and just move the logic inside the `shouldRender()` method:

```js
  static shouldRender(outletArgs, helper) {
    return helper.siteSettings.ai_sentiment_enabled && outletArgs.model.sentiment && helper.currentUser && helper.currentUser.staff;
  }
  ```

we can also avoid injecting `siteSettings` and `currentUser` services this way since already provided via `helper`

File: assets/javascripts/discourse/connectors/user-summary-stat/sentiment.js [hunk 3]
```diff
@@ -0,0 +1,24 @@
+import Component from "@glimmer/component";
+import { inject as service } from "@ember/service";
+
+export default class Sentiment extends Component {
+  @service siteSettings;
+  @service currentUser;
+
+  get showSentiment() {
+    return this.currentUser && this.currentUser.staff;
+  }
```
keegangeorge, line 10: See comment in template file about using shouldRender() instead

assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
xfalcox: That is not happening

image

assets/javascripts/discourse/connectors/user-summary-stat/sentiment.hbs
xfalcox: :chef_kiss:

keegangeorge (approved):

PR #331, opened 2023-12-04
FEATURE: allow personas to provide command options

SamSaffron:
Personas now support providing options for commands.

This PR introduces a single option "base_query" for the SearchCommand. When supplied all searches the
persona will perform will also include the pre-supplied filter.

This can allow personas to search a subset of the forum (such as documentation) 

This system is extensible we can add options to any command trivially.

image


db/migrate/20231202013850_convert_ai_personas_commands_to_json.rb
```diff
@@ -0,0 +1,31 @@
...
+    execute <<~SQL
+      ALTER TABLE ai_personas
+      ALTER COLUMN commands
+      TYPE json USING array_to_json(commands)
+    SQL
+
+    execute <<~SQL
+      ALTER TABLE ai_personas
+      ALTER COLUMN commands
+      SET DEFAULT '[]'::json
```
xfalcox, line 24: ditto

xfalcox (approved):

SamSaffron:
I was torn on this one, but went with the trivial json here cause there is just so little data stored there a lot of the magic we get with jsonb is just not that important in this use case.

PR #332, opened 2023-12-05
FIX: Use Guardian.basic_user instead of new (anon)

martin-brennan:
c.f. discourse/discourse@de983796e1b66aa2ab039a4fb6e32cec8a65a098

There will soon be additional login_required checks
for Guardian, and the intent of many checks by automated
systems is better fulfilled by using BasicUser, which
simulates a logged in TL0 forum user, rather than an
anon user.

### DIFF
--- a/spec/lib/modules/toxicity/entry_point_spec.rb
+++ b/spec/lib/modules/toxicity/entry_point_spec.rb
@@ -58,7 +58,7 @@ describe DiscourseAi::Toxicity::EntryPoint do
       fab!(:chat_message) { Fabricate(:chat_message) }
       let(:updater) do
         Chat::UpdateMessage.call(
-          guardian: Guardian.new(chat_message.user),
+          guardian: chat_message.user.guardian,
           message_id: chat_message.id,
           message: "This is my updated message",
         )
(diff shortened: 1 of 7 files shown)


SamSaffron:
maybe keep this cross compatible? If not then we need to now flag that AI no longer works with stable via the plugin compatibility thing.

martin-brennan:
@SamSaffron Ah damn I forget about the .discourse-compatibility thing. I am not sure how to keep it compatible in this case -- `Guardian.new` will no longer work for login required sites, so I don't think we have a choice in this case, unless we backport that `BasicUser` class to core stable.

We are due for a version bump soon, and also I wouldn't think AI is on many sites running `stable` (I could be wrong though).

martin-brennan:
@SamSaffron there is already a .discourse-compatibility entry for X introduced by Roman here:

https://github.com/discourse/discourse-ai/commit/e0691e70e87d512517f68fa323682db5f027259d

So this change will have no impact on stable. Just waiting on a core fix for guardian which is causing the test failure here then I need to merge this (and other plugin changes) so I can do redeployments.

spec/lib/modules/toxicity/entry_point_spec.rb
```diff
@@ -58,7 +58,7 @@
       fab!(:chat_message) { Fabricate(:chat_message) }
       let(:updater) do
         Chat::UpdateMessage.call(
-          guardian: Guardian.new(chat_message.user),
+          guardian: chat_message.user.guardian,
```
Drenmi, line 61: Nice! 👍

PR #345, opened 2023-12-11
FEATURE: implement GPT-4 turbo support

SamSaffron:
Keep in mind:

- GPT-4 is only going to be fully released next year - so this hardcodes preview model for now
- Fixes streaming bugs which became a big problem with GPT-4 turbo
- Adds Azure endpoing for turbo as well

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -42,7 +42,7 @@ en:
     ai_openai_gpt35_16k_url: "Custom URL used for GPT 3.5 16k chat completions. (for Azure support)"
     ai_openai_gpt4_url: "Custom URL used for GPT 4 chat completions. (for Azure support)"
     ai_openai_gpt4_32k_url: "Custom URL used for GPT 4 32k chat completions. (for Azure support)"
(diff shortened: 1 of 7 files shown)


config/locales/server.en.yml
```diff
@@ -42,6 +42,7 @@ en:
     ai_openai_gpt35_16k_url: "Custom URL used for GPT 3.5 16k chat completions. (for Azure support)"
     ai_openai_gpt4_url: "Custom URL used for GPT 4 chat completions. (for Azure support)"
     ai_openai_gpt4_32k_url: "Custom URL used for GPT 4 32k chat completions. (for Azure support)"
+    ai_openai_gpt4_turbo_url: "Custom URL used for GPT 4 turbo chat completions. (for Azure support)"
```
martin-brennan, line 45: nit:

```suggestion
    ai_openai_gpt4_turbo_url: "Custom URL used for GPT 4 Turbo chat completions. (for Azure support)"
```

martin-brennan (approved):

PR #349, opened 2023-12-12
FEATURE: try including views/username/likes in search results

SamSaffron:
This is somewhat experimental, but the context of likes/view/username
can help the llm find out what content is more important or even
common users that produce great content

This inflates the amount of tokens somewhat, but given it is all numbers
and search columns titles are only included once this is not severe

### DIFF
--- a/lib/ai_bot/commands/search_command.rb
+++ b/lib/ai_bot/commands/search_command.rb
@@ -208,9 +208,14 @@ module DiscourseAi::AiBot::Commands
           row = {
             title: post.topic.title,
             url: Discourse.base_path + post.url,
+            username: post.user&.username,
             excerpt: post.excerpt,
             created: post.created_at,
             category: category_names,
+            likes: post.like_count,
+            topic_views: post.topic.views,
+            topic_likes: post.topic.like_count,
+            topic_replies: post.topic.posts_count - 1,
           }
 
           if SiteSetting.tagging_enabled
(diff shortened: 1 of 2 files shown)


lib/ai_bot/commands/search_command.rb
```diff
@@ -208,9 +208,14 @@ def process(**search_args)
           row = {
             title: post.topic.title,
             url: Discourse.base_path + post.url,
+            username: post.user&.username,
             excerpt: post.excerpt,
             created: post.created_at,
             category: category_names,
+            likes: post.like_count,
+            topic_views: post.topic.views,
```
martin-brennan, line 216: Just checking, is `post.topic` and `post.user` already preloaded with `includes` elsewhere to avoid N1?

lib/ai_bot/commands/search_command.rb
```diff
@@ -208,9 +208,14 @@ def process(**search_args)
...
+            topic_likes: post.topic.like_count,
+            topic_replies: post.topic.posts_count - 1,
```
martin-brennan, line 218: There is a `reply_count` on `topic` already

martin-brennan (approved):

lib/ai_bot/commands/search_command.rb
SamSaffron: yeah we already had it for title.

lib/ai_bot/commands/search_command.rb
SamSaffron: I think it has slightly diff semantics

martin-brennan: @SamSaffron wow yeah I looked it up and tested it, and it's the count of the number of posts replying to other posts in the topic :S And it's not changed if the post is deleted.

PR #351, opened 2023-12-12
DEV: Update linting

CvX:


CvX:
cc: ai-team - the failing spec file previously didn't have `_spec`  suffix so it never ran. could you take a look?

CvX:
Closed in favor of #423

PR #357, opened 2023-12-14
FEATURE: LLM based peroidical summary report

SamSaffron:
Introduce a Discourse Automation based periodical report. Depends on Discourse Automation. 

Report works best with very large context language models such as GPT-4-Turbo and Claude 2. 

- Introduces `final_insts` to generic llm format, for claude to work best it is better to guide the last assistant message (we should add this to other spots as well) 
- Adds GPT-4 turbo support to generic llm interface


lib/automation/report_runner.rb
```diff
@@ -0,0 +1,140 @@
...
+            tokenizer: @llm.tokenizer,
+          )
+        input = <<~INPUT
+          #{@instructions}
+
+          <context>
+          #{context}
+          </context>
+
+          #{@instructions}
```
romanrizzi, line 92: Looks like we're adding instructions twice.

lib/automation/report_runner.rb
```diff
@@ -0,0 +1,140 @@
...
+            max_posts: @sample_size,
+            tags: @tags,
+            category_ids: @category_ids,
+            prioritized_group_ids: prioritized_group_ids,
+            allow_secure_categories: @allow_secure_categories,
+            tokens_per_post: @tokens_per_post,
+            tokenizer: @llm.tokenizer,
+          )
+        input = <<~INPUT
+          #{@instructions}
```
romanrizzi, line 86: I wonder if this should be added to `insts` instead. My understanding is that guidelines should be part of the system message?

romanrizzi (commented):
If possible, I would like to see the runner tested. It should be easy with `Lllm#with_prepared_responses`

lib/completions/dialects/claude.rb
```diff
@@ -17,7 +17,9 @@ def translate(generic_prompt)
 
           claude_prompt << "#{generic_prompt[:post_insts]}\n" if generic_prompt[:post_insts]
 
-          claude_prompt << "Assistant:\n"
+          claude_prompt << "Assistant:"
+          claude_prompt << " #{generic_prompt[:final_insts]}:" if generic_prompt[:final_insts]
```
xfalcox, line 21: Why are final instructions here under Assistant? Should the lines be swapped in order?

romanrizzi: This is what we talked the other day - Injecting the first sentence of the assistant response gives you a  free boost

lib/automation/report_runner.rb
romanrizzi: TIL

lib/completions/dialects/claude.rb
xfalcox: Ohhh it works well with Claude? That's awesome!

lib/automation/report_runner.rb
SamSaffron: I don't think it helps, I tried it but this was the best way to ground it for giant contexts

lib/completions/dialects/claude.rb
SamSaffron: yeah works super well we should add it everywhere ... saves us from so many annoyances

xfalcox: Yup, and if you are using XML tags to delimit the output, you can even put a small blurb + the opening tag to ensure a consistent reply from the bot.

romanrizzi (approved):

PR #358, opened 2023-12-14
FEATURE: Support for Gemini in AiHelper / Search

xfalcox:


romanrizzi (approved):
Excellent! Would love to see the endpoint tested 😄. We have shared examples you can use.

romanrizzi (approved):

PR #368, opened 2023-12-19
FEATURE: support sending AI report to an email address

SamSaffron:


lib/automation/report_runner.rb
```diff
@@ -138,7 +149,22 @@ def run!
...
+        end
+
+        if @email_receivers.present?
+          @email_receivers.each do |to_address|
+            Email::Sender.new(
+              ::AiReportMailer.send_report(
+                to_address,
+                subject: @title,
+                body: result,
+                from_address: SiteSetting.notification_email,
```
martin-brennan, line 163: You shouldn't need this, the default is to use `notification_email`...I am pretty sure

https://github.com/discourse/discourse/blob/4c8bc34475d0c9dc22d790f5ef3b05570e97048f/lib/email/message_builder.rb#L283-L287

lib/automation/report_runner.rb
```diff
@@ -113,19 +121,22 @@ def run!
...
-            title: @title,
-            archetype: Archetype.private_message,
-            target_usernames: @receivers.map(&:username).join(","),
-            skip_validations: true,
-          )
-
-        if @debug_mode
-          input = input.split("\n").map { |line| "    #{line}" }.join("\n")
-          raw = <<~RAW
+        users = @receivers.map(&:username).join(",")
```
martin-brennan, line 124: nit, but I think this should be called `receiver_usernames` or something, otherwise this variable could appear to be an array of User

SamSaffron:
note the extra patch here fixed a bug in logs on meta.

lib/automation/report_runner.rb
```diff
@@ -52,6 +51,7 @@ def initialize(
       )
         @sender = User.find_by(username: sender_username)
         @receivers = User.where(username: receivers)
+        @email_receivers = receivers.filter { |r| r.include? "@" }
```
martin-brennan, line 54: Maybe we should do a `receivers.filter!` before we use it to look up `User.where()` to save unnecessary querying.

Also to check which of the receivers are an email address, you can use `EmailAddressValidator.valid_value?(r)`

lib/automation/report_runner.rb
```diff
@@ -70,6 +70,14 @@ def initialize(
 
       def run!
         start_date = (@offset + @days).days.ago
+
+        @title =
+          @title.gsub(
```
martin-brennan, line 75:
```suggestion
          @title.gsub!("%DATE%", "#{start_date.strftime("%Y-%m-%d")} - #{end_date.strftime("%Y-%m-%d")}"
```

Does the gsub in place with gsub!, and makes TGX happy because he prefers string interpolation :)

lib/automation/report_runner.rb
```diff
@@ -70,6 +70,14 @@ def initialize(
 
       def run!
         start_date = (@offset + @days).days.ago
```
martin-brennan, line 72:
```suggestion
        start_date = (@offset + @days).days.ago
        end_date = (start_date + @days.days)
```

lib/automation/report_runner.rb
```diff
@@ -113,19 +121,22 @@ def run!
...
+        users = @receivers.map(&:username).join(",")
+
+        if users.present?
+          post =
+            PostCreator.create!(
+              @sender,
+              raw: result,
+              title: @title,
+              archetype: Archetype.private_message,
+              target_usernames: @receivers.map(&:username).join(","),
```
Drenmi, line 133: I guess this can reuse the `users` variable from L:124.

martin-brennan (approved):
Approving because the comments are mostly cosmetic :+1:

lib/automation/report_runner.rb
```diff
@@ -138,7 +149,22 @@ def run!
 
             #{input}
           RAW
-          PostCreator.create!(@sender, raw: raw, topic_id: post.topic_id, skip_validations: true)
+            PostCreator.create!(@sender, raw: raw, topic_id: post.topic_id, skip_validations: true)
+          end
+        end
+
+        if @email_receivers.present?
+          @email_receivers.each do |to_address|
```
Drenmi, line 157: `@email_receivers` should always be an array, so we technically don't need the `#present?` check since `#each` would be a no-op on an empty array, but maybe there's a reason for us to do the check. 🙂

lib/automation/report_runner.rb
SamSaffron: yeah but that amend the caller... I dislike there

lib/automation/report_runner.rb
```diff
@@ -70,6 +70,14 @@ def initialize(
 
       def run!
         start_date = (@offset + @days).days.ago
+
+        @title =
+          @title.gsub(
+            "%DATE%",
+            start_date.strftime("%Y-%m-%d") + " - " +
+              (start_date + @days.days).strftime("%Y-%m-%d"),
```
Drenmi, line 78: I was going to suggest `start_date.iso8601`, but realized `start_date` is actually a `DateTime`. Maybe we could convert `to_date` or rename the variable.

Drenmi (approved):
LGTM! 🚀 

Just some minor suggestions, but nothing in dire need of addressing.

lib/automation/report_runner.rb
martin-brennan: Ah yep fair enough :+1:

lib/automation/report_runner.rb
SamSaffron: I like to keep this nil safe just in case ... but yeah

SamSaffron:
thanks @Drenmi and @martin-brennan !

PR #369, opened 2023-12-19
FIX: triage no longer working with claude

SamSaffron:


SamSaffron:
@romanrizzi this is odd ... I think we should just clean this so we support general translated params: 

that way instead of this awkward code it would be:

```
params: {
   max_tokens: 100
}
```

Way simpler and the dialect can translate this to `max_tokens_to_sample` or whatever. 

```
params: {
   model_name => { ... }
}
```
can then be reserved for extreme edge cases ...

romanrizzi (approved):
I'll add a step for translating parameters in a follow-up PR.

PR #376, opened 2023-12-21
Mixtral

xfalcox:


lib/completions/endpoints/base.rb
```diff
@@ -16,6 +16,7 @@ def self.endpoint_for(model_name)
             DiscourseAi::Completions::Endpoints::OpenAi,
             DiscourseAi::Completions::Endpoints::HuggingFace,
             DiscourseAi::Completions::Endpoints::Gemini,
+            DiscourseAi::Completions::Endpoints::Vllm,
```
romanrizzi, line 19: Remember that we still rely on order here (will change in the future as we discussed). As it is right now, we'll always use HF for these models.

lib/completions/endpoints/vllm.rb
```diff
@@ -0,0 +1,64 @@
...
+        end
+
+        def prepare_request(payload)
+          headers = { "Referer" => Discourse.base_url, "Content-Type" => "application/json" }
+
+          Net::HTTP::Post.new(model_uri, headers).tap { |r| r.body = payload }
+        end
+
+        def extract_completion_from(response_raw)
+          pp response_raw
```
romanrizzi, line 43:
```suggestion
```

lib/completions/endpoints/base.rb
xfalcox: Yup that's the plan. Customers may use HF so it takes priority.

romanrizzi (approved):

PR #377, opened 2023-12-22
FEATURE: allow sending AI based report to a topic

SamSaffron:
This makes the reporting far more flexible cause it can target a
far wider audience by pointing it at a topic in a secure category
or an existing PM

### DIFF
--- a/lib/automation/report_runner.rb
+++ b/lib/automation/report_runner.rb
@@ -36,6 +36,7 @@ module DiscourseAi
       def initialize(
         sender_username:,
         receivers:,
+        topic_id:,
         title:,
         model:,
         category_ids:,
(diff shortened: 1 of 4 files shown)


lib/automation/report_runner.rb
```diff
@@ -35,25 +35,30 @@ def self.run!(**args)
...
+        priority_group_id: nil,
+        allow_secure_categories: false,
+        debug_mode: false
       )
         @sender = User.find_by(username: sender_username)
         @receivers = User.where(username: receivers)
-        @email_receivers = receivers.filter { |r| r.include? "@" }
-        @title = title
-
+        @email_receivers = receivers&.filter { |r| r.include? "@" }
```
martin-brennan, line 55: IIRC I said this on the last PR, but using `EmailAddressValidator.valid_value?(r)` would be better for filtering here

martin-brennan (approved):

lib/automation/report_runner.rb
SamSaffron: I get that ... but then there is even more code required... eg `sam@sam` is not a valid email and certainly not a valid user. 

So in that case we would need to take that ... and raise an exception that you typoed an email... which is fine, but complicates the code

martin-brennan: Ah I see, I hadn't seen the AI UI for this yet -- which doesn't allow entering a bad email address :+1:

PR #385, opened 2023-12-28
FEATURE: allow easy sharing of bot conversations

SamSaffron:
This adds a "share" button to the bottom of all bot posts, which can be used to copy bot conversations to the clipboard: 

image

When clicked there are 2 options: 

1. If there are less than 4 posts in the stream, full context is copied

image

2. If there are more than 4 posts in the stream a new modal opens which allows you to select how much to copy

image

When copied we format using a clear indicator that content came from AI 

image


assets/javascripts/initializers/ai-bot-replies.js
```diff
@@ -166,6 +170,43 @@ function initializePersonaDecorator(api) {
...
+  }
+
+  let shareAiResponse = async function (args) {
+    if (args.post.post_number <= AUTO_COPY_THRESHOLD) {
+      await copyConversation(args.post.topic, 1, args.post.post_number);
+      args.showFeedback("discourse_ai.ai_bot.conversation_shared");
+    } else {
+      modal.show(ShareModal, { model: args.post });
+    }
+  };
```
tgxworld, line 186:
```suggestion
  let shareAiResponse = async function ({ post, showFeedback }) {
    if (post.post_number <= AUTO_COPY_THRESHOLD) {
      await copyConversation(post.topic, 1, post.post_number);
      showFeedback("discourse_ai.ai_bot.conversation_shared");
    } else {
      modal.show(ShareModal, { model: post });
    }
  };
```

Minor one so we don't have to keep referring to `args`.

SamSaffron: agree, looks better, modern js is so nice!

assets/javascripts/discourse/components/modal/share-modal.gjs
```diff
@@ -0,0 +1,108 @@
...
+  @action
+  async copyContext() {
+    const from =
+      this.allPosts[this.allPosts.length - this.contextValue * 2].post_number;
+    const to = this.args.model.post_number;
+    await copyConversation(this.args.model.topic, from, to);
+    this.justCopiedText = t("discourse_ai.ai_bot.conversation_shared");
+    setTimeout(() => {
+      this.justCopiedText = "";
+    }, 2000);
```
tgxworld, line 69: Ember recommends using [ember/later](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/later) instead of `setTimeout`

> You should use this method whenever you need to run some action after a period of time instead of using setTimeout(). This method will ensure that items that expire during the same script execution cycle all execute together, which is often more efficient than using a real setTimeout.

In Discourse, we use `discourseLater` though which I believe has some optimization for the testing env.

SamSaffron: sorted

spec/system/ai_bot/share_spec.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+RSpec.describe "Share conversation", type: :system, js: true do
```
tgxworld, line 2:
```suggestion
RSpec.describe "Share conversation", type: :system do
```

The `js: true` metadata does not do anything for our system tests setup and is therefore not required.

spec/system/ai_bot/share_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      clip_text = page.evaluate_script("window.discourseAiClipboard")
+      expect(clip_text).to be_present
+    end
+
+    clip_text = page.evaluate_script("window.discourseAiClipboard")
+
+    expect(clip_text).to include("Conversation with AI")
+    expect(clip_text).to include("user reply 1")
+    expect(clip_text).to include("bot reply 1")
+    expect(clip_text).not_to include("bot reply 2")
```
tgxworld, line 61:
```suggestion
    expect(clip_text).to eq(<<~RAW.chomp)
    <details class='ai-quote'>
    <summary>
    <span>This is my special PM</span>
    <span title='Conversation with AI'>AI</span>
    </summary>


    **#{admin.username}:**

    test test test user reply 1

    **#{bot_user.username}:**

    test test test bot reply 1
    </details>
    RAW
```

Personally I think we should just test the entire content in the clipboard to properly cover the feature since the markdown is important as well.

spec/system/ai_bot/share_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    # clipboard functionality is extremely hard to test
+    # we would need special permissions in chrome driver to enable full access
+    # instead we use a secret variable to signal that we want to store clipboard
+    # data in window.discourseAiClipboard
+    page.execute_script("window.discourseAiTestClipboard = true")
+
+    find("#post_2 .post-action-menu__share").click
+
+    try_until_success do
```
tgxworld, line 51:
```suggestion
    clip_text = nil

    try_until_success do
```

spec/system/ai_bot/share_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      clip_text = page.evaluate_script("window.discourseAiClipboard")
+      expect(clip_text).to be_present
+    end
+
+    clip_text = page.evaluate_script("window.discourseAiClipboard")
```
tgxworld, line 56:
```suggestion
```

This can be removed with https://github.com/discourse/discourse-ai/pull/385/files#r1438001462

spec/system/ai_bot/share_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    find("#post_6 .post-action-menu__share").click
+    find(".ai-share-modal__slider input").set("2")
+    find(".ai-share-modal button.btn-primary").click
+
+    try_until_success do
+      clip_text = page.evaluate_script("window.discourseAiClipboard")
+      expect(clip_text).to be_present
+    end
+
+    clip_text = page.evaluate_script("window.discourseAiClipboard")
```
tgxworld, line 75:
```suggestion
```

This can be removed with https://github.com/discourse/discourse-ai/pull/385/files#r1438001462

tgxworld (approved):

spec/system/ai_bot/share_spec.rb
SamSaffron: legit, addressed

spec/system/ai_bot/share_spec.rb
SamSaffron: addressed

spec/system/ai_bot/share_spec.rb
SamSaffron: addressed this (I prefer my version, check it out, just explicitly picked names.)

PR #386, opened 2023-12-28
FEATURE: Render Related Topics for Crawlers

xfalcox:
Depends on https://github.com/discourse/discourse/pull/25060

Looks like

image

### DIFF
--- a/app/views/list/related_topics.html.erb
+++ b/app/views/list/related_topics.html.erb
@@ -1,0 +1,71 @@
+<div id="related-topics" class="more-topics__list " role="complementary" aria-labelledby="related-topics-title">
+  <h3 id="related-topics-title" class="more-topics__list-title">
+    <%= t 'js.discourse_ai.related_topics.title' %>
+  </h3>
+  <div class="topic-list-container" itemscope itemtype='http://schema.org/ItemList'>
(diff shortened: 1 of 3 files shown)


app/views/list/related_topics.html.erb
```diff
@@ -0,0 +1,71 @@
...
+      <tbody>
+        <% @list.each_with_index do |t,i| %>
+          <tr class="topic-list-item">
+            <td class="main-link" itemprop='itemListElement' itemscope itemtype='http://schema.org/ListItem'>
+              <meta itemprop='position' content='<%= i + 1 %>'>
+              <span class="link-top-line">
+                <a itemprop='url' href='<%= t.url %>' class='title raw-link raw-topic-link'><%= t.title %></a>
+              </span>
+              <div class="link-bottom-line">
+                <% if (!@category || @category.has_children?) && t.category && !t.category.uncategorized? %>
```
romanrizzi, line 26: Do we need to check category is public?

romanrizzi (approved):

app/views/list/related_topics.html.erb
xfalcox: Handled by https://github.com/discourse/discourse-ai/blob/1287ef4428fbec2ec211bdf62f26444ac56576d3/lib/embeddings/semantic_topic_query.rb#L16C7-L16C18 AFAIK

PR #387, opened 2023-12-28
FEATURE: Per post embeddings

xfalcox:


spec/lib/modules/embeddings/vector_representations/vector_rep_shared_examples.rb
```diff
@@ -25,7 +25,7 @@
         )
       stub_vector_mapping(text, @expected_embedding)
 
-      vector_rep.generate_topic_representation_from(topic)
+      vector_rep.generate_representation_from(topic)
```
SamSaffron, line 28: also add a minimal test around here for a post embedding? at least code will get covered.

SamSaffron:
Overall super happy to have this... I think we are going to need a third variant as well for "arbitrary" embeddings. 

For example a "Browse web" tool want to embed web pages and associate that with topic_id/tool ... 

will be a follow on for this work.

romanrizzi (approved):

PR #402, opened 2024-01-04
FEATURE: AI Bot Gemini support.

romanrizzi:
It also corrects the syntax around tool support, which was wrong.

Gemini doesn't want us to include messages about previous tool invocations, so I had to shuffle around some code to send the response it generated from those invocations instead. For this, I created the "multi_turn" context, which bundles all the context involved in the interaction.

### DIFF
--- a/lib/completions/dialects/gemini.rb
+++ b/lib/completions/dialects/gemini.rb
@@ -15,6 +15,9 @@ module DiscourseAi
         end
 
         def translate
+          # Gemini complains if we don't alternate model/user roles.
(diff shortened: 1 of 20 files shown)


lib/completions/dialects/gemini.rb
```diff
@@ -15,14 +15,17 @@ def tokenizer
         end
 
         def translate
+          # Gemini complains if we don't alternate model/user roles.
+          noop_model_response = { role: "model", parts: { text: "Ok." } }
```
xfalcox, line 19: lol I hate this too

xfalcox (approved):

PR #405, opened 2024-01-05
FIX: array arguments not parsed correctly

SamSaffron:
- FIX: array arguments not parsed correctly
- Side quest ... don't use update! it calls validations and will now fail due to email validation

### DIFF
--- a/lib/ai_bot/personas/persona.rb
+++ b/lib/ai_bot/personas/persona.rb
@@ -136,9 +136,7 @@ module DiscourseAi
                 end
             end
 
-            if value
-              arguments[name.to_sym] = value if value
-            end
+            arguments[name.to_sym] = value if value
           end
 
           tool_klass.new(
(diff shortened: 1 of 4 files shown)


SamSaffron:
side quest used to work, something changed in core ...

lib/ai_bot/personas/persona.rb
```diff
@@ -122,16 +122,24 @@ def find_tool(partial)
...
+                begin
+                  JSON.parse(value)
+                rescue JSON::ParserError
+                  nil
+                end
+            end
+
+            if value
+              arguments[name.to_sym] = value if value
+            end
```
tgxworld, line 141: There is a double `if value` conditional here. Not sure why Github doesn't allow me to use code suggestion here.

tgxworld, line 141:
```
            if value
              if param[:type] == "array"
                value =
                begin
                  JSON.parse(value)
                rescue JSON::ParserError
                  nil
                end
              end

              arguments[name.to_sym] = value
            end
```

SamSaffron: Oops will fix

martin-brennan (approved):

PR #411, opened 2024-01-09
FIX: Use claude-2.1 to enable system prompts

romanrizzi:


spec/lib/completions/dialects/claude_spec.rb
```diff
@@ -48,8 +48,8 @@
   describe "#translate" do
     it "translates a prompt written in our generic format to Claude's format" do
       anthropic_version = <<~TEXT
-      #{prompt[:insts]}
-      #{prompt[:input]}
+      System: #{prompt[:insts]}
```
xfalcox, line 51: ditto

xfalcox (changes requested):
https://docs.anthropic.com/claude/docs/how-to-use-system-prompts#legacy-system-prompts-via-text-completions-api

romanrizzi:
@xfalcox - My bad. Removed the `System: `

xfalcox (approved):

PR #414, opened 2024-01-10
FEATURE: Support for SRV records for Discourse services

xfalcox:
This allows admins to configure services with multiple backends using DNS SRV records. This PR also adds support for shared secret auth via headers for TEI and vLLM endpoints, so they are inline with the other ones.


romanrizzi (approved):
Should these settings be hidden?

PR #419, opened 2024-01-11
DEV: Fix SemanticRelated module load error

martin-brennan:
Followup 2636efcd1bf6eaa0a6d0d868affb9d41d49bdda2,
whenever ruby code was changed locally this would break
module loading, giving an "uninitialized constant
DiscourseAi::Embeddings::EntryPoint::SemanticRelated" error.

### DIFF
--- a/app/views/list/related_topics.html.erb
+++ b/app/views/list/related_topics.html.erb
@@ -15,23 +15,23 @@
         </tr>
       </thead>
       <tbody>
-        <% @list.each_with_index do |t,i| %>
-          <tr class="topic-list-item">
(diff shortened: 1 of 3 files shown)


SamSaffron (approved):

SamSaffron:
I guess we should have a trivial test here, maybe just make an http request as a crawler and see what happens, a bit annoying cause we need to add mock data to semantic related, but seems feasible.

martin-brennan:
@SamSaffron I will take a shot at adding a spec, I see there are crawler-based topic specs in core :+1:

PR #420, opened 2024-01-12
FEATURE: even smoother streaming

SamSaffron:
- Account properly for function calls, don't stream through <details> blocks
- Rush cooked content back to client
- Wait longer (up to 60 seconds) before giving up on streaming
- Clean up message bus channels so we don't have leftover data
- Make ai streamer much more reusable and much easier to read
- If buffer grows quickly, rush update so you are not artificially waiting


lib/completions/prompt.rb
```diff
@@ -8,13 +8,19 @@ class Prompt
...
-      def initialize(system_msg, messages: [], tools: [])
+      def initialize(system_msg = nil, messages: [], tools: [])
         raise ArgumentError, "messages must be an array" if !messages.is_a?(Array)
         raise ArgumentError, "tools must be an array" if !tools.is_a?(Array)
 
-        system_message = { type: :system, content: system_msg }
+        @messages = []
+
+        if system_msg
+          system_message = { type: :system, content: system_msg }
```
martin-brennan, line 18: I think this is setting the `system_message` defined by `attr_reader`; in this case it might be slightly clearer to do `@system_message = `, I was confused otherwise why there was an intermediate var here

test/javascripts/unit/lib/ai-streamer-test.js
```diff
@@ -1,5 +1,53 @@
...
-import { addProgressDot } from "discourse/plugins/discourse-ai/discourse/lib/ai-streamer";
+import {
+  addProgressDot,
+  applyProgress,
+  MIN_LETTERS_PER_INTERVAL,
+} from "discourse/plugins/discourse-ai/discourse/lib/ai-streamer";
+
+class FakeStreamUpdater {
+  constructor() {
+    this._streaming = true;
```
martin-brennan, line 10: The preferred syntax for private vars is to use `#` now, e.g. `this.#streaming = true`

martin-brennan: You just need to define them at the top of the class like this:

https://github.com/discourse/discourse/blob/f5fc7790ce7128e99c32e14157506ff2636b239d/app/assets/javascripts/discourse/app/services/loading-slider.js#L15-L18

assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -41,69 +42,140 @@ export function addProgressDot(element) {
...
     }
   }
 
-  return postStatus.done;
+  if (status.raw !== undefined) {
+    let newRaw = status.raw;
+
+    if (!status.done) {
+      // rush update if we have a </details> tag (function call)
+      if (oldRaw.length === 0 && newRaw.indexOf("</details>") !== -1) {
```
martin-brennan, line 151:
```suggestion
      if (oldRaw.length === 0 && newRaw.includes("</details>")) {
```

martin-brennan (approved):
I think it is better, it's a little bit tricky to tell though. It's a complex system so I'm not sure when it does slow down whether it's this streaming part on the client, or on the server, or simply because my internet calls to OpenAI are slow.

Roman should take a look tomorrow too, but I am approving since there are just minor JS things I have commented on.

assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -41,69 +42,140 @@ export function addProgressDot(element) {
   lastBlock.appendChild(dotElement);
 }
 
-async function applyProgress(postStatus, postStream) {
-  postStatus.startTime = postStatus.startTime || Date.now();
-  let post = postStream.findLoadedPost(postStatus.post_id);
+// this is the interface we need to implement
+// for a streaming updater
+class StreamUpdater {
+  set streaming(value) {}
```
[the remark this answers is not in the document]
SamSaffron: will do

assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -41,69 +42,140 @@ export function addProgressDot(element) {
...
+  async setRaw(value, done) {
+    this.post.set("raw", value);
+    const cooked = await cook(value);
 
     // resets animation
-    postElement.classList.remove("streaming");
-    void postElement.offsetWidth;
-    postElement.classList.add("streaming");
+    this.element.classList.remove("streaming");
+    void this.element.offsetWidth;
```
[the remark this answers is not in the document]
SamSaffron: its a trick to trigger animation resets

test/javascripts/unit/lib/ai-streamer-test.js
SamSaffron: oh will fix

assets/javascripts/discourse/lib/ai-streamer.js
SamSaffron: will change but I need to cover this code

assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -41,69 +42,140 @@ export function addProgressDot(element) {
...
-  if (!post) {
-    // wait till later
-    return false;
+  get element() {
+    return this.postElement;
   }
 
-  const oldRaw = post.get("raw") || "";
+  get loaded() {
+    return !!this.postElement;
```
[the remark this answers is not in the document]
SamSaffron: actually I think loaded is just not needed, removing from interface, simpler interface feels better

lib/completions/prompt.rb
SamSaffron: oh yeah cleaning this up

assets/javascripts/discourse/lib/ai-streamer.js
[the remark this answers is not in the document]
martin-brennan: That's cool, didn't know that, this explains it nicely https://stackoverflow.com/a/60686560 :)

PR #422, opened 2024-01-12
DEV: Prevent HyDE search from being called multiple times

keegangeorge:
`onAppEvent("full-page-search:trigger-search")` is triggered at minimum two times when a search is queried, sometimes three. This was causing a HyDE search request multiple times likely resulting in reaching rate limits (especially on mobile).

This PR prevents HyDE from being called multiple times for a single query by checking if a search is currently in progress before calling `performHyDESearch()`

### DIFF
--- a/assets/javascripts/discourse/connectors/full-page-search-below-search-header/semantic-search.gjs
+++ b/assets/javascripts/discourse/connectors/full-page-search-below-search-header/semantic-search.gjs
@@ -124,7 +124,7 @@ export default class SemanticSearch extends Component {
       this.preventAISearch = false;
     }
 
-    if (this.initialSearchTerm) {
+    if (this.initialSearchTerm && !this.searching) {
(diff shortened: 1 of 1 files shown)


pmusaraj (approved):

pmusaraj:
> onAppEvent("full-page-search:trigger-search") is triggered at minimum two times when a search is queried, sometimes three

I wonder whether we should also fix this upstream in core so that the event is only triggered once per search query...

keegangeorge:
> > onAppEvent("full-page-search:trigger-search") is triggered at minimum two times when a search is queried, sometimes three
> 
> I wonder whether we should also fix this upstream in core so that the event is only triggered once per search query...

Yes, I think it would be good if we do. Though on a quick look, I'm not sure why its being triggered more than once as its only called once? Will have to do more digging.

PR #424, opened 2024-01-15
DEV: Update test setup to work with auto groups

Drenmi:
### What is this change?

We're updating core to change TL based access settings to be group based. This requires some updates of tests to work correctly. (The existing test setup gives false positives.)

Related: https://github.com/discourse/discourse/pull/24257


discourse-ai
davidtaylorhq, line 1: Don't think this symlink should've been included?

davidtaylorhq (changes requested):

davidtaylorhq (approved):

discourse-ai
Drenmi: MemeMe

PR #425, opened 2024-01-15
FIX: Clean unicode usernames when adding messages through prompt's constructor

romanrizzi:


xfalcox (approved):

SamSaffron:
I think this is the wrong place for this concern, open ai dialect should take care of this and also just preserve it else wise, by changing content to `#{id}: content`

SamSaffron:
I followed up, see my refactors today

PR #436, opened 2024-01-19
UX: Validations to LLM-backed features (except AI Bot)

romanrizzi:
This change is part of an ongoing effort to prevent enabling a broken feature due to lack of configuration. We also want to explicit which provider we are going to use. For example, Claude models are available through AWS Bedrock and Anthropic, but the configuration differs.

Validations are:

* You must choose a model before enabling the feature.
* You must turn off the feature before setting the model to blank.
* You must configure each model settings before being able to select it.

Screenshot 2024-01-19 at 12 21 48

---

Screenshot 2024-01-19 at 12 07 54

---

Screenshot 2024-01-19 at 12 08 19


lib/completions/llm.rb
```diff
@@ -17,27 +17,55 @@ module Completions
...
-      def self.with_prepared_responses(responses)
-        @canned_response = DiscourseAi::Completions::Endpoints::CannedResponse.new(responses)
+      class << self
+        def models_by_provider
+          # ChatGPT models are listed under open_ai but they are actually available through OpenAI and Azure.
+          # However, since they use the same URL/key settings, there's no reason to duplicate them.
+          {
+            aws_bedrock: %w[claude-instant-1 claude-2],
+            anthropic: %w[claude-instant-1 claude-2],
+            vllm: %w[mistralai/Mixtral-8x7B-Instruct-v0.1 mistralai/Mistral-7B-Instruct-v0.2],
```
xfalcox, line 27: Technically we can support 

- vLLM + Llama2
- vLLM + StableBeluga2
- vLLM + Upstage Llama2

xfalcox (approved):

PR #447, opened 2024-01-30
FEATURE: allow excluding tags and categories from LLM report

SamSaffron:
Also

- Better diagnostics, output model being used
- Prompt LLM that true content is being injected in <context> tag

### DIFF
--- a/lib/automation/report_context_generator.rb
+++ b/lib/automation/report_context_generator.rb
@@ -50,7 +50,7 @@ module DiscourseAi
           ) if exclude_category_ids.present?
 
         if exclude_tags.present?
-          exclude_tag_ids = Tag.where(name: exclude_tags).select(:id)
+          exclude_tag_ids = Tag.where_name(exclude_tags).select(:id)
           @posts =
             @posts.where(
               "topics.id NOT IN (?)",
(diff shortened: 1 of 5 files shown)


lib/automation/report_context_generator.rb
```diff
@@ -41,6 +43,20 @@ def initialize(
...
         @posts = @posts.where("categories.read_restricted = ?", false) if !@allow_secure_categories
         @posts = @posts.where("categories.id IN (?)", @category_ids) if @category_ids.present?
+        @posts =
+          @posts.where(
+            "categories.id NOT IN (?)",
+            exclude_category_ids,
+          ) if exclude_category_ids.present?
+
+        if exclude_tags.present?
+          exclude_tag_ids = Tag.where(name: exclude_tags).select(:id)
```
martin-brennan, line 53: Minor, but I think the `where_name` scope should probably be used for this:

https://github.com/discourse/discourse/blob/74edd165a87b861591d40d62d21be5f7b635392c/app/models/tag.rb#L24-L28

martin-brennan (approved):
LGTM, just one minor comment :+1:

lib/automation/report_context_generator.rb
SamSaffron: oh yeah it can work, will change  ... fwiw we have an index on lower(name) ... so yay

PR #453, opened 2024-01-31
DEV: Increase embeddings backfill job frequency

xfalcox:
The idea is to increase the frequency so we can run with smaller batch sizes.
Big batches cause problems when running backups, so it's better to have shorter but
more frequent jobs.


romanrizzi (approved):
Should we decrease the 50k limit?

xfalcox:
> Should we decrease the 50k limit?

Since we ship with a 250 default value, and it's  a hidden setting, I think we should leave it as it will only impact people who are deep into the weeds.

PR #457, opened 2024-02-01
UX: Re-introduce embedding settings validations

romanrizzi:


lib/configuration/embeddings_model_validator.rb
```diff
@@ -0,0 +1,47 @@
...
+module DiscourseAi
+  module Configuration
+    class EmbeddingsModelValidator
+      def initialize(opts = {})
+        @opts = opts
+      end
+
+      def valid_value?(val)
+        # Bypass default validation. Enabling the feature will take care of this.
+        return true if val == "bge-large-en"
```
xfalcox, line 12: Do we need this? Do we run validation for defaults?

xfalcox (approved):

PR #466, opened 2024-02-09
FEATURE: mentionable personas and random picker tool, context limits

SamSaffron:
This rather large PR adds support for the following: 

image

image

1. Personas are now **optionally** mentionable, meaning that you can mention them either from public topics or PMs
    1. Mentioning from PMs helps "switch" persona mid conversation, meaning if you want to look up sites setting you can invoke the site setting bot, or if you want to generate an image you can invoke dall e
    2. Mentioning outside of PMs allows you to inject a bot reply in a topic trivially
2. We also add the support for max_context_posts this allow you to limit the amount of context you feed in, which can help control costs

### DIFF
--- a/app/controllers/discourse_ai/admin/ai_personas_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_personas_controller.rb
@@ -3,7 +3,7 @@
 module DiscourseAi
   module Admin
     class AiPersonasController < ::Admin::AdminController
-      before_action :find_ai_persona, only: %i[show update destroy]
+      before_action :find_ai_persona, only: %i[show update destroy create_user]
 
       def index
         ai_personas =
@@ -30,6 +30,11 @@ module DiscourseAi
         else
           render_json_error ai_persona
         end
+      end
+
+      def create_user
+        user = @ai_persona.create_user!
+        render json: BasicUserSerializer.new(user, root: "user")
       end
 
       def update
@@ -64,6 +69,9 @@ module DiscourseAi
             :priority,
             :top_p,
             :temperature,
+            :default_llm,
+            :user_id,
+            :mentionable,
             allowed_group_ids: [],
           )
 
--- a/app/controllers/discourse_ai/admin/ai_personas_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_personas_controller.rb
@@ -76,6 +76,7 @@ module DiscourseAi
             :default_llm,
             :user_id,
             :mentionable,
+            :max_context_posts,
             allowed_group_ids: [],
           )
 
--- a/app/controllers/discourse_ai/admin/ai_personas_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_personas_controller.rb
@@ -16,7 +16,11 @@ module DiscourseAi
           DiscourseAi::AiBot::Personas::Persona.all_available_tools.map do |tool|
             AiToolSerializer.new(tool, root: false)
           end
-        render json: { ai_personas: ai_personas, meta: { commands: tools } }
+        llms =
+          DiscourseAi::Configuration::LlmEnumerator.values.map do |hash|
+            { id: hash[:value], name: hash[:name] }
+          end
+        render json: { ai_personas: ai_personas, meta: { commands: tools, llms: llms } }
       end
 
       def show
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -86,6 +86,7 @@ class AiPersona < ActiveRecord::Base
     allowed_group_ids = self.allowed_group_ids
     id = self.id
     system = self.system
+    user_id = self.user_id
 
     persona_class = DiscourseAi::AiBot::Personas::Persona.system_personas_by_id[self.id]
     if persona_class
@@ -142,6 +143,10 @@ class AiPersona < ActiveRecord::Base
 
       define_singleton_method :name do
         name
+      end
+
+      define_singleton_method :user_id do
+        user_id
       end
 
       define_singleton_method :description do
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -232,8 +232,27 @@ class AiPersona < ActiveRecord::Base
     end
   end
 
+  FIRST_PERSONA_USER_ID = -1200
+
   def create_user!
     raise "User already exists" if user_id && User.exists?(user_id)
+
+    # find the first id smaller than FIRST_USER_ID that is not taken
+    id = nil
+
+    id = DB.query_single(<<~SQL, FIRST_PERSONA_USER_ID, FIRST_PERSONA_USER_ID - 200).first
+        WITH seq AS (
+          SELECT generate_series(?, ?, -1) AS id
+          )
+        SELECT seq.id FROM seq
+        LEFT JOIN users ON users.id = seq.id
+        WHERE users.id IS NULL
+        ORDER BY seq.id DESC
+      SQL
+
+    id = DB.query_single(<<~SQL).first if id.nil?
+        SELECT min(id) - 1 FROM users
+      SQL
 
     # note .invalid is a reserved TLD which will route nowhere
     user =
@@ -244,6 +263,7 @@ class AiPersona < ActiveRecord::Base
         active: true,
         approved: true,
         trust_level: TrustLevel[4],
+        id: id,
       )
     user.save!(validate: false)
 
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -197,7 +197,7 @@ class AiPersona < ActiveRecord::Base
   end
 
   def create_user!
-    raise "User already exists" if user_id
+    raise "User already exists" if user_id && User.exists?(user_id)
 
     # note .invalid is a reserved TLD which will route nowhere
     user =
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -64,9 +64,15 @@ class AiPersona < ActiveRecord::Base
       .where(mentionable: true)
       .where(enabled: true)
       .joins(:user)
-      .pluck("users.username, allowed_group_ids")
-      .map do |username, allowed_group_ids|
-        { username: username, allowed_group_ids: allowed_group_ids }
+      .pluck("ai_personas.id, users.id, lower(users.username), allowed_group_ids, default_llm")
+      .map do |id, user_id, username, allowed_group_ids, default_llm|
+        {
+          id: id,
+          user_id: user_id,
+          username: username,
+          allowed_group_ids: allowed_group_ids,
+          default_llm: default_llm,
+        }
       end
   end
 
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -65,7 +65,7 @@ class AiPersona < ActiveRecord::Base
       .where(mentionable: true)
       .where(enabled: true)
       .joins(:user)
-      .pluck("ai_personas.id, users.id, lower(users.username), allowed_group_ids, default_llm")
+      .pluck("ai_personas.id, users.id, users.username_lower, allowed_group_ids, default_llm")
       .map do |id, user_id, username, allowed_group_ids, default_llm|
         {
           id: id,
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -199,15 +199,14 @@ class AiPersona < ActiveRecord::Base
   def create_user!
     raise "User already exists" if user_id
 
+    # note .invalid is a reserved TLD which will route nowhere
     user =
       User.new(
-        email: "no_email_#{name}",
+        email: "no_email_#{name}@does-not-exist.invalid",
         name: name.titleize,
-        username: UserNameSuggester.suggest(name),
+        username: UserNameSuggester.suggest(name + "_bot"),
         active: true,
         approved: true,
-        admin: true,
-        moderator: true,
         trust_level: TrustLevel[4],
       )
     user.save!(validate: false)
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -9,6 +9,9 @@ class AiPersona < ActiveRecord::Base
   validates :system_prompt, presence: true, length: { maximum: 10_000_000 }
   validate :system_persona_unchangeable, on: :update, if: :system
 
+  belongs_to :created_by, class_name: "User"
+  belongs_to :user
+
   before_destroy :ensure_not_system
 
   class MultisiteHash
@@ -54,6 +57,17 @@ class AiPersona < ActiveRecord::Base
       .all
       .limit(MAX_PERSONAS_PER_SITE)
       .map(&:class_instance)
+  end
+
+  def self.mentionables
+    persona_cache[:mentionable_usernames] ||= AiPersona
+      .where(mentionable: true)
+      .where(enabled: true)
+      .joins(:user)
+      .pluck("users.username, allowed_group_ids")
+      .map do |username, allowed_group_ids|
+        { username: username, allowed_group_ids: allowed_group_ids }
+      end
   end
 
   after_commit :bump_cache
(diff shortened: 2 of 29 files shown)


app/models/ai_persona.rb
```diff
@@ -171,6 +232,25 @@ def class_instance
     end
   end
 
+  def create_user!
+    raise "User already exists" if user_id && User.exists?(user_id)
```
jjaffeux, line 236:
```suggestion
    raise "User already exists" if User.exists?(user_id)
```

`User.exists?(nil) # false`

SamSaffron: looks correct as a change but feels a bit too magical to me, I like the extra deliberate code in this specific case.

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -103,26 +123,42 @@ export default class PersonaEditor extends Component {
...
-        await this.args.model.update({ enabled: this.args.model.enabled });
-      } catch (e) {
-        popupAjaxError(e);
-      }
-    }
+    this.toggleField("enabled");
   }
 
   @action
   async togglePriority() {
```
jjaffeux, line 130: what's the goal of async here?

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -103,26 +123,42 @@ export default class PersonaEditor extends Component {
...
-    this.args.model.set("priority", !this.args.model.priority);
-    this.editingModel.set("priority", this.args.model.priority);
+    this.toggleField("priority", true);
+  }
+
+  @action
+  async toggleMentionable() {
```
jjaffeux, line 135: what's the goal of async here?

spec/requests/admin/ai_personas_controller_spec.rb
```diff
@@ -17,21 +17,40 @@
       )
     end
 
+    it "sideloads llms" do
+      get "/admin/plugins/discourse-ai/ai_personas.json"
```
jjaffeux, line 21: I thought the convention would be `/.../.../ai-personas.json`

assets/javascripts/discourse/components/ai-persona-editor.gjs
SamSaffron: I guess cause toggleField is async and I wanted to signal that... I think maybe I should await here which at least will clarify this.

assets/javascripts/discourse/admin/models/ai-persona.js
```diff
@@ -45,6 +51,18 @@ export default class AiPersona extends RestModel {
     this.commands = properties.commands;
   }
 
+  async createUser() {
+    const result = await ajax(
+      `/admin/plugins/discourse-ai/ai-personas/${this.id}/create_user.json`,
```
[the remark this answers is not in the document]
SamSaffron: will change to create-user but I like these long paths they feel much clearer

lib/ai_bot/tools/random_picker.rb
```diff
@@ -0,0 +1,73 @@
...
+                  "An array where each element is either a range (e.g., '1-6') or a comma-separated list of options (e.g., 'sam,jane,joe')",
+                type: "array",
+                item_type: "string",
+                required: true,
+              },
+            ],
+          }
+        end
+
+        def self.name
```
jjaffeux, line 25: That's just me, but I think it's always risky to override name in ruby. Given name is defined on modules. It can lead to oddities, probably fine here.

```
module Test
end

Test.name # "Test"
```

jjaffeux (approved):
Didn't see anything block-worthy 👍 Might be good to wait for @martin-brennan if he has time.

assets/javascripts/discourse/components/ai-persona-editor.gjs
jjaffeux: If you don't await, I don't think it awaits, it will just return a promise that's all

assets/javascripts/discourse/admin/models/ai-persona.js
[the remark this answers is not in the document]
jjaffeux: Yes, sorry I just meant to change the _ to -, the rest is totally fine

lib/ai_bot/tools/random_picker.rb
SamSaffron: hmm interesting, will be quite an involved change cause we would need to change all tools, and then have lack of parity with #description ... will have a think.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -32,6 +36,11 @@ def create
         end
       end
 
+      def create_user
+        user = @ai_persona.create_user!
```
martin-brennan, line 40: I wonder if we should make these into actual non-human (negative ID) users? I know we have no system yet for "reserving" these in core so it does make things trickier. Maybe we could start at `-10_000` for these AI bot users and count down from there, using a postgres sequence could make this easier:

```
CREATE SEQUENCE ai_bot_user_ids AS integer INCREMENT BY -1 START WITH -10001;
SELECT nextval('ai_bot_user_ids');
=> -10001
```

Maybe this is too complex, but I am just thinking it may get confusing to have regular users floating around that are _really_ bot users.

app/models/ai_persona.rb
```diff
@@ -56,6 +60,23 @@ def self.all_personas
       .map(&:class_instance)
   end
 
+  def self.mentionables
+    persona_cache[:mentionable_usernames] ||= AiPersona
+      .where(mentionable: true)
+      .where(enabled: true)
+      .joins(:user)
+      .pluck("ai_personas.id, users.id, lower(users.username), allowed_group_ids, default_llm")
```
martin-brennan, line 68: We already store `username_lower` on users:

```suggestion
      .pluck("ai_personas.id, users.id, users.username_lower, allowed_group_ids, default_llm")
```

config/locales/client.en.yml
```diff
@@ -108,6 +108,16 @@ en:
       ai_persona:
         name: Name
         description: Description
+        no_llm_selected: "No language model selected"
+        max_context_posts: "Max Context Posts"
+        max_context_posts_help: "The maximum number of posts to use as context for the AI when responding to a user. (empty for default)"
```
martin-brennan, line 113: Could be helpful to mention the default is 40, the only problem then is that you have to keep the translation and the server const in sync :D

app/models/ai_persona.rb
```diff
@@ -171,6 +232,25 @@ def class_instance
     end
   end
 
+  def create_user!
+    raise "User already exists" if user_id && User.exists?(user_id)
+
+    # note .invalid is a reserved TLD which will route nowhere
+    user =
+      User.new(
+        email: "#{SecureRandom.hex}@does-not-exist.invalid",
```
martin-brennan, line 241: We do this in a few places in core, would be cool to have a `User.generate_fake_email` method in core. No need to change here, just saying.

assets/javascripts/discourse/admin/models/ai-persona.js
```diff
@@ -45,6 +51,18 @@ export default class AiPersona extends RestModel {
...
+  async createUser() {
+    const result = await ajax(
+      `/admin/plugins/discourse-ai/ai-personas/${this.id}/create-user.json`,
+      {
+        type: "POST",
+      }
+    );
+    this.user = result["user"];
+    this.user_id = result["user"]["id"];
+    return result["user"];
```
martin-brennan, line 63:
```suggestion
    this.user = result.user;
    this.user_id = this.user.id;
    return this.user;
```

nit

lib/ai_bot/playground.rb
```diff
@@ -3,26 +3,101 @@
...
+          schedule_bot_reply(post)
         end
       end
 
       def conversation_context(post)
         # Pay attention to the `post_number <= ?` here.
         # We want to inject the last post as context because they are translated differently.
+
+        # also setting default to 40, allowing huge contexts costs lots of tokens
+        max_posts = 40
```
martin-brennan, line 93:
```suggestion
        max_posts = DEFAULT_MAX_CONTEXT_POSTS
```

Probably should have a const for this

lib/ai_bot/playground.rb
```diff
@@ -3,26 +3,101 @@
...
+        if bot.persona.class.respond_to?(:max_context_posts)
+          max_posts = bot.persona.class.max_context_posts || 40
```
martin-brennan, line 95:
```suggestion
          max_posts = bot.persona.class.max_context_posts || DEFAULT_MAX_CONTEXT_POSTS
```

lib/ai_bot/playground.rb
```diff
@@ -96,6 +171,9 @@ def reply_to(post)
         reply = +""
         start = Time.now
 
+        post_type =
+          post.post_type == Post.types[:whisper] ? Post.types[:whisper] : Post.types[:regular]
```
martin-brennan, line 175:
```suggestion
          post.whisper? Post.types[:whisper] : Post.types[:regular]
```

lib/ai_bot/playground.rb
```diff
@@ -3,26 +3,101 @@
...
         # We want to inject the last post as context because they are translated differently.
+
+        # also setting default to 40, allowing huge contexts costs lots of tokens
+        max_posts = 40
+        if bot.persona.class.respond_to?(:max_context_posts)
+          max_posts = bot.persona.class.max_context_posts || 40
+        end
+
+        post_types = [Post.types[:regular]]
+        post_types << Post.types[:whisper] if post.post_type == Post.types[:whisper]
```
martin-brennan, line 99:
```suggestion
        post_types << Post.types[:whisper] if post.whisper?
```

lib/ai_bot/playground.rb
```diff
@@ -106,50 +184,84 @@ def reply_to(post)
...
+        reply_user = bot.bot_user
+        if bot.persona.class.respond_to?(:user_id)
+          reply_user = User.find_by(id: bot.persona.class.user_id) || reply_user
+        end
+
+        stream_reply = post.topic.private_message?
+
+        # we need to ensure persona user is allowed to reply to the pm
+        if post.topic.private_message?
+          if !post.topic.topic_allowed_users.where(user_id: reply_user.id).exists?
```
martin-brennan, line 196:
```suggestion
          if !post.topic.topic_allowed_users.exists?(user_id: reply_user.id)
```

lib/ai_bot/tools/random_picker.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+
+        def options
+          parameters[:options]
+        end
+
+        def invoke(_bot_user, _llm)
+          result = nil
+          # can be a naive list of strings
+          if options.all? { |option| !option.match?(/\A\d+-\d+\z/) && !option.include?(",") }
```
martin-brennan, line 36:
```suggestion
          if options.none? { |option| option.match?(/\A\d+-\d+\z/) || option.include?(",") }
```

Nitpicky, but I think it reads better if we do the opposite. Goes from "any none of these" to "none of either of these"

spec/requests/admin/ai_personas_controller_spec.rb
martin-brennan: Yeah I agree URLs should have `-` not `_`, but this is probably annoying to change now.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -32,6 +36,11 @@ def create
         end
       end
 
+      def create_user
```
martin-brennan, line 39: Probably a good idea to rate limit this. Less dangerous since it's admin-only, but still.

config/locales/client.en.yml
martin-brennan: We might be able to pass it down via the `Site` serializer and then use that to pass to the translation/the placeholder here:

image

martin-brennan (commented):
From a UI perspective, these list items with "(not configured)" are a bit confusing. What happens if I choose one? Is there some validation to prevent me from doing it? Would be nice if select kit supports it (not sure if it does) to disable these items in the list at the least:

image

Also, when I toggled on Mentionable then didn't choose the "Default Language Model", I got a "Persona saved!" toast, but the mentionable toggle silently went back to false. IMO we should show a validation error if the user toggles on Mentionable but doesn't choose a "Default Language Model".

Also, even if I choose a language model, saving the page always toggles Mentionable back to false, so there is a bug here somewhere.

The actual mentioning and switching worked nicely :) The "let's welcome them to the community" message will not display with bot users, so if we change to using a bot that would be good. Also the list of avatars in the top left of the PM doesn't change until the page refreshes, not a huge deal:

image

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: yeah something about "flagging bots" is a bit of a dark science ... I can at a minimum pick a negative id out nowhere ... will do that for now

app/models/ai_persona.rb
SamSaffron: yeah agree... the secure random hex is such a tricky thing to come up with and it is non obvious... plus I needed to wikipedia to find out about `.invalid`

config/locales/client.en.yml
SamSaffron: hmm tricky... yeah maybe this should be a site setting? I can see people who want to save money set it lower ... bot probably operates fine at 10 context and you would save a big pile of money, especially considering giant convos @tgxworld tends to have :)

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: hmmm yeah I am not sure actually, you could just create personas non stop anyway, maybe default rate limiting is enough.

config/locales/client.en.yml
martin-brennan: Yeah site setting could be fine as well, at least as the **default**, because then you can use that on the client too for the input and translation.

app/models/ai_persona.rb
martin-brennan: Yeah it's not obvious what `.invalid` does unless you've already had to deal with it

app/controllers/discourse_ai/admin/ai_personas_controller.rb
martin-brennan: Ah yeah true I forget about the defaults

SamSaffron:
> Also, when I toggled on Mentionable then didn't choose the "Default Language Model", I got a "Persona saved!" toast, but the mentionable toggle silently went back to false. IMO we should show a validation error if the user toggles on Mentionable but doesn't choose a "Default Language Model". 

was a bug... all sorted

lib/ai_bot/tools/random_picker.rb
SamSaffron: agree, reads better

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: I moved to using negative numbers in a reserved range, should be much better now... not ideal given there is no range registration but better for sure.

app/models/ai_persona.rb
```diff
@@ -232,9 +232,28 @@ def class_instance
...
 
+  FIRST_PERSONA_USER_ID = -1200
+
   def create_user!
     raise "User already exists" if user_id && User.exists?(user_id)
 
+    # find the first id smaller than FIRST_USER_ID that is not taken
+    id = nil
+
+    id = DB.query_single(<<~SQL, FIRST_PERSONA_USER_ID, FIRST_PERSONA_USER_ID - 200).first
```
martin-brennan, line 243: I guess we don't worry about this for now...but what if > 200 persona users are made :thinking: Might be good to make a site-wide persona count limit to deal with this. Anyway, not holding up this current PR :+1:

app/models/ai_persona.rb
```diff
@@ -232,9 +232,28 @@ def class_instance
...
+        WITH seq AS (
+          SELECT generate_series(?, ?, -1) AS id
```
martin-brennan, line 245: This is nice, TIL

app/models/ai_persona.rb
SamSaffron: oh there is a fallback afterwards just in case... we take minimum ... very unlikely to hit it, but it is handled.

PR #472, opened 2024-02-14
DEV: Make prompts available on `CurrentUserSerializer`

keegangeorge:
Instead of making a request to `/discourse-ai/prompts.json` to fetch the list of available prompts for the AI Helper, we ship the prompts automatically with the `CurrentUserSerializer` so its available without any delay. Prompts are also cached.

### DIFF
--- a/lib/ai_helper/assistant.rb
+++ b/lib/ai_helper/assistant.rb
@@ -6,37 +6,34 @@ module DiscourseAi
       AI_HELPER_PROMPTS_CACHE_KEY = "ai_helper_prompts"
 
       def available_prompts
-        prompts = Discourse.cache.read(AI_HELPER_PROMPTS_CACHE_KEY)
+        Discourse
+          .cache
+          .fetch(AI_HELPER_PROMPTS_CACHE_KEY) do
+            prompts = CompletionPrompt.where(enabled: true)
 
-        if !prompts
-          prompts = CompletionPrompt.where(enabled: true)
+            # Hide illustrate_post if disabled
+            prompts =
+              prompts.where.not(
+                name: "illustrate_post",
+              ) if SiteSetting.ai_helper_illustrate_post_model == "disabled"
 
-          # Hide illustrate_post if disabled
-          prompts =
-            prompts.where.not(
-              name: "illustrate_post",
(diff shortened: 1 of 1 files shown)


lib/ai_helper/entry_point.rb
```diff
@@ -19,6 +19,20 @@ def inject_into(plugin)
...
+        plugin.add_to_serializer(
+          :current_user,
+          :ai_helper_prompts,
+          include_condition: -> do
+            SiteSetting.composer_ai_helper_enabled && scope.authenticated? &&
+              scope.user.in_any_groups?(SiteSetting.ai_helper_allowed_groups_map)
+          end,
+        ) do
+          ActiveModel::ArraySerializer.new(
+            DiscourseAi::AiHelper::Assistant.new.available_prompts,
```
SamSaffron, line 32: this is great for starters, but I think we can eliminate this query to the database with a cache, feels kind of important cause there is no need to run through all of this work on every page load for users with access.

SamSaffron (approved):

lib/ai_helper/entry_point.rb
SamSaffron: not blocking can happen in a followup pr.

SamSaffron:
we are fiddling with the current user serializer, so we should at least ensure the code is covered and that it is rendered in a spec.  I think simplest is to make a get request to current user and then you have it all in JSON format

lib/ai_helper/assistant.rb
```diff
@@ -3,35 +3,40 @@
 module DiscourseAi
   module AiHelper
     class Assistant
-      def available_prompts(name_filter: nil)
-        cp = CompletionPrompt
-        prompts = []
+      AI_HELPER_PROMPTS_CACHE_KEY = "ai_helper_prompts"
+
+      def available_prompts
+        prompts = Discourse.cache.read(AI_HELPER_PROMPTS_CACHE_KEY)
```
xfalcox, line 9: This pattern of `return if it's on cache, otherwise fetch it, is already implemented by Discourse.cache.fetch

romanrizzi (approved):

PR #473, opened 2024-02-16
FEATURE: new Discourse Helper persona

SamSaffron:
This persona searches Discourse Meta for help with Discourse and
points users at relevant posts.

It is somewhat similar to using "Forum Helper" on meta, with the
notable difference that we can not lean on semantic search so using
some prompt engineering we try to keep it simple.

Overall this works quite well and may even field some support for
us.


xfalcox (approved):

SamSaffron:
@xfalcox we could just provide a cheap semantic search endpoint and do the hyde in the tool on the remote server

PR #474, opened 2024-02-18
DEV: Make more group-based settings client: false

lis2:
Affects the following settings:

ai_toxicity_groups_bypass
ai_helper_allowed_groups
ai_helper_custom_prompts_allowed_groups
post_ai_helper_allowed_groups

This turns off client: true for these group-based settings, because there is no guarantee that the current user gets all their group memberships serialized to the client. Better to check server-side first.


martin-brennan (approved):
Looks great, ty!

PR #476, opened 2024-02-19
FEATURE: Handle secure uploads in image caption

xfalcox:


romanrizzi (approved):

app/controllers/discourse_ai/ai_helper/assistant_controller.rb
```diff
@@ -109,11 +109,15 @@ def caption_image
         image_url = params[:image_url]
         raise Discourse::InvalidParameters.new(:image_url) if !image_url
 
-        image = Upload.where(url: params[:image_url])
+        if SiteSetting.secure_uploads?
```
pmusaraj, line 112: This setting has gotten a little bit more complicated, it's no longer an all or nothing, there's now `secure_uploads_allow_embed_images_in_emails` and `secure_uploads_pm_only`. So, you can't tell if an upload is secure from just checking that setting now. 

Can we pull the upload record here and check `.secure?`?

pmusaraj (approved):

PR #479, opened 2024-02-20
FIX: unable to share conversations with persona user

SamSaffron:
Persona users are still bots, but we were not properly accounting
for it and share icon was not showing up.

This depends on a core change that adds .topic to transformed posts

### DIFF
--- a/assets/javascripts/initializers/ai-bot-replies.js
+++ b/assets/javascripts/initializers/ai-bot-replies.js
@@ -143,6 +143,8 @@ function initializePersonaDecorator(api) {
   );
 }
 
+const MAX_PERSONA_USER_ID = -1200;
+
 function initializeShareButton(api) {
   const currentUser = api.getCurrentUser();
   if (!currentUser || !currentUser.ai_enabled_chat_bots) {
@@ -159,13 +161,20 @@ function initializeShareButton(api) {
   };
 
   api.addPostMenuButton("share", (post) => {
-    // very hacky and ugly, but there is no `.topic` in attrs
+    // for backwards compat so we don't break if topic is undefined
+    if (post.topic?.archetype !== "private_message") {
+      return;
+    }
+
     if (
       !currentUser.ai_enabled_chat_bots.any(
         (bot) => post.username === bot.username
       )
     ) {
-      return;
+      // special handling for personas (larger than -1200 means real user)
+      if (post.user_id > MAX_PERSONA_USER_ID) {
+        return;
+      }
     }
 
     return {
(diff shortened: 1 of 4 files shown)


assets/javascripts/initializers/ai-bot-replies.js
```diff
@@ -159,13 +161,20 @@ function initializeShareButton(api) {
...
+      return;
+    }
+
     if (
       !currentUser.ai_enabled_chat_bots.any(
         (bot) => post.username === bot.username
       )
     ) {
-      return;
+      // special handling for personas (larger than -1200 means real user)
```
martin-brennan, line 174:
```suggestion
      // special handling for personas (persona bot users start at ID -1200 and go down)
```

minor nit, but I don't think calling the ones higher than -1200 real users is accurate, because they are still bots too...just not persona bots :P

spec/system/ai_bot/share_spec.rb
```diff
@@ -35,11 +35,56 @@
...
+    Fabricate(:post, topic: pm, user: admin, raw: "How do I do stuff?")
+    Fabricate(:post, topic: pm, user: persona.user, raw: "No idea")
+
+    visit(pm.url)
+
+    # clipboard functionality is extremely hard to test
+    # we would need special permissions in chrome driver to enable full access
+    # instead we use a secret variable to signal that we want to store clipboard
+    # data in window.discourseAiClipboard
+    page.execute_script("window.discourseAiTestClipboard = true")
```
martin-brennan, line 55: We already have this sorted in core, nothing fancy needed. We have a special "CDP" page object:


```ruby
let(:cdp) { PageObjects::CDP.new }
```

Which is the chrome dev tools comms protocol. Once you know something is on the clipboard, all you have to do is:

```ruby
expect(cdp.read_clipboard).to eq(clip_text)
```

We have several core and chat examples e.g. https://github.com/discourse/discourse/blob/b057f1b2b4b534c6618e8c19ea32166722a9ee01/spec/system/post_selection_copy_quote_spec.rb . Sometimes you need to use `.chomp` too

martin-brennan: You can also use this CDP thing to e.g. limit network speed, we have a couple of tests that slow down the network for uploads for example. It can do tons of other stuff too

SamSaffron: will try cdp, I tried it previously and had no luck, but will give it a shot again.

martin-brennan: Oh weird, yeah definitely LMK if it's not working and I can check out locally. It's working fine over many tests in core + chat

SamSaffron: all working now... just pushed a re-implementation.

martin-brennan (approved):

PR #488, opened 2024-02-27
FEATURE: support Chat with AI Persona via a DM 

SamSaffron:
Add support for chat with AI personas

- Allow enabling chat for AI personas that have an associated user
- Add new setting `allow_chat` to AI persona to enable/disable chat
- When a message is created in a DM channel with an allowed AI persona user, schedule a reply job
- AI replies to chat messages using the persona's `max_context_posts` setting to determine context
- Store tool calls and custom prompts used to generate a chat reply on the `ChatMessageCustomPrompt` table
- Add tests for AI chat replies with tools and context

At the moment unlike posts we do not carry tool calls in the context.

No @mention support yet for ai personas in channels, this is future work


app/models/message_custom_prompt.rb
```diff
@@ -0,0 +1,20 @@
...
+#  id            :bigint           not null, primary key
+#  message_id    :bigint           not null
+#  custom_prompt :json             not null
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+# Indexes
+#
+#  index_message_custom_prompts_on_message_id  (message_id) UNIQUE
+#
```
jjaffeux, line 20: Maybe we prefix with `chat_` here. To simplify things in future?

`self.table_name = "chat_message_custom_prompts"`

SamSaffron: I can rename no problem ... just tried to leave this even with PostCustomPrompt

app/jobs/regular/create_ai_chat_reply.rb
```diff
@@ -0,0 +1,25 @@
...
+module ::Jobs
+  class CreateAiChatReply < ::Jobs::Base
+    sidekiq_options retry: false
+
+    def execute(args)
+      channel = ::Chat::Channel.find_by(id: args[:channel_id])
+      return if channel.blank?
+
+      message = ::Chat::Message.find_by(id: args[:message_id])
+      return if message.blank?
```
jjaffeux, line 12:
```suggestion
      message = ::Chat::Message.find_by(id: args[:message_id], chat_channel_id: args[:channel_id])
      return if message.blank?
```

lib/ai_bot/playground.rb
```diff
@@ -184,25 +206,153 @@ def title_playground(post)
...
+        #  ChatSDK::Thread.messages(
+        #    channel_id: channel.id,
+        #    thread_id: message.thread_id,
+        #    guardian: guardian,
+        #  )
+
+        max_messages = 40
+        if bot.persona.class.respond_to?(:max_context_posts)
+          max_messages = bot.persona.class.max_context_posts || 40
+        end
```
jjaffeux, line 233:
```suggestion
        max_messages = bot.persona.class&.max_context_posts || 40
```

SamSaffron: no worky for cases like this, though our internal design can be improved `"test"&.not_there` errors

jjaffeux:
I  gave this a try locally sam, few notes.

1 - I got this error in logs when starting to discuss:

```
/Users/joffreyjaffeux/Code/discourse-ai/app/jobs/regular/create_ai_chat_reply.rb:18:in `execute'
/Users/joffreyjaffeux/Code/pr-discourse/app/jobs/base.rb:305:in `block (2 levels) in perform' 
rails_multisite-6.0.0/lib/rails_multisite/connection_management/null_instance.rb:49:in `with_connection'
rails_multisite-6.0.0/lib/rails_multisite/connection_management.rb:21:in `with_connection'
```

I figured out it was related to me not choosing a `Default Language Model` on my persona, but we should probably show an error if you save without having one? Bare minimum the error in logs should be clearer.

2 - After fixing default language model, I tried to discuss again and got this:

```
activerecord-7.0.8.1/lib/active_record/core.rb:284:in `find' 
/Users/joffreyjaffeux/Code/discourse-ai/app/jobs/regular/create_ai_chat_reply.rb:20:in `execute'
```

Looking at my DB I see this:

Screenshot 2024-05-03 at 12 33 14

And the query was actually looking for `gpt-4-turbo` at id `-113`, I suspect than when I changed my default llm, the new user for this llm has not been created? I changed my default language model to `gpt-3.5-turbo` and that fixed it. But something doesn't look right here.

SamSaffron:
Yikes will add some validations on Monday

SamSaffron:
both are fixed @jjaffeux

jjaffeux (approved):

PR #495, opened 2024-02-28
DEV: improve internal design of ai persona and bug fix

SamSaffron:
- Fixes bug where OpenAI could not describe images
- Fixes bug where mentionable personas could not be mentioned unless overarching bot was enabled
- Improves internal design of playground and bot to allow better for non "bot" users
- Allow PMs directly to persona users (previously bot user would also have to be in PM)
- Simplify internal code
- Drops feature of low_cost model which is better implemented as a tool param

### DIFF
--- a/app/models/ai_persona.rb
+++ b/app/models/ai_persona.rb
@@ -78,9 +78,7 @@ class AiPersona < ActiveRecord::Base
         end
 
     if user
-      all_mentionables.select do |mentionable|
-        user.in_any_groups?(mentionable[:allowed_group_ids])
-      end
+      all_mentionables.select { |mentionable| user.in_any_groups?(mentionable[:allowed_group_ids]) }
     else
       all_mentionables
     end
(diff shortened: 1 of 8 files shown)


app/models/ai_persona.rb
```diff
@@ -60,21 +60,30 @@ def self.all_personas
...
+            user_id: user_id,
+            username: username,
+            allowed_group_ids: allowed_group_ids,
+            default_llm: default_llm,
+          }
+        end
+
+    if user
+      all_mentionables.select do |mentionable|
+        (user.group_ids & mentionable[:allowed_group_ids]).present?
```
martin-brennan, line 82:
```suggestion
        user.in_any_groups?(mentionable[:allowed_group_ids])
```

martin-brennan (approved):
Looks fine from a code POV to me (minor nit only), and I assume you know what you are doing with this refactor :)

app/models/ai_persona.rb
martin-brennan: image

PR #504, opened 2024-03-03
fix stuff

SamSaffron:
- **FIX: only update system attributes when updating system persona**
- **FIX: update participant count by hand so bot messages show in inbox**

### DIFF
--- a/lib/ai_bot/playground.rb
+++ b/lib/ai_bot/playground.rb
@@ -276,7 +276,7 @@ module DiscourseAi
         # since we are skipping validations and jobs we
         # may need to fix participant count
         if reply_post.topic.private_message? && reply_post.topic.participant_count < 2
-          reply_post.topic.update(participant_count: 2)
+          reply_post.topic.update!(participant_count: 2)
         end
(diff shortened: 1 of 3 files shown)


lib/ai_bot/playground.rb
```diff
@@ -273,6 +273,12 @@ def reply_to(post)
           reply_post.post_custom_prompt.update!(custom_prompt: prompt)
         end
 
+        # since we are skipping validations and jobs we
+        # may need to fix participant count
+        if reply_post.topic.private_message? && reply_post.topic.participant_count < 2
+          reply_post.topic.update(participant_count: 2)
```
jjaffeux, line 279:
```suggestion
          reply_post.topic.update!(participant_count: 2)
```

We generally prefer this

jjaffeux (approved):

lib/ai_bot/playground.rb
SamSaffron: worried about exploding the bot on this... but I guess its ok.

PR #509, opened 2024-03-05
FIX: Correct max_length calculation to prevent potential token overfl…

wozulong:
FIX: Correct max_length calculation to prevent potential token overflow in embeddings model.

The current max_length calculation was too simplistic, potentially causing the actual token count to surpass the embeddings model's maximum limit. This was particularly noticeable in complex Unicode content. The fix addresses this issue by accurately determining the token count, ensuring compatibility with a wide range of text inputs.

PR #515, opened 2024-03-07
FIX: Handle unicode on tokenizer

xfalcox:
Our fast track code broke when strings had characters who are longer in tokens than in UTF-8.

Follow-up to #509 with tests added

### DIFF
--- a/lib/tokenizer/basic_tokenizer.rb
+++ b/lib/tokenizer/basic_tokenizer.rb
@@ -17,10 +17,20 @@ module DiscourseAi
         end
 
         def truncate(text, max_length)
+          # fast track common case, /2 to handle unicode chars
+          # than can take more than 1 token per char
+          return text if !SiteSetting.ai_strict_token_counting && text.size < max_length / 2
+
           tokenizer.decode(tokenizer.encode(text).ids.take(max_length))
         end
 
         def can_expand_tokens?(text, addition, max_length)
+          # fast track common case, /2 to handle unicode chars
+          # than can take more than 1 token per char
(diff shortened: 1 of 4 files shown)


lib/tokenizer/basic_tokenizer.rb
```diff
@@ -17,15 +17,10 @@ def size(text)
         end
 
         def truncate(text, max_length)
-          # Fast track the common case where the text is already short enough.
-          return text if text.size < max_length
```
SamSaffron, line 21: maybe we can still fast track for  `text.size < max_length / 2` , I think it looks pretty safe, looking at examples?

SamSaffron: gives some could be 3 tokens long or maybe for ... we can shortcut on / 3 extremely safely for now. I worry about all the extra work for ultra short strings.

wozulong: > gives some could be 3 tokens long or maybe for ... we can shortcut on / 3 extremely safely for now. I worry about all the extra work for ultra short strings.

I'm concerned that there might be some special cases causing exceptions, considering that the token count calculation is not simply based on character length. Perhaps we could use environment variables to configure whether this acceleration is necessary?

SamSaffron: I follow, maybe we just merge this for now and see

xfalcox: @wozulong `DISCOURSE_AI_STRICT_TOKEN_COUNTING: true` will the ENV to disable acceleration.

romanrizzi (approved):

PR #521, opened 2024-03-11
FEATURE: Share conversations with AI via a URL

SamSaffron:
This allows users to share a static page of an AI conversation with
the rest of the world.

A follow on will be a custom onebox in Discourse to display this
nicely.

By default this feature is disabled, it is enabled by turning on
ai_bot_allow_public_sharing via site settings

Precautions are taken when sharing

1. We make a carbonite copy
2. We minimize work generating page
3. We limit to 100 interactions
4. Many security checks - including disallowing if there is a mix
of users in the PM.


lib/ai_bot/entry_point.rb
```diff
@@ -56,6 +56,29 @@ def self.map_bot_model_to_user_id(model_name)
         end
       end
 
+      def self.ai_share_error(topic, guardian)
```
martin-brennan, line 59: I feel like this would work much better in a guardian extension...something like 

```ruby
def can_share_ai_conversation?(target)
  // checks go here
end
```

Otherwise this code is a bit weird and not very standard with other things we do. Chat uses a guardian extension here:

https://github.com/discourse/discourse/blob/446e4620a8373396f3979819ad9c7b1e45e5e802/plugins/chat/plugin.rb#L62-L67

https://github.com/discourse/discourse/blob/446e4620a8373396f3979819ad9c7b1e45e5e802/plugins/chat/lib/chat/guardian_extensions.rb

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      def ensure_allowed_preview!
+        @topic = Topic.find_by(id: params[:topic_id])
+        raise Discourse::NotFound if !@topic
+
+        @shared_conversation = SharedAiConversation.find_by(topic_id: params[:topic_id])
+
+        @error = DiscourseAi::AiBot::EntryPoint.ai_share_error(@topic, guardian)
+        if @error == :not_allowed
+          raise Discourse::InvalidAccess.new(
```
martin-brennan, line 87: You can just pass the message as the first param like this:

```ruby
Discourse::InvalidAccess.new(I18n.t("discourse_ai.share_ai.errors.#{@error}"))
```

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+                  nil,
+                  custom_message: "discourse_ai.share_ai.errors.#{@error}",
+                )
+        end
+      end
+
+      def ensure_allowed_destroy!
+        @shared_conversation = SharedAiConversation.find_by(share_key: params[:share_key])
+
+        raise Discourse::InvalidAccess.new if @shared_conversation.blank?
```
martin-brennan, line 98:
```suggestion
        raise Discourse::InvalidAccess if @shared_conversation.blank?
```

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+
+        if @shared_conversation.user_id != current_user.id && !current_user.admin?
+          raise Discourse::InvalidAccess.new
```
martin-brennan, line 101:
```suggestion
          raise Discourse::InvalidAccess
```

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+                )
+        end
+      end
+
+      def ensure_allowed_destroy!
+        @shared_conversation = SharedAiConversation.find_by(share_key: params[:share_key])
+
+        raise Discourse::InvalidAccess.new if @shared_conversation.blank?
+
+        if @shared_conversation.user_id != current_user.id && !current_user.admin?
```
martin-brennan, line 100: Another thing that should be in a guardian extension `can_see_shared_conversation?(conv)`

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+          raise Discourse::InvalidAccess.new
+        end
+      end
+
+      def ensure_allowed_create!
+        @topic = Topic.find_by(id: params[:topic_id])
+        error = DiscourseAi::AiBot::EntryPoint.ai_share_error(@topic, guardian)
```
martin-brennan, line 107: Same comments about this being guardian, and also InvalidAccess passing a string as the first param for the message.

assets/javascripts/discourse/components/modal/share-full-topic-modal.gjs
```diff
@@ -0,0 +1,123 @@
...
+    } catch (e) {
+      popupAjaxError(e);
+    }
+  }
+
+  @action
+  async share() {
+    try {
+      await clipboardCopyAsync(this.generateShareURL.bind(this));
+      this.justCopiedText = I18n.t("discourse_ai.ai_bot.conversation_shared");
```
martin-brennan, line 79: No need for this (or discourseLater here), we can just use the `toasts` service:

```javascript
this.toasts.success({
  duration: 3000,
  data: {
    message: I18n.t("discourse_ai.ai_bot.conversation_shared"),
  },
});
```

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+      @id = post[:id]
+      @user_id = post[:user_id]
+      @created_at = DateTime.parse(post[:created_at])
+      @cooked = post[:cooked]
+    end
+  end
+
+  def populated_posts
+    return @populated_posts if @populated_posts
+    @populated_posts = posts.map { |post| SharedPost.new(post.symbolize_keys) }
```
martin-brennan, line 40: Since the params needed to share the post are quite small, might be better to just pass the ones you need:

```ruby
SharedPost.new(id: post.id, user_id: post.user_id, created_at: post.created_at, cooked: post.cooked)
```

config/routes.rb
```diff
@@ -19,6 +19,13 @@
     post "post/:post_id/stop-streaming" => "bot#stop_streaming_response"
     get "bot-username" => "bot#show_bot_username"
   end
+
+  scope module: :ai_bot, path: "/ai-bot/shared-ai-conversations" do
+    post "/" => "shared_ai_conversations#create"
+    delete "/:share_key" => "shared_ai_conversations#destroy"
+    get "/:share_key" => "shared_ai_conversations#show"
```
martin-brennan, line 26: Bit of a nitpick, but the URL is super long:

```
http://localhost:4200/discourse-ai/ai-bot/shared-ai-conversations/PHBmFKwiEOMd1g2p7NRbAg
```

Could be cool to have a short share URL:

```
http://localhost:4200/discourse-ai/shared/PHBmFKwiEOMd1g2p7NRbAg

OR

http://localhost:4200/discourse-ai/convo/PHBmFKwiEOMd1g2p7NRbAg
```

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+    populate_user_info!(@populated_posts)
+    @populated_posts
+  end
+
+  def self.excerpt(posts)
+    excerpt = +""
+    posts.each do |post|
+      excerpt << "#{post.user.username}: #{post.excerpt(100)} "
```
martin-brennan, line 48: Should use `post.user.display_name`, since that takes into account `SiteSetting.prioritize_username_in_ux`

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+      break if excerpt.length > 1000
+    end
+    excerpt
+  end
+
+  def formatted_excerpt
+    "AI Conversation with #{llm_name}:\n #{excerpt}"
```
martin-brennan, line 55: Should have I18n :)

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def self.build_conversation_data(topic, max_posts: DEFAULT_MAX_POSTS, include_usernames: false)
+    llm_name = nil
+    topic.topic_allowed_users.each do |tu|
+      if DiscourseAi::AiBot::EntryPoint::BOT_USER_IDS.include?(tu.user_id)
+        _, _, llm_name =
```
martin-brennan, line 62: I find these lines super hard to read with all the morse code `_`... might be nice to have something like this in `EntryPoint`:

```ruby
Bot = Struct.new(:id, :name, :llm)

def self.find_bot_by_id(id)
  found = DiscourseAi::AiBot::EntryPoint::BOTS.find { |bot| bot[0] == id }
  return if !found
  Bot.new(found[0], found[1], found[2])
end
```

Then you can change this to:

```ruby
llm_name = DiscourseAi::AiBot::EntryPoint.find_bot_by_id(tu.user_id)&.llm
```

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+          DiscourseAi::AiBot::EntryPoint::BOTS.find { |user_id, _, _| user_id == tu.user_id }
+        break
+      end
+    end
+
+    llm_name = ActiveSupport::Inflector.humanize(llm_name) if llm_name
+    llm_name ||= "unknown AI model"
```
martin-brennan, line 69: Since we are showing this in the UI this should have I18n

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    posts =
+      topic
+        .posts
+        .by_post_number
+        .where(post_type: Post.types[:regular])
+        .where.not(cooked: nil)
+        .where(deleted_at: nil)
```
martin-brennan, line 77: Shouldn't be necessary since the default posts scope doesn't include deleted posts

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+      excerpt: excerpt(posts),
+      posts:
+        posts.map do |post|
+          mapped = {
+            id: post.id,
+            user_id: post.user_id,
+            created_at: post.created_at,
+            cooked: post.cooked,
+          }
+          mapped[:username] = post.user&.username if include_usernames
```
martin-brennan, line 92: I was a bit confused by this at first, but now I understand -- this is so we don't bake the usernames into the data in the DB right?

martin-brennan: Also this should be:

```suggestion
          mapped[:username] = post.user&.display_name if include_usernames
```

assets/javascripts/discourse/components/modal/share-full-topic-modal.gjs
martin-brennan: You would need to import it and do `@service toasts`

assets/javascripts/discourse/components/modal/share-modal.gjs
```diff
@@ -104,6 +116,13 @@ export default class ShareModal extends Component {
           @label="discourse_ai.ai_bot.share_modal.copy"
         />
         <span class="ai-share-modal__just-copied">{{this.justCopiedText}}</span>
+        {{#if this.siteSettings.ai_bot_allow_public_sharing}}
```
martin-brennan, line 119: If the site setting becomes a group one this would be `this.currentUser.can_share_ai_bot_conversations`

assets/javascripts/initializers/ai-bot-replies.js
```diff
@@ -145,6 +146,28 @@ function initializeShareButton(api) {
...
 
+function initializeShareTopicButton(api) {
+  const siteSettings = api.container.lookup("site-settings:main");
+  api.registerTopicFooterButton({
+    id: "share-ai-conversation",
+    icon: "share",
+    label: "discourse_ai.ai_bot.share_ai_conversation.name",
+    title: "discourse_ai.ai_bot.share_ai_conversation.title",
+    action() {
+      const modal = api.container.lookup("service:modal");
```
martin-brennan, line 157: Should do this lookup in the same place as `siteSettings` above, otherwise you have to look it up every time you click the button

db/migrate/20240309034751_add_shared_ai_conversations.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddSharedAiConversations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :shared_ai_conversations do |t|
+      t.integer :user_id, null: false
+      t.integer :topic_id, null: false
```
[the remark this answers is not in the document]
martin-brennan: I know this might be a pain, but it will be much more painful to do this later from experience

public/ai-share/highlight.min.css
martin-brennan, line 1: Hmm we already have highlight JS in core, I don't think we need to re-include it in AI. I am not sure how you would use it exactly on the static page, I think David would know.

public/ai-share/share.css
```diff
@@ -0,0 +1,194 @@
...
+  --primary-color: #007bff;
+  --secondary-color: #6c757d;
+  --background-color: #f8f9fa;
+  --post-background-color: #ffffff;
+  --details-background-color: #e9ecef;
+  --details-hover-color: #dee2e6;
+}
+
+body {
+  font-family: Söhne, ui-sans-serif, system-ui, -apple-system, Segoe UI, Roboto, Ubuntu, Cantarell, Noto Sans, sans-serif, Helvetica Neue, Arial, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
```
martin-brennan, line 11: Why have so many fonts here?

martin-brennan: Should probably get designer Kris to take a pass with this CSS and page

martin-brennan (commented):
It's a big one :D

martin-brennan:
Oh @SamSaffron this also doesn't work with secure uploads, which will especially be a problem on meta where we have them enabled in PMs:

image

I think when you are doing `build_conversation_data` you will need to dig into `post.cooked` and do...something with these images/uploads.

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def populate_user_info!(posts)
+    users = User.where(id: posts.map(&:user_id).uniq).map { |u| [u.id, u] }.to_h
+    posts.each { |post| post.user = users[post.user_id] }
+  end
+
+  def generate_share_key
+    self.share_key = SecureRandom.urlsafe_base64(16)
+  end
+end
```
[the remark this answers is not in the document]
SamSaffron: doing it ! thanks

db/migrate/20240309034751_add_shared_ai_conversations.rb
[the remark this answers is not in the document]
SamSaffron: I thought about this, I think I can safely migrate it once ready ... chat is going to need a lot of new ui here anyway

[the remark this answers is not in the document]
SamSaffron: but will do so ... sure not too hard

app/models/shared_ai_conversation.rb
SamSaffron: not sure about this ... maybe we want both name and username depending on prefs ... its a tricky bit of rendering.

public/ai-share/share.css
SamSaffron: yeah going to hand this back to Kris tomorrow, for sure.

SamSaffron:
going to do dall e shares in a follow up ... its a tough one ... for sure, but also looks so cool.

assets/javascripts/initializers/ai-bot-replies.js
SamSaffron: true will change

public/ai-share/highlight.min.css
SamSaffron: it was so complicated ... @davidtaylorhq can help but embedding on a static page is pretty tough cause we have it as a module... so I need to redo all the module logic to get it going.

db/migrate/20240309034751_add_shared_ai_conversations.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.string :title, null: false, max_length: 1024
+      t.string :llm_name, null: false, max_length: 1024
+      t.jsonb :posts, null: false
```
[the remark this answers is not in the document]
SamSaffron: true ... will change

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+  module AiBot
+    class SharedAiConversationsController < ::ApplicationController
+      requires_plugin ::DiscourseAi::PLUGIN_NAME
+      requires_login only: %i[create update destroy]
+      before_action :ensure_allowed_create!, only: %i[create]
+      before_action :ensure_allowed_destroy!, only: %i[destroy]
+      before_action :ensure_allowed_preview!, only: %i[preview]
+      before_action :require_site_settings!
+
+      skip_before_action :preload_json, :check_xhr, :redirect_to_login_if_required, only: %i[show]
```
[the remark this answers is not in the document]
SamSaffron: agree! for sure

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module AiBot
+    class SharedAiConversationsController < ::ApplicationController
+      requires_plugin ::DiscourseAi::PLUGIN_NAME
+      requires_login only: %i[create update destroy]
+      before_action :ensure_allowed_create!, only: %i[create]
```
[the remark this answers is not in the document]
SamSaffron: cool will move it

martin-brennan:
@SamSaffron just found another UI issue, we are not showing the human error if you are in a PM with extra people:

image

SamSaffron:
Oh yeah ... let me sort this out!

app/models/shared_ai_conversation.rb
```diff
@@ -4,26 +4,31 @@ class SharedAiConversation < ActiveRecord::Base
...
+  validates :context, presence: true
   validates :share_key, presence: true, uniqueness: true
 
   before_validation :generate_share_key, on: :create
 
-  def self.share_conversation(user, topic, max_posts: DEFAULT_MAX_POSTS)
-    conversation = find_by(user: user, topic: topic)
-    conversation_data = build_conversation_data(topic, max_posts: max_posts)
+  def self.share_conversation(user, target, max_posts: DEFAULT_MAX_POSTS)
+    raise "Target must be a topic for now" if !target.is_a?(Topic)
```
martin-brennan, line 17: Yes I think this is fine to do for now along with the SharedPost thing, we just want at least the DB columns to be flexible from the start :+1:

SamSaffron:
image

Getting  human error, but I think if an admin walks into another PM and tries to share it they get an opaque error

martin-brennan (approved):
:rocket:

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,201 @@
...
+
+  def url
+    "#{Discourse.base_uri}/discourse-ai/ai-bot/shared-ai-conversations/#{share_key}"
+  end
+
+  def html_excerpt
+    html = +""
+    populated_context.each do |post|
+      text =
+        PrettyText.excerpt(
```
martin-brennan, line 72: You can just do `post.excerpt` here too like you do lower in `self.excerpt`; it accepts the same arguments as using PrettyText.excerpt directly:

https://github.com/discourse/discourse/blob/6435b790abc932cb99ceb81ceca7df0d15186f7d/app/models/post.rb#L521-L524

SamSaffron: diff post object :)

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,201 @@
...
+          post.cooked,
+          400,
+          text_entities: true,
+          strip_links: true,
+          strip_details: true,
+        )
+
+      html << "<p><b>#{post.user.username}</b>: #{text}</p>"
+      if html.length > 1000
```
martin-brennan, line 81: Rather than doing this, could you just call `PrettyText.excerpt(html, 1000)`?

app/models/shared_ai_conversation.rb
```diff
@@ -0,0 +1,201 @@
...
+    html << "<a href='#{url}'>#{I18n.t("discourse_ai.share_ai.read_more")}<a>"
+    html
+  end
+
+  def onebox
+    <<~HTML
+    <div>
+      <aside class="onebox allowlistedgeneric" data-onebox-src="#{url}">
+      <header class="source">
+        <span class="onebox-ai-llm-title">#{I18n.t("discourse_ai.share_ai.onebox_title", llm_name: llm_name)}</span>
```
martin-brennan, line 95: Should just move this into a `onebox_title` method like `formatted_excerpt`

app/models/shared_ai_conversation.rb
SamSaffron: its tricky cause I am getting up to N chars per post ... so this is not exactly what I want... its a "fancy excerpt"

app/models/shared_ai_conversation.rb
SamSaffron: maybe we can extract to a method, but in the share page we have a lot more room so can format differently

PR #526, opened 2024-03-12
FIX: don't show share conversation incorrectly

SamSaffron:
- ai_persona_name can be null vs undefined leading to button showing up where it should not
- do not allow sharing of conversations where user is sending PMs to self

### DIFF
--- a/lib/guardian_extensions.rb
+++ b/lib/guardian_extensions.rb
@@ -25,12 +25,11 @@ module DiscourseAi
         return false if allowed_user_ids.any? { |id| id > 0 && id != user.id }
 
         # no bot in the PM
-        bots = DiscourseAi::AiBot::EntryPoint.all_bot_ids
-        return false if allowed_user_ids.none? { |id| bots.include?(id) }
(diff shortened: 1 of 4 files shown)


lib/guardian_extensions.rb
```diff
@@ -17,15 +17,22 @@ def can_share_ai_bot_conversation?(target)
...
+        return false if allowed_user_ids.any? { |id| id > 0 && id != user.id }
+
+        # no bot in the PM
+        bot_ids = DiscourseAi::AiBot::EntryPoint.all_bot_ids
+        return false if allowed_user_ids.none? { |id| bot_ids.include?(id) }
 
         # other content in PM
         return false if target.posts.where("user_id > 0 and user_id <> ?", user.id).exists?
+      else
+        return false
```
SamSaffron, line 35: hmmm bug maybe ... looking

SamSaffron: oh no bug ... it just drifted from top ... all good

lis2 (approved):

martin-brennan (approved):

PR #527, opened 2024-03-12
FIX: Filter soft-deleted topics when backfilling sentiment

romanrizzi:
The query included soft-deleted topics, causing errors when doing `post.topic`. The `Trashable` module default scope wasn't included by default.


martin-brennan (approved):

martin-brennan:
Oh I approved but might want to add a spec too @romanrizzi

PR #528, opened 2024-03-13
FIX: expire assets when CSS changes FIX: missing translation on share page

SamSaffron:
Also add trivial image styling

### DIFF
--- a/app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
+++ b/app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
@@ -32,6 +32,10 @@ module DiscourseAi
       end
 
       def show
+        # we don't use an asset packager, so remember to bump this
+        # when CSS changes
+        @asset_version = 1
+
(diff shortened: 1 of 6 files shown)


featheredtoast (approved):

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
```diff
@@ -32,6 +32,10 @@ def destroy
       end
 
       def show
+        # we don't use an asset packager, so remember to bump this
+        # when CSS changes
+        @asset_version = 1
```
martin-brennan, line 37: Can you just use DiscourseAi::AiBot::SharedAiConversationsHelper::VERSION here?

martin-brennan (approved):

app/controllers/discourse_ai/ai_bot/shared_ai_conversations_controller.rb
SamSaffron: oh was left over code, removed it! :hugs:

PR #531, opened 2024-03-14
UX: Add title suffix to shared AI pages

davidtaylorhq:
Previously, the title of the shared AI page would be the title of the PM. When shared somewhere else (e.g. another Discourse site, or social media platform), this can make it look like an authoritative article published by the forum owner.

This commit adds an `- AI Conversation` suffix to the page title.

This serves a few purposes:

1. Increases awareness of discourse-ai's bot feature

2. Bring external sharing in line with the internal-onebox for this feature, which already has a clear 'AI conversation' label

3. Makes it clearer that these pages are not an authoritative source, and are not necessarily endorsed by the forum owner

This is especially important in situations where the og:description is not displayed (e.g. in Discourse inline oneboxes)

### DIFF
--- a/app/views/discourse_ai/ai_bot/shared_ai_conversations/show.html.erb
+++ b/app/views/discourse_ai/ai_bot/shared_ai_conversations/show.html.erb
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title><%= @shared_conversation.title %></title>
-    <meta property="og:title" content="<%= @shared_conversation.title %>">
+    <title><%= I18n.t("discourse_ai.share_ai.title", title: @shared_conversation.title, site_name: SiteSetting.title) %></title>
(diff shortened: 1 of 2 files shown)


davidtaylorhq:
AI approves of this PR ;)

https://meta.discourse.org/discourse-ai/ai-bot/shared-ai-conversations/GV22iLHXu9_qsJXfeiIlVA

SamSaffron:
I hear you but I am a bit on the fence, totally get the problem ... maybe what we want is a suffix vs a prefix. We do label the pages quite carefully already and do not index them.

davidtaylorhq:
We do label pages reasonably well once you visit them. But I think it would be good wor people to have that context before clicking-through. A suffix could work :+1:

Something like

> Some Title Goes Here - AI Conversation - Discourse Meta

That would be in-line with our page titles for regular topics, which go "{title} - {category} - {site name}"

ZogStriP (approved):

PR #537, opened 2024-03-19
FEATURE: AI Bot RAG support.

romanrizzi:
This PR lets you associate uploads to an AI persona, which we'll split and generate embeddings from. When building the system prompt to get a bot reply, we'll do a similarity search followed by a re-ranking (if available). This will let us find the most relevant fragments from the body of knowledge you associated with the persona, resulting in better, more informed responses.

For now, we'll only allow plain-text files, but this will change in the future.

### DIFF
--- a/lib/ai_bot/personas/persona.rb
+++ b/lib/ai_bot/personas/persona.rb
@@ -128,6 +128,7 @@ module DiscourseAi
               <<~TEXT.strip,
             #{system_insts}
             #{available_tools.map(&:custom_system_message).compact_blank.join("\n")}
+            #{rag_fragments_prompt(context[:conversation_context].to_a)}
           TEXT
               messages: context[:conversation_context].to_a,
               topic_id: context[:topic_id],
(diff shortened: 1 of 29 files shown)


lib/ai_bot/personas/persona.rb
```diff
@@ -171,6 +172,45 @@ def find_tool(parsed_function)
...
+          strategy = DiscourseAi::Embeddings::Strategies::Truncation.new
+          vector_rep =
+            DiscourseAi::Embeddings::VectorRepresentations::Base.current_representation(strategy)
+
+          interactions_vector = vector_rep.vector_from(latest_interactions)
+
+          candidate_fragment_ids =
+            vector_rep.asymmetric_rag_fragment_similarity_search(
+              interactions_vector,
+              limit: 10,
```
xfalcox, line 196: We can take more here, like 50 or 100, and pass to our reranker model and then limit to less.

lib/ai_bot/personas/persona.rb
```diff
@@ -171,6 +176,68 @@ def find_tool(parsed_function)
...
+              limit: 50,
+              offset: 0,
+            )
+
+          guidance =
+            RagDocumentFragment.where(upload_id: upload_refs, id: candidate_fragment_ids).pluck(
+              :fragment,
+            )
+
+          if DiscourseAi::Inference::HuggingFaceTextEmbeddings.reranker_configured?
```
xfalcox, line 213: Do we want to test for this a few lines before and only get 10 lines when this isn't available?

lib/ai_bot/personas/persona.rb
```diff
@@ -171,6 +176,68 @@ def find_tool(parsed_function)
             persona_options: options[tool_klass].to_h,
           )
         end
+
+        def rag_fragments_prompt(conversation_context)
+          upload_refs =
+            UploadReference.where(target_id: id, target_type: "AiPersona").pluck(:upload_id)
```
xfalcox, line 182: `id` is null for me here, is this supposed to be `@persona.id` here and elsewhere in this method?

The `id` method you defined on line 87 only works for `system_personas`.

xfalcox (approved):

PR #538, opened 2024-03-20
FIX: mention suppression was not working right

SamSaffron:
We were only suppressing non mentions, ones that become spans.

@sam in the test was not resolving to a mention cause the user
did not exist.

depends on: https://github.com/discourse/discourse/pull/26253

For tests to pass.

### DIFF
--- a/lib/automation/report_runner.rb
+++ b/lib/automation/report_runner.rb
@@ -237,6 +237,10 @@ Follow the provided writing composition instructions carefully and precisely ste
         parsed
           .css("a")
           .each do |a|
+            if a["class"] == "mention"
+              a.inner_html = a.inner_html.sub("@", "")
+              next
+            end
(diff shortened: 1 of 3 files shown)


tgxworld (approved):
LGTM 👍

tgxworld:
I'm wondering if we should consider adding a `.discourse-compatibility` entry here because this fix depends on another fix in core. So if this plugin is deployed on an older version of core, this fix you added here will not work.

SamSaffron:
I don't think we need compatibility at all, the fix in core is a major edge case that is very likely not to hit.

Eg, you click "suppress mentions" in the report, @bsam is mentioned ... remapped to `bsam` then our mention code incorrectly thinks sam was mentioned and is notified... given so few use the report and even fewer use the suppression I think we are good.

PR #543, opened 2024-03-21
FIX: Fetch categories data using specific endpoint

nbianca:
It used to fetch it from /site.json, but /categories.json is the more appropriate one. This one also implements pagination, so we have to do one request per page.


xfalcox (approved):

SamSaffron:
this looks fine @nbianca can we merge?

nbianca:
I realized that this might not work when lazy_load_categories is disabled because any page will just return all categories, leading to an infinite loop. That is fixed with https://github.com/discourse/discourse/pull/26421.

The other thing that I do not particularly like is that it will request one empty page to stop loading.

xfalcox (approved):

PR #544, opened 2024-03-22
FIX: Avoid replying to the reply user for llm_triage automation

nattsw:
The "Triage posts using AI" automation allows admins to set a "Reply User" to represent the generated response.

When the "Reply User" is not a bot and creates a reply, the automation will be triggered again. This causes an infinite conversation from the reply_user to the reply_user.

This PR ignores posts from the "Reply User"

Here's an idea of what the full automation looks like (no change):
Screenshot 2024-03-22 at 12 09 47 PM

### DIFF
--- a/discourse_automation/llm_triage.rb
+++ b/discourse_automation/llm_triage.rb
@@ -36,6 +36,13 @@ if defined?(DiscourseAutomation)
 
     script do |context, fields|
       post = context["post"]
+      canned_reply = fields.dig("canned_reply", "value")
+      canned_reply_user = fields.dig("canned_reply_user", "value")
+
+      # nothing to do if we already replied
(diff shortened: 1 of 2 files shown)


nattsw:
LOL lint

discourse_automation/llm_triage.rb
```diff
@@ -52,6 +52,8 @@
       canned_reply = fields.dig("canned_reply", "value")
       canned_reply_user = fields.dig("canned_reply_user", "value")
 
+      next if post.user.username == canned_reply_user
```
tgxworld, line 55: Micro optimisation here but we can move this up the sequence in the block to return earlier instead of doing alot of work to gather the variables which may be thrown away.

tgxworld (approved):

PR #548, opened 2024-03-28
FEATURE: web browsing tool

SamSaffron:
This pull request makes several improvements and additions to the GitHub-related tools and personas in the `discourse-ai` repository:

1. It adds the `WebBrowser` tool to the  `Researcher` persona, allowing the AI to visit web pages, retrieve HTML content, extract the main content, and convert it to plain text.

2. It updates the `GithubFileContent`, `GithubPullRequestDiff`, and `GithubSearchCode` tools to handle HTTP responses more robustly (introducing size limits). 

3. It refactors the `send_http_request` method in the `Tool` class to follow redirects when specified, and to read the response body in chunks to avoid memory issues with large responses. (only for WebBrowser)

4. It updates the system prompt for the `Researcher` persona to provide more detailed guidance on when to use Google search vs web browsing, and how to optimize tool usage and reduce redundant requests.

5. It adds a new `web_browser_spec.rb` file with tests for the `WebBrowser` tool, covering various scenarios like handling different HTML structures and following redirects.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -209,6 +209,7 @@ en:
             name: "Base Search Query"
             description: "Base query to use when searching. Example: '#urgent' will prepend '#urgent' to the search query and only include topics with the urgent category or tag."
       command_summary:
+        web_browser: "Browse Web"
         github_search_code: "GitHub code search"
         github_file_content: "GitHub file content"
         github_pull_request_diff: "GitHub pull request diff"
@@ -227,6 +228,7 @@ en:
         dall_e: "Generate image"
         search_meta_discourse: "Search Meta Discourse"
       command_help:
+        web_browser: "Browse web page using the AI Bot"
         github_search_code: "Search for code in a GitHub repository"
         github_file_content: "Retrieve content of files from a GitHub repository"
         github_pull_request_diff: "Retrieve a GitHub pull request diff"
@@ -245,6 +247,7 @@ en:
         dall_e: "Generate image using DALL-E 3"
         search_meta_discourse: "Search Meta Discourse"
       command_description:
+        web_browser: "Reading <a href='%{url}'>%{url}</a>"
         github_search_code: "Searched for '%{query}' in %{repo}"
         github_pull_request_diff: "<a href='%{url}'>%{repo} %{pull_id}</a>"
         github_file_content: "Retrieved content of %{file_paths} from %{repo_name}@%{branch}"
--- a/lib/ai_bot/tools/tool.rb
+++ b/lib/ai_bot/tools/tool.rb
@@ -81,10 +81,17 @@ module DiscourseAi
           raise "Expecting caller to use a block" if !block_given?
 
           uri = nil
+          url = UrlHelper.normalized_encode(url)
+          uri =
+            begin
+              URI.parse(url)
+            rescue StandardError
+              nil
+            end
+
+          return if !uri
 
           if follow_redirects
-            url = UrlHelper.normalized_encode(url)
-            URI.parse(url) # ensure url parses, will raise if not
             fd =
               FinalDestination.new(
                 url,
@@ -94,13 +101,6 @@ module DiscourseAi
               )
 
             uri = fd.resolve
-          else
-            uri =
-              begin
-                URI.parse(url)
-              rescue StandardError
-                nil
-              end
           end
 
           return if uri.blank?
--- a/lib/ai_bot/tools/tool.rb
+++ b/lib/ai_bot/tools/tool.rb
@@ -77,8 +77,34 @@ module DiscourseAi
 
         protected
 
-        def send_http_request(url, headers: {}, authenticate_github: false)
-          uri = URI(url)
+        def send_http_request(url, headers: {}, authenticate_github: false, follow_redirects: false)
+          raise "Expecting caller to use a block" if !block_given?
+
+          uri = nil
+
+          if follow_redirects
+            url = UrlHelper.normalized_encode(url)
+            URI.parse(url) # ensure url parses, will raise if not
+            fd =
+              FinalDestination.new(
+                url,
+                validate_uri: true,
+                max_redirects: 5,
+                follow_canonical: true,
+              )
+
+            uri = fd.resolve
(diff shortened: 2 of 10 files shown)


config/locales/server.en.yml
```diff
@@ -245,6 +247,7 @@ en:
         dall_e: "Generate image using DALL-E 3"
         search_meta_discourse: "Search Meta Discourse"
       command_description:
+        web_browser: "Reading <a href='%{url}'>%{url}</a>"
```
martin-brennan, line 250: Do we need any of those other attrs like `target="_blank"`, `rel="nofollow"` etc? Just a thought, I don't know either.

lib/ai_bot/tools/web_browser.rb
```diff
@@ -0,0 +1,114 @@
...
+
+module DiscourseAi
+  module AiBot
+    module Tools
+      class WebBrowser < Tool
+        def self.signature
+          {
+            name: name,
+            description:
+              "Visits a web page, retrieves the HTML content, extracts the main content, converts it to plain text, and returns the result.",
```
martin-brennan, line 11: I can't remember, are these texts only sent to the LLM? If not and they are shown to the user then we should I18n them.

lib/ai_bot/tools/web_browser.rb
```diff
@@ -0,0 +1,114 @@
...
+        def description_args
+          { url: url }
+        end
+
+        private
+
+        def extract_main_content(html)
+          doc = Nokogiri.HTML(html)
+
+          # Remove script, style, and comments
```
martin-brennan, line 59: I don't think any of these comments are really necessary, it's fairly obvious what is being done.

lib/ai_bot/tools/web_browser.rb
```diff
@@ -0,0 +1,114 @@
...
+
+          # Find the main content container
+          main_content = find_main_content(doc)
+
+          # If main content not found, fallback to the body
+          main_content ||= doc.at("body")
+
+          # Extract text from the main content
+          buffer = +""
+          nodes_to_text(main_content, buffer)
```
martin-brennan, line 70: This is interesting, I would have thought you needed to do this for `buffer` to be updated?

```ruby
buffer = nodes_to_text(main_content, buffer)
```

So the original string is modified inside the method?

lib/ai_bot/tools/tool.rb
```diff
@@ -77,8 +77,34 @@ def standalone?
...
-        def send_http_request(url, headers: {}, authenticate_github: false)
-          uri = URI(url)
+        def send_http_request(url, headers: {}, authenticate_github: false, follow_redirects: false)
+          raise "Expecting caller to use a block" if !block_given?
+
+          uri = nil
+
+          if follow_redirects
+            url = UrlHelper.normalized_encode(url)
+            URI.parse(url) # ensure url parses, will raise if not
```
martin-brennan, line 87: Why raise here for `URI.parse` and `rescue StandardError` below if not following redirects?

lib/ai_bot/tools/tool.rb
```diff
@@ -77,8 +77,34 @@ def standalone?
 
         protected
 
-        def send_http_request(url, headers: {}, authenticate_github: false)
-          uri = URI(url)
+        def send_http_request(url, headers: {}, authenticate_github: false, follow_redirects: false)
```
martin-brennan, line 80: Still feels really clunky to have `authenticate_github` as a top-level keyword arg here, IMO as I was saying in the previous review the github tools should just pass in the `Authorization` header themselves as part of `headers`

martin-brennan (approved):
I have comments/questions but I don't think they are blockers. The prompt for this new tool is nice and detailed :+1:

config/locales/server.en.yml
SamSaffron: Oh I think we get it automatically when it is cooked

lib/ai_bot/tools/web_browser.rb
SamSaffron: for llm only :)

lib/ai_bot/tools/web_browser.rb
SamSaffron: lol ... partially generated ... will nuke

lib/ai_bot/tools/web_browser.rb
SamSaffron: yes it I like the efficiency of just appending to a buffer, so we need to pass around, otherwise so many string copies ...

lib/ai_bot/tools/tool.rb
SamSaffron: hmmm I am not sure maybe I just return null and then it give no response back to caller and it gets an unknown to the llm

lib/ai_bot/tools/tool.rb
SamSaffron: yeah but Its so much ceremony in 3 tools, kind of saves a bunch for them I hear you though.

lib/ai_bot/tools/tool.rb
SamSaffron: fixed

SamSaffron:
Note... one concern is ... web browsing is going to be expensive we are dumping tons of content on the context so stuff can add up. regardless though ... the usefulness is so high and you have to start by configuring google anyway.

lib/ai_bot/tools/tool.rb
SamSaffron: oh ... we are cooking as admin so we get follows :( will add nofollows to a few places ...

config/locales/server.en.yml
SamSaffron: rabbit hole warning ... we cook as TL3... so everything the AI says gets follow ... I guess its a general framework fix we need. https://github.com/discourse/discourse-ai/blob/c22e5e66c50559508bee81b30d447ab526609395/lib/ai_bot/playground.rb#L268-L282 no way to tell this at the moment that we want it to nofollow and it is tied to the bot user which is tl3 :( going to track this on dev.

PR #554, opened 2024-04-04
FIX: Secure/unsecure uploads when sharing AI conversations

martin-brennan:
This commit uses a new plugin modifier introduced in https://github.com/discourse/discourse/pull/26508
to mark all uploads as _not_ secure in shared PM AI conversations.
This is so images created by the AI bot (or uploaded by the user)
do not end up as broken URLs because of the security requirements
around them.

This relies on the TopicUploadSecurityManager in core as well,
which is fired when an AI conversation is shared or deleted.

**Relies on the core PR mentioned above to be merged first.**

### DIFF
--- a/app/jobs/regular/shared_conversation_adjust_upload_security.rb
+++ b/app/jobs/regular/shared_conversation_adjust_upload_security.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module ::Jobs
+  class SharedConversationAdjustUploadSecurity < ::Jobs::Base
+    def execute(args)
+      # The conversation context includes post cooked content so this
+      # must be updated when target uploads security changes.
+      update_conversation(args[:conversation_id]) if args[:conversation_id].present?
+
+      # If we deleted the conversation then we just need to update the target's
+      # uploads security, no need to update the conversation.
+      if args[:target_id].present? && args[:target_type].present?
+        update_target(args[:target_id], args[:target_type])
+      end
+    end
+
+    def update_conversation(conversation_id)
+      conversation = SharedAiConversation.find(conversation_id)
+      return if conversation.blank?
+
+      # NOTE: Only Topics are supported for now, in future we will need a more flexible
+      # way of doing this.
+      if conversation.target_type == "Topic"
+        rebaked_posts = TopicUploadSecurityManager.new(conversation.target).run
+
+        if rebaked_posts.any?
+          new_context =
+            conversation.context.map do |context_post|
+              rebaked_post = rebaked_posts.find { |p| p.id == context_post["id"] }
+              context_post[:cooked] = rebaked_post.cooked if rebaked_post
+              context_post
+            end
+
+          conversation.update!(context: new_context)
+        end
+      end
+    end
+
+    def update_target(target_id, target_type)
+      # NOTE: Only Topics are supported for now, in future we will need a more flexible
+      # way of doing this.
+      if target_type == "Topic"
+        topic = target_type.constantize.find_by(id: target_id)
+        return if topic.blank?
+        TopicUploadSecurityManager.new(topic).run
+      end
+    end
+  end
+end
--- a/app/jobs/regular/shared_conversation_adjust_upload_security.rb
+++ b/app/jobs/regular/shared_conversation_adjust_upload_security.rb
@@ -3,19 +3,21 @@
 module ::Jobs
   class SharedConversationAdjustUploadSecurity < ::Jobs::Base
     def execute(args)
-      # The conversation context includes post cooked content so this
-      # must be updated when target uploads security changes.
-      update_conversation(args[:conversation_id]) if args[:conversation_id].present?
-
-      # If we deleted the conversation then we just need to update the target's
-      # uploads security, no need to update the conversation.
-      if args[:target_id].present? && args[:target_type].present?
+      if args[:conversation_id].present?
+        # The conversation context includes post cooked content so this
+        # must be updated when target uploads security changes.
+        update_conversation(args[:conversation_id])
+      elsif args[:target_id].present? && args[:target_type].present?
+        # If we deleted the conversation then we just need to update the target's
+        # uploads security, no need to update the conversation.
         update_target(args[:target_id], args[:target_type])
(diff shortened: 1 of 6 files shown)


plugin.rb
```diff
@@ -65,4 +65,12 @@ module ::DiscourseAi
   end
 
   reloadable_patch { |plugin| Guardian.prepend DiscourseAi::GuardianExtensions }
+
+  register_modifier(:post_with_secure_uploads?) do |_, _, topic|
+    if SharedAiConversation.exists?(target: topic)
```
SamSaffron, line 70: I think to save querying only do this check on PMs?

app/jobs/regular/shared_conversation_adjust_upload_security.rb
```diff
@@ -0,0 +1,49 @@
...
+      update_conversation(args[:conversation_id]) if args[:conversation_id].present?
+
+      # If we deleted the conversation then we just need to update the target's
+      # uploads security, no need to update the conversation.
+      if args[:target_id].present? && args[:target_type].present?
+        update_target(args[:target_id], args[:target_type])
+      end
+    end
+
+    def update_conversation(conversation_id)
```
Drenmi, line 17: Minor one, but shall we make these two methods `private` to indicate "implementation detail" and avoid accidental coupling from collaborators?

app/jobs/regular/shared_conversation_adjust_upload_security.rb
```diff
@@ -0,0 +1,49 @@
...
+      conversation = SharedAiConversation.find(conversation_id)
+      return if conversation.blank?
```
Drenmi, line 19: I believe `SharedAiConversation` is an ActiveRecord object, so `#find` would already have raised an error here. Perhaps we change to `SharedAiConversation.find_by(id: conversation_id)`?

app/jobs/regular/shared_conversation_adjust_upload_security.rb
```diff
@@ -0,0 +1,49 @@
...
+      # NOTE: Only Topics are supported for now, in future we will need a more flexible
+      # way of doing this.
+      if conversation.target_type == "Topic"
+        rebaked_posts = TopicUploadSecurityManager.new(conversation.target).run
+
+        if rebaked_posts.any?
+          new_context =
+            conversation.context.map do |context_post|
+              rebaked_post = rebaked_posts.find { |p| p.id == context_post["id"] }
+              context_post[:cooked] = rebaked_post.cooked if rebaked_post
```
Drenmi, line 30: Just want to confirm that `context_post` supports indifferent access, since we're using both `"id"` (string) and `:cooked` (symbol) here.

app/jobs/regular/shared_conversation_adjust_upload_security.rb
```diff
@@ -0,0 +1,49 @@
...
+    def execute(args)
+      # The conversation context includes post cooked content so this
+      # must be updated when target uploads security changes.
+      update_conversation(args[:conversation_id]) if args[:conversation_id].present?
+
+      # If we deleted the conversation then we just need to update the target's
+      # uploads security, no need to update the conversation.
+      if args[:target_id].present? && args[:target_type].present?
+        update_target(args[:target_id], args[:target_type])
+      end
```
Drenmi, line 14: I assume the call to `update_conversation` and `update_target` are intended to be mutually exclusive? If so, can we do an `if elsif` to make the case of both of them running impossible?

app/jobs/regular/shared_conversation_adjust_upload_security.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        if rebaked_posts.any?
+          new_context =
+            conversation.context.map do |context_post|
+              rebaked_post = rebaked_posts.find { |p| p.id == context_post["id"] }
+              context_post[:cooked] = rebaked_post.cooked if rebaked_post
+              context_post
+            end
+
+          conversation.update!(context: new_context)
```
Drenmi, line 34: If this fails, will we be triggering (default) retries due to an exception being raised? Seems like it would be unrecoverable.

plugin.rb
```diff
@@ -65,4 +65,12 @@ module ::DiscourseAi
...
+      false
+    else
+      nil
```
Drenmi, line 73: I'm on the fence on whether the explicit `nil` makes it more or less clear what's going on. I'm leaning on less. It's a bit of a "wat?" factor. 🙂 To me it's clearer that this is a modifier if the "if" stands by itself.

Drenmi (approved):
A few nits, but LGTM! 🤖

plugin.rb
martin-brennan: Yes definitely makes sense :+1:

app/jobs/regular/shared_conversation_adjust_upload_security.rb
martin-brennan: Yep makes sense :+1:

app/jobs/regular/shared_conversation_adjust_upload_security.rb
martin-brennan: Yes you're right, I think I missed this from an earlier refactor :+1:

app/jobs/regular/shared_conversation_adjust_upload_security.rb
martin-brennan: Yes that's right, and sure I can change this :+1:

plugin.rb
martin-brennan: Ah yes true, for some reason I was thinking that this was necessary :+1:

app/jobs/regular/shared_conversation_adjust_upload_security.rb
martin-brennan: True, will fix. I think I need some actual specs for this job, I missed a few things!

SamSaffron:
@martin-brennan feel free to merge, holding off just in case this impacted by any core renaming.

martin-brennan:
Thanks @SamSaffron , merging now :)

PR #555, opened 2024-04-04
DEV: replace diffhtml with morphlex

ZogStriP:
morphlex is a lighter and faster morphing library.

cf. https://github.com/discourse/discourse/commit/c5f1fc3a08eff7d9df9d2caf3688d1035dbc15d7


assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -113,13 +113,16 @@ class PostUpdater extends StreamUpdater {
...
-  async setCooked(value) {
-    this.post.set("cooked", value);
+  async setCooked(cooked) {
+    this.post.set("cooked", cooked);
 
     const oldElement = this.postElement.querySelector(".cooked");
 
-    await loadScript("/javascripts/diffhtml.min.js");
-    window.diff.innerHTML(oldElement, value);
+    // TODO: use `morphInner` once version 0.0.16 of morphlex is out
```
ZogStriP, line 121: `morphlex` doesn't support morphing between a node and a string in version 0.0.15. It has been added though - https://github.com/phlex-ruby/morphlex/commit/bf208fb20535add663aaa1c7729c5f00836ae8a3

assets/javascripts/discourse/lib/ai-streamer.js
```diff
@@ -113,13 +113,16 @@ class PostUpdater extends StreamUpdater {
...
+    const newElement = oldElement.cloneNode(false);
+    newElement.innerHTML = cooked;
+
+    (await import("morphlex")).morph(oldElement, newElement);
```
davidtaylorhq, line 125: Huh - does this actually work? I didn't think plugin JS was run through webpack... so I don't think the `import()` will be transpiled into something functional 🤔

davidtaylorhq (approved):

PR #576, opened 2024-04-12
FEATURE: Add Cohere models to AI helper and automation

SamSaffron:


nattsw (approved):
Any tests for translate_model?

lib/automation.rb
```diff
@@ -19,11 +19,14 @@ module Automation
...
         name: "discourse_automation.ai_models.mistral_7b_instruct_v0_2",
       },
+      { id: "command-r", name: "discourse_automation.ai_models.command_r" },
+      { id: "command-r-plus", name: "discourse_automation.ai_models.command_r_plus" },
     ]
 
     def self.translate_model(model)
       return "google:gemini-pro" if model == "gemini-pro"
       return "open_ai:#{model}" if model.start_with? "gpt"
+      return "cohere:#{model}" if model.start_with? "command"
```
martin-brennan, line 29: My only comment would be that, maybe we prefix these command models with `cohere_`/`cohere-` then this `model.start_with?` will be more specific; maybe some other model will start with command later on? Anyway not a big deal

martin-brennan (approved):

SamSaffron:
I think much of this code is likely going to change when @romanrizzi is finally done with his refactor, it is so odd to need to add an LLM into so many spots (bot/automation/helper) we need to normalize all of this.

lib/automation.rb
SamSaffron: yeah this is likely going to change over the next few months... this whole pattern is not great

SamSaffron:
I don't think we have tests for translate model we really need to normalize and simplify this code ... its just in too many places. @nattsw

PR #586, opened 2024-04-23
DEV: Moving around admin persona config routes again

martin-brennan:
The initial setup done in fb0d56324f9139f432c8797f25c30151ae66723b
clashed with other plugins, I found this when trying to do the same
for Gamification. This uses a better routing setup and removes the
need to define the config nav link for Settings -- that is always inserted.

Relies on https://github.com/discourse/discourse/pull/26707 . Specs will
not pass until this is merged.


assets/javascripts/discourse/admin-discourse-ai-plugin-route-map.js
```diff
@@ -4,11 +4,9 @@ export default {
   path: "/plugins",
 
   map() {
-    this.route("discourse-ai", { path: "/" }, function () {
```
martin-brennan, line 7: This top level route was the problem because it clashed with `/admin/plugins/:id` for the root show route.

jjaffeux (approved):

jjaffeux:
I approved but the specs and linting failures are legit

davidtaylorhq:
Something (your editor?) has added trailing newlines to the hbs files. The prettier standard is to omit trailing newlines on hbs (so that unwanted whitespace is not introduced between components when putting them together in a parent template)

martin-brennan:
@davidtaylorhq It's weird though because I run the exact same prettier command locally (outside my editor) that CI is running and it doesn't show any problem, with the same prettier + yarn versions. Will try again today, maybe something is out of sync.

@jjaffeux thanks but yes I expect the spec failures, they will go away after the core PR referenced in the OP is merged.

martin-brennan:
Okay so `yarn prettier blah/file.hbs` didn't show any output or warning like it usually does if the format is wrong, but `yarn prettier -w blah/file.hbs` fixed the newline issue and wrote the file :shrug: Anyway, fixed now.

PR #588, opened 2024-04-23
REFACTOR: Migrate Vllm/TGI-served models to the OpenAI format.

romanrizzi:
Both endpoints provide OpenAI-compatible servers. The only difference is that Vllm doesn't have tool support and relies on the XML format.

Thanks to this, we can remove some of our existing dialects and have specific ones for these services, allowing us to support any LLM they can serve. With the upcoming concept of data-driven LLM configurations in mind, this is very important.

 I rewrote the "translate" method to use a template method and extracted the tool support strategies into its classes to simplify the code.


xfalcox (approved):

SamSaffron:
have we tested ollama here as well on local? would be nice to land ollama with this as well.

romanrizzi:
@SamSaffron - Had to rethink this refactor a bit. It will add Ollama support now (dev only).

xfalcox (approved):

PR #594, opened 2024-04-27
FIX: api key header error

devhaozi:
According to the documentation of huggingface: `By default the server responds to every request. With an api key set, the requests must have the Authorization header set with the api key as Bearer token.`, the use of API_KEY in the plugin is incorrect, and should use Bearer Token.

Source doc url: [https://github.com/huggingface/text-embeddings-inference?tab=readme-ov-file#docker](https://github.com/huggingface/text-embeddings-inference?tab=readme-ov-file#docker)


xfalcox:
Our implementation predates TEI support for API keys by a couple months, so we relied on a reverse proxy doing API Key header verification that uses this signature, meaning we can't simply merge this.

devhaozi:
> Our implementation predates TEI support for API keys by a couple months, so we relied on a reverse proxy doing API Key header verification that uses this signature, meaning we can't simply merge this.

If so, what do you think about adding both the `X-API-KEY` and `Authorization` headers?

devhaozi:
@xfalcox fixed, PTAL.

SamSaffron:
@xfalcox what should we do here?

xfalcox:
Hey @devhaozi I have rebased it at https://github.com/discourse/discourse-ai/pull/839 and tests pass. Can you rebase it properly, or can I merge #839 ?

devhaozi:
> Hey @devhaozi I have rebased it at #839 and tests pass. Can you rebase it properly, or can I merge #839 ?

Yes, you can.

PR #595, opened 2024-04-29
Add missing gems for `rake ai:embeddings:backfill` task

fokx:
Add missing gems `ruby-progressbar` and `parallel` in plugin.rb


xfalcox:
Thanks for the PR @fokx, but I've left those out unintentionally as the rake task fell out out favor and should only be used in rare occasions by experienced operators who can easily install those out of band.

PR #599, opened 2024-05-01
FIX: call after tool calls failing on OpenAI / Gemini

SamSaffron:
A recent change meant that llm instance got cached internally, repeat calls
to inference would cache data in Endpoint object leading model to
failures.

Both Gemini and Open AI expect a clean endpoint object cause they
set data.

This amends internals to make sure llm.generate will always operate
on clean objects

### DIFF
--- a/lib/completions/llm.rb
+++ b/lib/completions/llm.rb
@@ -100,23 +100,24 @@ module DiscourseAi
             if @canned_llm && @canned_llm != model_name
               raise "Invalid call LLM call, expected #{@canned_llm} but got #{model_name}"
             end
-            return new(dialect_klass, @canned_response, model_name)
+            return new(dialect_klass, nil, model_name, gateway: @canned_response)
           end
 
-          gateway =
+          gateway_klass =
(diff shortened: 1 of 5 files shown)


lib/completions/llm.rb
```diff
@@ -174,6 +175,7 @@ def generate(
 
         model_params.keys.each { |key| model_params.delete(key) if model_params[key].nil? }
 
+        gateway = @gateway || gateway_klass.new(model_name, dialect_klass.tokenizer)
```
jjaffeux, line 178: Feels like it's easy to get in a situation where `@gateway` is nil and `gateway_klass` is nil too

jjaffeux (approved):

jjaffeux (approved):

lib/completions/llm.rb
SamSaffron: yeah I just added that for testing, I think the testing pattern is off but that is a bigger refactor 

Thanks heaps

PR #600, opened 2024-05-02
DEV: Update engine config to match skeleton

davidtaylorhq:
- adds `lib/` autoloading (and removes the plugin.rb workaround)
- adds eager loading of scheduled jobs in development (otherwise they don't work at all)


CvX (approved):

SamSaffron:
we need to merge this a bit faster ... looks like a full rebase is needed.

SamSaffron:
it is dead again :cry: needs another rebase

PR #601, opened 2024-05-06
FIX: uploader not removing app event FIX: restricted to mentionabled vs allowed_chat by mistake

SamSaffron:


lis2 (approved):

assets/javascripts/discourse/components/persona-rag-uploader.gjs
```diff
@@ -52,8 +51,13 @@ export default class PersonaRagUploader extends Component.extend(
     );
   }
```
lis2, line 53: The solution is correct. 
I am just wondering if the current one is not working because missing `@action`

SamSaffron: I think that may work as well, but kind of like this better cause it feels like less moving parts and magic

SamSaffron: TIL about this rube goldberg machine :) https://meta.discourse.org/discourse-ai/ai-bot/shared-ai-conversations/wNEqjfx0HB-OAI7kIqj6sg

lis2: Nice explanation! I think that's because of that binding feature we are using @action in other places in combination with `willDestroy`. Otherwise `this` in `removeListener` is `undefined`. So technically not even @action was required but simple @bind should work. But yeah, explicit `willDestory` block is great as well.

PR #608, opened 2024-05-08
FIX: context repairs for @mentioned bot

SamSaffron:
When the bot is @mentioned, we need to be a lot more careful
about constructing context otherwise bot gets ultra confused.

This changes multiple things:

1. We were omitting all thread first messages (fixed)
2. Include thread title (if available) in context
3. Construct context in a clearer way separating user request from data


jjaffeux (approved):

jjaffeux:
I approved but tests failures look legit

SamSaffron:
thanks will sort out tests should be easy

PR #613, opened 2024-05-13
FIX: improve function call parsing

SamSaffron:
- support " / ' wrapped values
- coerce integer to integer
- enforce enum at boundary

### DIFF
--- a/lib/ai_bot/personas/persona.rb
+++ b/lib/ai_bot/personas/persona.rb
@@ -208,6 +208,15 @@ module DiscourseAi
                 rescue JSON::ParserError
                   [value.to_s]
                 end
+            elsif param[:type] == "string" && value
+              value = strip_quotes(value).to_s
+            elsif param[:type] == "integer" && value
+              value = strip_quotes(value).to_i
+            end
+
+            if param[:enum] && value && !param[:enum].include?(value)
+              # invalid enum value
+              value = nil
             end
 
             arguments[name.to_sym] = value if value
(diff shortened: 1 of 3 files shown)


lib/ai_bot/personas/persona.rb
```diff
@@ -223,6 +232,20 @@ def tool_instance(parsed_function, bot_user:, llm:, context:)
           )
         end
 
+        def strip_quotes(value)
+          if value.is_a?(String)
+            if value.start_with?('"') && value.end_with?('"')
+              value = value[1..-2]
+            elsif value.start_with?("'") && value.end_with?("'")
+              value = value[1..-2]
+            else
```
jjaffeux, line 241: Not sure about your use case here, but should we attempt to target more quotes?

```ruby
if value.match?(/\A[‘’“”‹›«»"'].*[‘’“”‹›«»"']\z/)
    value = value[1..-2]
  end
```

jjaffeux (approved):

lib/ai_bot/personas/persona.rb
jjaffeux: chat gpt suggest this regex:

```ruby
def strip_opposite_quotes(value)
  # Use regex to strip matching opposite quotes at the start and end of a string
  value.gsub(/\A(?:["'](.*?)["']|“(.*?)”|‘(.*?)’|«(.*?)»|「(.*?)」)\z/, '\1\2\3\4\5')
end
```

Might be too convulted for our case, but I have been bitten multiple times by these non "classic" quotes, and prefer to mention them.

SamSaffron: I hear you, but this is pretty safe as is cause this is all llm generated function calls so the odds are pretty low it will mix in something incorrect here. possible but unlikely.

PR #618, opened 2024-05-14
FEATURE: GPT4o support and better auditing

SamSaffron:
- Introduce new support for GPT4o (automation / bot / summary / helper)
- Properly account for token counts on OpenAI models
- Track feature that was used when generating AI completions

### DIFF
--- a/app/models/ai_api_audit_log.rb
+++ b/app/models/ai_api_audit_log.rb
@@ -22,12 +22,13 @@ end
 #  id                   :bigint           not null, primary key
 #  provider_id          :integer          not null
 #  user_id              :integer
-#  topic_id             :integer
-#  post_id              :integer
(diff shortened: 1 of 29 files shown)


nattsw:
@SamSaffron looks like legit failures - 

```
1) DiscourseAi::AiBot::Tools::SearchSettings#process can return more many settings with no descriptions if there are lots of hits
     Failure/Error: { name: strategy.display_name, value: strategy.model }
     
     NoMethodError:
       undefined method `display_name' for nil:NilClass
     # ./plugins/discourse-ai/lib/summarization/models/custom_llm.rb:8:in `display_name'
```

SamSaffron:
Yep, will fix we have a flake

lis2 (approved):

PR #622, opened 2024-05-14
FIX: respect "external_links_in_new_tab" user preference...

ZogStriP:
... for AI conversations since they look & feel "external".

Internal ref - /t/128182

### DIFF
--- a/assets/javascripts/discourse/routes/discourse-ai-shared-conversation-show.js
+++ b/assets/javascripts/discourse/routes/discourse-ai-shared-conversation-show.js
@@ -1,8 +1,19 @@
+import { service } from "@ember/service";
 import DiscourseRoute from "discourse/routes/discourse";
 
-export default DiscourseRoute.extend({
+export default class DiscourseAiSharedConversationShowRoute extends DiscourseRoute {
+  @service currentUser;
+
   beforeModel(transition) {
-    window.location = transition.intent.url;
+    if (this.currentUser?.user_option?.external_links_in_new_tab) {
+      window.open(transition.intent.url, "_blank");
+    } else {
+      this.redirect(transition.intent.url);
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/routes/discourse-ai-shared-conversation-show.js
```diff
@@ -1,8 +1,19 @@
...
+    } else {
+      this.redirect(transition.intent.url);
+    }
     transition.abort();
-  },
-});
+  }
+
+  redirect(url) {
+    window.location = url;
```
martin-brennan, line 17: Minor nit, but why bother with this function when it's only used on one line and it's also a one liner?

martin-brennan (approved):
LGTM :+1:

assets/javascripts/discourse/routes/discourse-ai-shared-conversation-show.js
ZogStriP: Otherwise we can't test it. It's really hard to mock/stub `window.location` 🤷‍♂️

test/javascripts/unit/routes/discourse-ai-shared-conversation-show-route-test.js
```diff
@@ -0,0 +1,64 @@
...
+          url,
+          "https://www.discourse.org",
+          "redirect was called with the correct URL"
+        );
+      };
+
+      route.beforeModel(transition);
+
+      // Reset
+      window.open = originalOpen;
```
davidtaylorhq, line 60: If the test fails, this will not run. I think it would be better in a `} finally {` block.

Or even better, use `sinon` to handle the stubbing - it'll reset itself afterwards.

PR #626, opened 2024-05-16
FEATURE: safeguard to avoid over triage

SamSaffron:
- a post can be triage a maximum of twice a minute
- system can run a total of 40 triages a minute

Low defaults were picked to safeguard against any possible loops


nattsw (approved):
Looks ok but some thoughts about the "hidden" site settings on high traffic sites.

PR #630, opened 2024-05-17
FEATURE: JavaScript evaluation tool

SamSaffron:
This is similar to code interpreter by ChatGPT, except that it uses
JavaScript as the execution engine.

Safeguards were added to ensure memory is constrained and evaluation
times out.

### DIFF
--- a/lib/ai_bot/tools/javascript_evaluator.rb
+++ b/lib/ai_bot/tools/javascript_evaluator.rb
@@ -58,7 +58,17 @@ module DiscourseAi
               marshal_stack_depth: MARSHAL_STACK_DEPTH,
             )
 
-          result = context.eval(script)
+          # works around llms like anthropic loving console.log
+          eval_script = <<~JS
+            let console = {};
+            console.log = function(val) {
+              return val;
+            };
+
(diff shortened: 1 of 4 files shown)


SamSaffron:
For example: 

image

lib/ai_bot/tools/javascript_evaluator.rb
```diff
@@ -0,0 +1,107 @@
...
+
+            #{script}
+          JS
+
+          result = context.eval(eval_script)
+
+          # only do special handling and truncating for long strings
+          if result.to_s.length > 1000
+            result = result.to_s
+            result = truncate(result.to_s, max_length: 10_000, percent_length: 0.3, llm: llm)
```
Drenmi, line 76: Seems like we're doubling up `result.to_s` here.

Drenmi (approved):
LGTM! 🤖

lib/ai_bot/tools/javascript_evaluator.rb
SamSaffron: true , will fix

PR #638, opened 2024-05-23
DRAFT: Create AI Bot users dynamically and support custom LlmModels

romanrizzi:


config/locales/client.en.yml
```diff
@@ -226,6 +227,7 @@ en:
         hints:
           max_prompt_tokens: "Max numbers of tokens for the prompt. As a rule of thumb, this should be 50% of the model's context window."
           name: "We include this in the API call to specify which model we'll use."
+          companion_bot_username: "Features like the AI Bot create an user with this username for users to interact with."
```
pmusaraj, line 230:
```suggestion
          companion_bot_username: "Some features, like the AI Bot, set up a companion user account other users can interact with. Set this companion user's username here."
```

config/settings.yml
```diff
@@ -342,30 +342,19 @@ discourse_ai:
...
     type: group_list
     list_type: compact
     default: "1|2" # 1: admins, 2: moderators
     allow_any: false
     refresh: true
-  ai_bot_enabled_chat_bots:
+  ai_bot_enabled_chat_bots: # TODO(roman): Make this dynamic
     type: list
-    default: "gpt-3.5-turbo"
+    default: ""
```
pmusaraj, line 354: Unsetting the default means some instances will have this feature disabled after a deploy. I guess you're working on a migration for that too, right? (Not sure I'm reading the internal topic correctly.)

romanrizzi: We track enabled models in the `llm_models` table. I made this hidden and update the migration to copy the data.

config/settings.yml
```diff
@@ -342,30 +342,18 @@ discourse_ai:
...
-      - gpt-4-turbo
-      - gpt-4o
-      - claude-2
-      - gemini-1.5-pro
-      - mixtral-8x7B-Instruct-V0.1
-      - claude-3-opus
-      - claude-3-sonnet
-      - claude-3-haiku
-      - cohere-command-r-plus
+    "gpt-3.5-turbo"
```
pmusaraj, line 354: Is this line missing a `default: `?

pmusaraj (approved):
Looks good to me!

app/models/llm_model.rb
```diff
@@ -20,4 +65,7 @@ def tokenizer_class
 #  updated_at        :datetime         not null
 #  url               :string
 #  api_key           :string
+#  bot_username      :string
```
SamSaffron, line 68: we should not store usernames here, we should store bot_user_id, we need to be robust to renames. given we already have user_id we should remove then?

romanrizzi: Sure, let's do it. I just wanted to make sure bot usernames stays consistent when deleting/creating by toggling the enabled attr. Maybe it was too much.

xfalcox (approved):

PR #658, opened 2024-06-06
DEV: Add summarization logic from core

keegangeorge:
This PR adds summarization logic back to the plugin from core. See core PR: https://github.com/discourse/discourse/pull/27373

### DIFF
--- a/assets/javascripts/discourse/connectors/topic-map-expanded-after/ai-summary-box.gjs
+++ b/assets/javascripts/discourse/connectors/topic-map-expanded-after/ai-summary-box.gjs
@@ -109,7 +109,10 @@ export default class AiSummaryBox extends Component {
 
   @bind
   unsubscribe() {
-    this.messageBus.unsubscribe("/summaries/topic/*", this._updateSummary);
+    this.messageBus.unsubscribe(
+      "/discourse-ai/summaries/topic/*",
+      this._updateSummary
+    );
   }
 
   @bind
@@ -144,7 +147,7 @@ export default class AiSummaryBox extends Component {
               @title="summary.buttons.hide"
               @label="summary.buttons.hide"
               @icon="chevron-up"
-              class="btn-primary topic-strategy-summarization"
+              class="btn-primary ai-topic-summarization"
             />
           {{else}}
             <DButton
@@ -153,7 +156,7 @@ export default class AiSummaryBox extends Component {
               @translatedTitle={{this.generateSummaryTitle}}
               @icon={{this.generateSummaryIcon}}
               @disabled={{this.loading}}
-              class="btn-primary topic-strategy-summarization"
+              class="btn-primary ai-topic-summarization"
             />
           {{/if}}
         {{/if}}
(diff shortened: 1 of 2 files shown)


db/migrate/20240606152117_copy_summary_sections_to_ai_summaries.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CopySummarySectionsToAiSummaries < ActiveRecord::Migration[7.0]
```
romanrizzi, line 3: This should be under `post_migrate`, and it should be irreversible.

db/migrate/20240606151348_create_ai_summaries_table.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateAiSummariesTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :ai_summaries do |t|
+      t.integer :target_id, null: false
+      t.string :target_type, null: false
+      t.int4range :content_range
+      t.string :summarized_text, null: false
+      t.integer :meta_section_id
```
romanrizzi, line 10: We should drop this, and only migrate those where `meta_section_id` is null. This means they are the final summary shown to users. We never really use these small fragments so it's a good opportunity to get rid of them

config/settings.yml
```diff
@@ -317,6 +317,15 @@ discourse_ai:
   ai_summarization_discourse_service_api_key:
     default: ""
     secret: true
+  ai_summarization_strategy:
```
romanrizzi, line 320: I think we are missing a way to migrate the old setting here? We don't want to break summaries

app/services/topic_summarization.rb
```diff
@@ -0,0 +1,129 @@
...
+    main_summary =
+      AiSummary.create!(
+        target: topic,
+        algorithm: strategy.display_name,
+        content_range: (post_numbers.first..post_numbers.last),
+        summarized_text: result[:summary],
+        original_content_sha: build_sha(post_numbers),
+      )
+
+    result[:chunks].each do |chunk|
```
romanrizzi, line 112: Dropping `meta_section_id` means we can simplify this

romanrizzi (changes requested):

app/services/topic_summarization.rb
romanrizzi: Yes, exactly

db/post_migrate/20240610232546_copy_custom_summarization_allowed_groups_to_ai_custom_summarization_allowed_groups.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class CopyCustomSummarizationAllowedGroupsToAiCustomSummarizationAllowedGroups < ActiveRecord::Migration[
+  7.0
```
xfalcox, line 4: lol

lib/summarization/entry_point.rb
```diff
@@ -67,7 +81,10 @@ def inject_into(plugin)
         #end
 
         foldable_models.each do |model|
-          plugin.register_summarization_strategy(Strategies::FoldContent.new(model))
+          DiscoursePluginRegistry.register_summarization_strategy(
```
romanrizzi, line 84: I think we can just return `foldable_models` in `Models::Base#available_strategies`? Nobody else consumes this register

romanrizzi (changes requested):

romanrizzi (approved):

PR #662, opened 2024-06-10
FEATURE: optional tool detail blocks

SamSaffron:
This is a rather huge refactor with 1 new feature (tool details can
be suppressed)

Previously we use the name "Command" to describe "Tools", this unifies
all the internal language and simplifies the code.

We also amended the persona UI to use less DToggles which aligns
with our design guidelines.


db/migrate/20240609061418_tool_details_and_command_removal.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    add_column :ai_personas, :tool_details, :boolean, default: true, null: false
+    add_column :ai_personas, :tools, :json, null: false, default: []
+    Migration::ColumnDropper.mark_readonly(:ai_personas, :commands)
+
+    execute <<~SQL
+      UPDATE ai_personas
+      SET tools = commands
+    SQL
+  end
```
romanrizzi, line 13: Should we split this into `up` and `down` and make it irreversible?

SamSaffron: I feel this is a bit of a fantasy on dev side, once we have gone up... the only way to go down is to go up again. it really is only useful in dev

Will sort out the tests though

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -433,20 +379,56 @@ export default class PersonaEditor extends Component {
...
+          />
+        </div>
+      {{/if}}
+      <div class="control-group ai-persona-editor__tool-details">
+        <label>
+          <Input @type="checkbox" @checked={{this.editingModel.tool_details}} />
+          {{I18n.t "discourse_ai.ai_persona.tool_details"}}</label>
+        <DTooltip
+          @icon="question-circle"
+          @content={{I18n.t "discourse_ai.ai_persona.tool_details"}}
```
martin-brennan, line 416:
```suggestion
          @content={{I18n.t "discourse_ai.ai_persona.tool_details_help"}}
```

Minor oversight, the tip is the same as the label atm but you have the help tip defined in en.yml

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -433,20 +379,56 @@ export default class PersonaEditor extends Component {
...
         />
       </div>
-      <div class="control-group">
-        <label>{{I18n.t "discourse_ai.ai_persona.max_context_posts"}}</label>
-        <Input
-          @type="number"
-          lang="en"
-          class="ai-persona-editor__max_context_posts"
-          @value={{this.editingModel.max_context_posts}}
+      {{#if this.editingModel.user}}
```
martin-brennan, line 382: All the checkbox changes here look good, definitely the way to go rather than toggles for this stuff :+1:

assets/javascripts/discourse/components/ai-persona-editor.gjs
```diff
@@ -456,6 +438,19 @@ export default class PersonaEditor extends Component {
...
+          lang="en"
+          class="ai-persona-editor__max_context_posts"
+          @value={{this.editingModel.max_context_posts}}
+        />
+        <DTooltip
+          @icon="question-circle"
+          @content={{I18n.t "discourse_ai.ai_persona.max_context_posts_help"}}
+        />
+      </div>
       <div class="control-group">
```
martin-brennan, line 454: I noticed in the UI Temperature and Top P are kind of grouped together; it's happening because they are in the same `control-group` div, just need to split it into 2:

image

assets/javascripts/discourse/components/ai-persona-tool-options.gjs
```diff
@@ -0,0 +1,81 @@
...
+
+export default class AiPersonaToolOptions extends Component {
+  get showToolOptions() {
+    const allTools = this.args.allTools;
+    if (!allTools) {
+      return false;
+    }
+
+    return this.toolNames.any(
+      (tool) => allTools.find((c) => c.id === tool)?.options
```
martin-brennan, line 13:
```suggestion
      (tool) => allTools.findBy("id", tool)?.options
```

assets/javascripts/discourse/components/ai-persona-tool-options.gjs
```diff
@@ -0,0 +1,81 @@
+import Component from "@glimmer/component";
+import I18n from "discourse-i18n";
+import AiPersonaToolOptionEditor from "./ai-persona-tool-option-editor";
+
+export default class AiPersonaToolOptions extends Component {
+  get showToolOptions() {
+    const allTools = this.args.allTools;
```
martin-brennan, line 7: I don't think there is any need to store this in a separate const, you don't do it in `toolNames`

assets/javascripts/discourse/components/ai-persona-tool-options.gjs
```diff
@@ -0,0 +1,81 @@
...
+      }
+    });
+  }
+
+  get toolOptions() {
+    if (!this.args.tools) {
+      return [];
+    }
+
+    const allTools = this.args.allTools;
```
martin-brennan, line 35: Also don't see a need to store this in const here

assets/javascripts/discourse/components/ai-persona-tool-options.gjs
```diff
@@ -0,0 +1,81 @@
...
+    if (!allTools) {
+      return [];
+    }
+
+    const options = [];
+    this.toolNames.forEach((toolId) => {
+      const tool = allTools.find((c) => c.id === toolId);
```
martin-brennan, line 42:
```suggestion
      const tool = allTools.findBy("id", toolId);
```

config/locales/client.en.yml
```diff
@@ -131,12 +131,14 @@ en:
         max_context_posts: "Max Context Posts"
         max_context_posts_help: "The maximum number of posts to use as context for the AI when responding to a user. (empty for default)"
         vision_enabled: Vision Enabled
-        vision_enabled_help: If enabled, the AI will attempt to understand images users post in the topic, depends on the model being used supporting vision. Anthropic Claude 3 models support vision.
+        vision_enabled_help: If enabled, the AI will attempt to understand images users post in the topic, depends on the model being used supporting vision. Supported by latest models from Anthropic / Google and OpenAI.
```
martin-brennan, line 134:
```suggestion
        vision_enabled_help: If enabled, the AI will attempt to understand images users post in the topic, depends on the model being used supporting vision. Supported by latest models from Anthropic, Google, and OpenAI.
```

minor nit

db/post_migrate/20240609232736_drop_commands_from_ai_personas.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+class DropCommandsFromAiPersonas < ActiveRecord::Migration[7.0]
```
martin-brennan, line 2: I think you need to add the `DROPPED_COLUMNS` const here too https://github.com/discourse/discourse/blob/10b9a32abb9967b3c0876a43ff8814c7bae8f701/db/post_migrate/20240202052058_drop_badge_image_column.rb ?

config/locales/client.en.yml
```diff
@@ -131,12 +131,14 @@ en:
...
         vision_max_pixels: Supported image size
         vision_max_pixel_sizes:
           low: Low Quality - cheapest (256x256)
           medium: Medium Quality (512x512)
           high: High Quality - slowest (1024x1024)
+        tool_details: Show Tool Details
+        tool_details_help: Will show end users details on tools the Language model has triggered.
```
martin-brennan, line 141:
```suggestion
        tool_details_help: Will show end users details on which tools the language model has triggered.
```

minor nit

martin-brennan (approved):
Mostly minor comments here :+1:

assets/javascripts/discourse/components/ai-persona-editor.gjs
SamSaffron: cool will sort this.

db/post_migrate/20240609232736_drop_commands_from_ai_personas.rb
SamSaffron: not sure why, that seems odd to force that... column dropper may though be a bit of a shortcut I could use.

assets/javascripts/discourse/components/ai-persona-tool-options.gjs
SamSaffron: its so minor but it avoids needing to do this.args.allTools twice, certainly not a hill I would die on.

db/post_migrate/20240609232736_drop_commands_from_ai_personas.rb
martin-brennan: The const is apparently for the backup restore, it's in our lore topic. Not sure tho :shrug:

PR #666, opened 2024-06-12
FEATURE: Add breadcrumbs to LLMs and Persona admin pages

martin-brennan:
Followup to https://github.com/discourse/discourse-ai/pull/656,
adding these back in with the new core component.


tgxworld (approved):
Code looks good to me. I do wonder if we should write system tests to ensure the way this plugin uses a core component is correct.

martin-brennan:
Noting that I am writing system specs for this, won't merge until I do. Ran into some breadcrumbs issues in core that cvx is fixing.

SamSaffron:
oh @martin-brennan this needs a rebase (to allow for new screen)

martin-brennan:
@SamSaffron yep all good, I am having some struggles with the breadcrumb component ATM, and I am writing system specs for this too, will put this back in Draft for now.

PR #670, opened 2024-06-18
UX: show image caption button on image hover

awesomerobot:
This hides the image caption button on non-touch devices (desktop) until the image is hovered. 

In https://github.com/discourse/discourse/commit/0a99407bfb363a28ae1a419cc5e4c16346b784f8 we made the image controls always visible (not just on hover), and the caption button can kind of get in the way. So this helps hide it until needed since it's a little large. 

Before (button obscures image always): 

image

After (button appears on control or image hover):
image


pmusaraj (approved):
This button feels too big to me (even on hover), but we can cross that bridge later.

PR #681, opened 2024-06-21
FEATURE: LLM presets for model creation

SamSaffron:
Previous to this users needed to look up complicated settings
when setting up models.

This introduces and extensible preset system with Google/OpenAI/Anthropic
presets.

This will cover all the most common LLMs, we can always add more as
we go.

Additionally:

- Proper support for Anthropic Claude Sonnet 3.5
- Stop blurring api keys when navigating away - this made it very complex to reuse keys

### DIFF
--- a/assets/javascripts/discourse/components/ai-llm-editor.gjs
+++ b/assets/javascripts/discourse/components/ai-llm-editor.gjs
@@ -46,7 +46,7 @@ export default class AiLlmEditor extends Component {
       return;
     }
 
-    let info = this.args.llms.resultSetMeta.presets.find(
+    const info = this.args.llms.resultSetMeta.presets.findBy("id", id);
       (_info) => _info.id === id
     );
     const modelInfo = info.models.findBy("name", model);
--- a/assets/javascripts/discourse/components/ai-llm-editor.gjs
+++ b/assets/javascripts/discourse/components/ai-llm-editor.gjs
@@ -8,11 +8,13 @@ import ComboBox from "select-kit/components/combo-box";
 import AiLlmEditorForm from "./ai-llm-editor-form";
 
 export default class AiLlmEditor extends Component {
-  @tracked ranWizard = false;
-  preConfiguredLlm = "none";
+  @tracked presetConfigured = false;
+  presetId = "none";
 
-  get showWizard() {
-    return this.args.model.isNew && !this.ranWizard && !this.args.model.url;
+  get showPresets() {
+    return (
+      this.args.model.isNew && !this.presetConfigured && !this.args.model.url
+    );
   }
 
   get preConfiguredLlms() {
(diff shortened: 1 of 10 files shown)


assets/javascripts/discourse/components/ai-llm-editor.gjs
```diff
@@ -1,306 +1,89 @@
...
-    };
-
-    return this.args.llms.resultSetMeta.providers.map((prov) => {
-      return { id: prov, name: t(prov) };
-    });
-  }
-
-  get adminUser() {
-    return AdminUser.create(this.args.model?.user);
+  get showWizard() {
```
martin-brennan, line 14: Would love to leave out `wizard` terminology here, don't want to set any precedents yet :laughing: 

Suggest `showPresets` here and `chosePreset` instead of `ranWizard`

assets/javascripts/discourse/components/ai-llm-editor.gjs
```diff
@@ -1,306 +1,89 @@
...
-    return this.testRunning || this.testResult !== null;
+    return options;
   }
 
   @action
-  makeApiKeySecret() {
-    this.apiKeySecret = true;
-  }
+  next() {
+    this.ranWizard = true;
```
martin-brennan, line 42: See other comment about `ranWizard`

assets/javascripts/discourse/components/ai-llm-editor.gjs
```diff
@@ -1,306 +1,89 @@
...
-            this.router.transitionTo(
-              "adminPlugins.show.discourse-ai-llms.index"
-            );
-          })
-          .catch(popupAjaxError);
-      },
+    let info = this.args.llms.resultSetMeta.presets.find(
+      (_info) => _info.id === id
+    );
+    let modelInfo = info.models.find((m) => m.name === model);
```
martin-brennan, line 52:
```suggestion
    const modelInfo = info.models.findBy("name", model);
```

assets/javascripts/discourse/components/ai-llm-editor.gjs
```diff
@@ -1,306 +1,89 @@
...
-          .destroyRecord()
-          .then(() => {
-            this.args.llms.removeObject(this.args.model);
-            this.router.transitionTo(
-              "adminPlugins.show.discourse-ai-llms.index"
-            );
-          })
-          .catch(popupAjaxError);
-      },
+    let info = this.args.llms.resultSetMeta.presets.find(
```
martin-brennan, line 49:
```suggestion
    const info = this.args.llms.resultSetMeta.presets.findBy("id", id);
```

martin-brennan: Also `info` is a bit vague, maybe `selectedPreset`?

assets/javascripts/discourse/components/ai-llm-editor.gjs
martin-brennan: Similar info comment, maybe `selectedPresetModel`

lib/completions/llm.rb
```diff
@@ -18,6 +18,63 @@ class Llm
       UNKNOWN_MODEL = Class.new(StandardError)
 
       class << self
+        def presets
+          # Sam: I am not sure if it makes sense to translate model names at all
```
martin-brennan, line 22: I think it's fine FWIW, these are essentially brand names

martin-brennan (approved):
Works and looks nice locally, mostly stylistic comments here. Really needs a system spec too, you will follow up with that next week yes?

assets/javascripts/discourse/components/ai-llm-editor.gjs
SamSaffron: will change

SamSaffron:
thanks @martin-brennan still have a few more mins will see if I can add a system spec, worst case will add it next week.

PR #686, opened 2024-06-24
FEATURE: Store provider-specific parameters.

romanrizzi:
Previously, we stored request parameters like the OpenAI organization and Bedrock's access key and region as site settings. This change stores them in the `llm_models` table instead, letting us drop more settings while also becoming more flexible.


SamSaffron (approved):

SamSaffron:
generally fine with this, but we need to get a bit better with JS testing, the system test we have could select bedrock and see that the right fields are there.

PR #692, opened 2024-06-27
FIX: never provide tools with invalid UTF-8 strings

SamSaffron:
Previous to this change, on truncation we could return invalid
UTF-8 strings to caller

This also allows tools to read up to 20 megs vs the old 4 megs.


SamSaffron:
cool will address Kelv in a comment, the 4mb was arbitrary to start with and it chokes with certain tools (github search files on Discourse)

PR #694, opened 2024-06-27
DEV: Clean up after #677

CvX:
Follow up to b863ddc94bf03e1868845e10ba744bef1f68841d

Ruby:
* Validate `summary` (the column is `not null`)
* Fix `name` validation (the column has `max_length` 100)
* Fix table annotations
* Accept missing `parameter` attributes (`required, `enum`, `enum_values`)

JS:
* Use native classes
* Don't use ember's array extensions
* Add explicit service injections
* Correct class names
* Use `||=` operator
* Use `store` service to create records
* Remove unused service injections
* Extract consts
* Group actions together
* Use `async`/`await`
* Use `withEventValue`
* Sort html attributes
* Use DButtons `@label` arg
* Use `input` elements instead of Ember's `Input` component (same w/ textarea)
* Remove `btn-default` class (automatically applied by DButton)
* Don't mix `I18n.t` and `i18n` in the same template
* Don't track props that aren't used in a template
* Correct invalid `target.value` code
* Remove unused/invalid `this.parameter`/`onChange` code
* Whitespace
* Use the new service import `inject as service` -> `service`
* Use `Object.entries()`
* Add missing i18n strings
* Fix an error in `addEnumValue` (calling `pushObject` on `undefined`)
* Use `TrackedArray`/`TrackedObject`
* Transform tool `parameters` keys (`enumValues` -> `enum_values`)

### DIFF
--- a/admin/assets/javascripts/discourse/routes/admin-plugins-show-discourse-ai-tools-show.js
+++ b/admin/assets/javascripts/discourse/routes/admin-plugins-show-discourse-ai-tools-show.js
@@ -1,17 +1,18 @@
 import DiscourseRoute from "discourse/routes/discourse";
 
-export default DiscourseRoute.extend({
+export default class DiscourseAiToolsShowRoute extends DiscourseRoute {
   async model(params) {
     const allTools = this.modelFor("adminPlugins.show.discourse-ai-tools");
     const id = parseInt(params.id, 10);
-    return allTools.findBy("id", id);
-  },
 
-  setupController(controller, model) {
-    this._super(controller, model);
+    return allTools.find((tool) => tool.id === id);
+  }
+
+  setupController(controller) {
+    super.setupController(...arguments);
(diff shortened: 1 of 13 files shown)


admin/assets/javascripts/discourse/routes/admin-plugins-show-discourse-ai-tools-show.js
```diff
@@ -1,17 +1,18 @@
...
+export default class DiscourseAiToolsShowRoute extends DiscourseRoute {
   async model(params) {
     const allTools = this.modelFor("adminPlugins.show.discourse-ai-tools");
     const id = parseInt(params.id, 10);
-    return allTools.findBy("id", id);
-  },
 
-  setupController(controller, model) {
-    this._super(controller, model);
+    return allTools.find((tool) => tool.id === id);
```
SamSaffron, line 8: @martin-brennan and you need to have a chat :) he keeps pushing me to use findBy ... I don't really care, I do get that find is a fraction of a micro second faster.

PR #698, opened 2024-07-02
DEV: Use site locale for composer helper translations

keegangeorge:
As requested on [Meta](https://meta.discourse.org/t/translation-option-of-ai-helper-in-combination-with-non-english-interface-language/313689?u=keegan): 

Currently, all translation prompts translate to the interface locale if it has been set by the user. However with this PR, it updates the AI helper work as:
- **Post** Helper: Translates to **Interface** Locale
- **Composer** Helper: Translates to **Site** locale

Instead of adding a separate prompt for handling this, we do this instead by adding a param to the `suggest` assistant controller action called `force_default_locale`. By default the param is false, however, in the composer AI helper, when we make the request, we pass it as true so that it forces the translation to be from the site locale instead of the interface locale.

### DIFF
--- a/assets/javascripts/discourse/connectors/after-d-editor/ai-helper-context-menu.js
+++ b/assets/javascripts/discourse/connectors/after-d-editor/ai-helper-context-menu.js
@@ -82,18 +82,17 @@ export default class AiHelperContextMenu extends Component {
         // AI helper by default returns interface locale on translations
         // Since we want site default translations (and we are using: force_default_locale)
         // we need to replace the translated_name with the site default locale name
-        const siteLocale = this.siteSettings?.default_locale;
+        const siteLocale = this.siteSettings.default_locale;
         const availableLocales = JSON.parse(
           this.siteSettings.available_locales
         );
         const locale = availableLocales.find((l) => l.value === siteLocale);
-        const translatePromptName = I18n.t(
+        const translatedName = I18n.t(
           "discourse_ai.ai_helper.context_menu.translate_prompt",
           {
-            language: locale?.name,
+            language: locale.name,
           }
         );
-        const translatedName = locale ? translatePromptName : "Translate";
 
         if (p.name === "translate") {
           return { ...p, translated_name: translatedName };


config/locales/server.en.yml
```diff
@@ -160,7 +160,7 @@ en:
       errors:
         completion_request_failed: "Something went wrong while trying to provide suggestions. Please, try again."
       prompts:
-        translate: Translate to English
+        translate: Translate to %{language}
```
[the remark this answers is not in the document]
gschlager: Translations will be cleared automatically.

assets/javascripts/discourse/connectors/after-d-editor/ai-helper-context-menu.js
```diff
@@ -76,7 +77,29 @@ export default class AiHelperContextMenu extends Component {
...
+          "discourse_ai.ai_helper.context_menu.translate_prompt",
+          {
+            language: locale?.name,
+          }
+        );
+        const translatedName = locale ? translatePromptName : "Translate";
+
+        if (p.name === "translate") {
+          return { ...p, translated_name: translatedName };
+        }
```
gschlager, line 100: This code block is a bit weird. Is it even possible that `this.siteSettings` can be `null` when this is called? If that's the case, then wouldn't `JSON.parse(this.siteSettings.available_locales);` fail as well? Also, is it really possible that the `siteLocale` can't be found in `availableLocales`?

config/locales/server.en.yml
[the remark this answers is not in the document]
gschlager: BTW, I have a quick question: Are those prompts just sent to AI, or are they shown to the user as well? If they are only for AI, is it really necessary to translate them?

lib/ai_helper/assistant.rb
```diff
@@ -27,9 +27,19 @@ def available_prompts
...
               prompts.map do |prompt|
-                translation =
-                  I18n.t("discourse_ai.ai_helper.prompts.#{prompt.name}", default: nil) ||
-                    prompt.translated_name || prompt.name
+                if prompt.name == "translate"
+                  locale = user.effective_locale
+                  locale_hash = LocaleSiteSetting.language_names[locale]
+                  translation =
+                    I18n.t(
+                      "discourse_ai.ai_helper.prompts.#{prompt.name}",
```
gschlager, line 35: You could even use the actual translation key in this case.

```suggestion
                      "discourse_ai.ai_helper.prompts.translate",
```

lib/ai_helper/assistant.rb
```diff
@@ -44,9 +54,12 @@ def available_prompts
...
       end
 
-      def custom_locale_instructions(user = nil)
+      def custom_locale_instructions(user = nil, force_default_locale)
         locale = SiteSetting.default_locale
-        locale = user.locale || SiteSetting.default_locale if SiteSetting.allow_user_locale && user
+        if !force_default_locale
+          locale = user.locale || SiteSetting.default_locale if SiteSetting.allow_user_locale &&
+            user
+        end
```
gschlager, line 62: Maybe simpler?

```suggestion
        locale = user.effective_locale if !force_default_locale
```

lib/ai_helper/assistant.rb
```diff
@@ -57,15 +70,19 @@ def custom_locale_instructions(user = nil)
...
           prompt.messages[0][:content] = prompt.messages[0][:content] + locale_instructions
         end
 
         if prompt.messages[0][:content].include?("%LANGUAGE%")
           locale = SiteSetting.default_locale
-          locale = user.locale if SiteSetting.allow_user_locale && user&.locale.present?
+
+          if !force_default_locale
+            locale = user.locale if SiteSetting.allow_user_locale && user&.locale.present?
+          end
```
gschlager, line 84:
```suggestion
          locale = user.effective_locale if user && !force_default_locale
```

gschlager (changes requested):

gschlager (approved):

PR #699, opened 2024-07-03
FEATURE: move summary to use llm_model

SamSaffron:
When we moved summary back from core it relied on legacy site settings, this completes the migration
and now llm_model is used for all AI features.


romanrizzi (changes requested):
Changes looks good, but we need to migrate the renamed setting

romanrizzi (approved):

romanrizzi (approved):

PR #725, opened 2024-07-24
FEATURE: Track if a model can do vision in the llm_models table

romanrizzi:


lib/completions/dialects/chat_gpt.rb
```diff
@@ -45,6 +45,10 @@ def max_prompt_tokens
           model_max_tokens - buffer
         end
 
+        def vision_support?
+          super || %w[gpt-4-vision gpt-4-turbo gpt-4o].any? { |vm| model_name.include?(vm) }
```
xfalcox, line 49: Can we remove this and rely on the vision column only? Maybe a migration that ticks this to true on models that match this name?

PR #727, opened 2024-07-26
FIX: Properly fix ai_summaries table sequence

xfalcox:
Previous attempt at 3815360 could fail due to a race introduced in 1b0ba91 where summaries are migrated to core in a post_migrate erroneously.

### DIFF
--- a/db/migrate/20240726164937_fix_ai_summaries_sequence.rb
+++ b/db/migrate/20240726164937_fix_ai_summaries_sequence.rb
@@ -2,30 +2,34 @@
 
 class FixAiSummariesSequence < ActiveRecord::Migration[7.0]
   def up
-    execute <<-SQL
-      SELECT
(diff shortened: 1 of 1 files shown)


db/migrate/20240726164937_fix_ai_summaries_sequence.rb
```diff
@@ -0,0 +1,34 @@
...
+                  SELECT
+                    MAX(id)
+                  FROM
+                    summary_sections
+                ),
+                (
+                  SELECT
+                    max(id)
+                  FROM
+                    summary_sections
```
romanrizzi, line 22: This is supposed to be the new table?

xfalcox: yes

pmusaraj (approved):

PR #729, opened 2024-07-29
DEV: Remove old code now that features rely on LlmModels.

romanrizzi:


lib/completions/endpoints/vllm.rb
```diff
@@ -44,16 +30,13 @@ def provider_id
...
+          if llm_model.url == LlmModel::RESERVED_VLLM_SRV_URL
+            service = DiscourseAi::Utils::DnsSrv.lookup(SiteSetting.ai_vllm_endpoint_srv)
             api_endpoint = "https://#{service.target}:#{service.port}/v1/chat/completions"
           else
-            api_endpoint = "#{SiteSetting.ai_vllm_endpoint}/v1/chat/completions"
+            api_endpoint = llm_model.url
           end
+
           @uri ||= URI(api_endpoint)
         end
```
xfalcox, line 41: Didn't we talk about using special `srv://` detection instead of ENV?

Like 

```ruby
        def model_uri
          @uri ||=
            if !llm_model.url.starts_with?("srv://")
              URI(llm_model.url)
            else
              service = DiscourseAi::Utils::DnsSrv.lookup(llm_model.url)
              URI("https://#{service.target}:#{service.port}/v1/chat/completions")
            end
        end
```

romanrizzi: I forgot about that, sorry. Will change

xfalcox (approved):

PR #734, opened 2024-08-01
FIX: Add tool support to open ai compatible dialect and vllm

SamSaffron:
Automatic tools are in progress in vllm see: https://github.com/vllm-project/vllm/pull/5649

Even when they are supported, initial support will be uneven, only some models have native tool support
notably mistral which has some special tokens for tool support.

After the above PR lands in vllm we will still need to swap to
XML based tools on models without native tool support.


SamSaffron:
Still need to add a test to validate this stuff cause the existing test is not properly addressing this.

xfalcox (approved):

PR #740, opened 2024-08-02
UX: summary fixed positioning

chapoi:
This commit:

* positions the topic summary on desktop fixed to the right hand side
* increases the max-width and uses the same value for default width, avoiding jumpyness
* introduces a sticky header + close button

image


xfalcox (approved):

janzenisaac (approved):
Code looks good, you can drop function param as it is not being used :)

PR #747, opened 2024-08-09
DEV: Clearly separate post/composer helper settings

keegangeorge:
**This PR makes a clearer distinction for site settings related to the AI helper:**
- It renames the setting `composer_ai_helper_enabled` → `ai_helper_enabled`. This is more fitting as it is an overarching setting for enabling AI helper features
- It renames `ai_helper_allowed_groups` to `composer_ai_helper_allowed_groups` to make the setting unique for only composer AI helper features. Since we already have the separate `post_ai_helper_allowed_groups` this makes sense and allows for a better distinction. 

This PR also helps satisfy the use-case mentioned [here](https://meta.discourse.org/t/allow-ai-helper-for-reading-only/320871/2), where we can have separate control of the post AI helper and composer AI helper.

### DIFF
--- a/db/post_migrate/20240809162837_rename_ai_helper_enabled_setting.rb
+++ b/db/post_migrate/20240809162837_rename_ai_helper_enabled_setting.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class RenameComposerAiHelperEnabledSetting < ActiveRecord::Migration[7.1]
+class RenameAiHelperEnabledSetting < ActiveRecord::Migration[7.1]
   def up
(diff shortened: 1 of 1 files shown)


db/post_migrate/20240809162837_rename_ai_helper_enabled_setting.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RenameComposerAiHelperEnabledSetting < ActiveRecord::Migration[7.1]
```
romanrizzi, line 3: Class name is different here.

xfalcox (approved):

PR #756, opened 2024-08-16
FEATURE: Initial support for seeded LLMs

xfalcox:
LLM list

image

Edit

image

### DIFF
--- a/assets/stylesheets/modules/llms/common/ai-llms-editor.scss
+++ b/assets/stylesheets/modules/llms/common/ai-llms-editor.scss
@@ -43,7 +43,7 @@
   }
 
   &.seeded {
-    .control-group:nth-child(n+4) {
+    .control-group:nth-child(n + 4) {
       display: none;
     }
   }
--- a/assets/stylesheets/modules/llms/common/ai-llms-editor.scss
+++ b/assets/stylesheets/modules/llms/common/ai-llms-editor.scss
@@ -42,6 +42,12 @@
     }
   }
 
+  &.seeded {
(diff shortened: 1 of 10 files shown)


assets/stylesheets/modules/llms/common/ai-llms-editor.scss
```diff
@@ -33,6 +33,12 @@
     }
   }
 
+  &.seeded {
+    .control-group:nth-child(n + 4) {
+      display: none;
```
romanrizzi, line 38: Isn't it better to conditionally render in the template?

lib/automation.rb
```diff
@@ -8,7 +8,16 @@ def self.available_models
         FROM llm_models
       SQL
 
-      values.each { |value_h| value_h["id"] = "custom:#{value_h["id"]}" }
+      values =
+        values
+          .filter do |value_h|
+            value_h["id"] > 0 ||
+              SiteSetting
+                .ai_automation_allowed_seeded_models
```
romanrizzi, line 16:
```suggestion
                .ai_automation_allowed_seeded_models_map
```

I believe you can drop the `.split("|")` and use this helper method instead.

lib/configuration/llm_validator.rb
```diff
@@ -15,6 +15,8 @@ def valid_value?(val)
           return !@parent_enabled
         end
 
+        allowed_seeded_model?(val)
```
romanrizzi, line 18:
```suggestion
        allowed_seeded_model!(val)
```

PR #757, opened 2024-08-19
FEATURE: LLM Triage support for systemless models.

romanrizzi:
This change adds support for OSS models without support for system messages. LlmTriage's system message field is no longer mandatory. We now send the post contents in a separate user message.

### DIFF
--- a/app/models/llm_model.rb
+++ b/app/models/llm_model.rb
@@ -17,12 +17,17 @@ class LlmModel < ActiveRecord::Base
   def self.provider_params
     {
       aws_bedrock: {
-        url_editable: false,
-        fields: %i[access_key_id region],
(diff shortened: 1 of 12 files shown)


app/models/llm_model.rb
```diff
@@ -17,12 +17,17 @@ class LlmModel < ActiveRecord::Base
...
       open_ai: {
-        url_editable: true,
-        fields: %i[organization],
+        organization: :text,
+      },
+      hugging_face: {
+        disable_system_prompt: :checkbox,
+      },
+      vllm: {
+        disable_system_prompt: :checkbox,
```
xfalcox, line 30: Should do the same for ollama right ?

xfalcox (approved):

PR #769, opened 2024-08-23
FEATURE: single click proofreading

SamSaffron:
Previously there was too much work proofreading text, new implementation
provides a single shortcut and easy way of proofreading text.

### DIFF
--- a/assets/javascripts/discourse/components/modal/diff-modal.gjs
+++ b/assets/javascripts/discourse/components/modal/diff-modal.gjs
@@ -66,7 +66,10 @@ export default class ModalDiffModal extends Component {
     }
 
     if (this.args.model.toolbarEvent && this.suggestion) {
-      this.args.model.toolbarEvent.replaceText(this.selectedText, this.suggestion);
+      this.args.model.toolbarEvent.replaceText(
+        this.selectedText,
+        this.suggestion
+      );
     }
   }
 
--- a/assets/javascripts/discourse/components/modal/diff-modal.gjs
+++ b/assets/javascripts/discourse/components/modal/diff-modal.gjs
@@ -61,9 +61,7 @@ export default class ModalDiffModal extends Component {
   @action
   triggerConfirmChanges() {
     this.args.closeModal();
-    if (this.args.model.confirm) {
-      this.args.model.confirm();
-    }
+    this.args.model?.confirm();
 
     if (this.args.model.toolbarEvent && this.suggestion) {
       this.args.model.toolbarEvent.replaceText(
--- a/assets/javascripts/discourse/components/modal/diff-modal.gjs
+++ b/assets/javascripts/discourse/components/modal/diff-modal.gjs
@@ -23,7 +23,7 @@ export default class ModalDiffModal extends Component {
     this.diff = this.args.model.diff;
 
     next(() => {
-      if (this.args.model.selected && this.args.model.toolbarEvent) {
+      if (this.args.model.toolbarEvent) {
         this.loadDiff();
       }
     });
@@ -47,19 +47,21 @@ export default class ModalDiffModal extends Component {
   }
 
   get selectedText() {
-    const selected = this.args.model.selected;
-    
+    const selected = this.args.model.toolbarEvent.selected;
+
     if (selected.value === "") {
       return selected.pre + selected.post;
     }
-    
+
     return selected.value;
   }
 
   @action
   triggerConfirmChanges() {
     this.args.closeModal();
-    this.args.model?.confirm();
+    if (this.args.model.confirm) {
+      this.args.model.confirm();
+    }
 
     if (this.args.model.toolbarEvent && this.suggestion) {
       this.args.model.toolbarEvent.replaceText(
--- a/assets/javascripts/discourse/components/modal/diff-modal.gjs
+++ b/assets/javascripts/discourse/components/modal/diff-modal.gjs
@@ -1,15 +1,85 @@
 import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
 import { action } from "@ember/object";
+import { next } from "@ember/runloop";
+import { inject as service } from "@ember/service";
 import { htmlSafe } from "@ember/template";
+import CookText from "discourse/components/cook-text";
 import DButton from "discourse/components/d-button";
 import DModal from "discourse/components/d-modal";
+import { ajax } from "discourse/lib/ajax";
 import i18n from "discourse-common/helpers/i18n";
 
 export default class ModalDiffModal extends Component {
+  @service currentUser;
+  @tracked loading = false;
+  @tracked diff;
+  suggestion = "";
+
+  PROOFREAD_ID = -303;
+
+  constructor() {
+    super(...arguments);
+    this.diff = this.args.model.diff;
+
+    next(() => {
+      if (this.args.model.selected && this.args.model.toolbarEvent) {
+        this.loadDiff();
+      }
+    });
+  }
+
+  async loadDiff() {
+    this.loading = true;
+
+    const suggestion = await ajax("/discourse-ai/ai-helper/suggest", {
+      method: "POST",
+      data: {
+        mode: this.PROOFREAD_ID,
+        text: this.selectedText,
+        force_default_locale: true,
+      },
+    });
+
+    this.diff = suggestion.diff;
+    this.suggestion = suggestion.suggestions[0];
+    this.loading = false;
+  }
+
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/components/modal/diff-modal.gjs
```diff
@@ -1,15 +1,76 @@
...
+    return selection;
+  }
+
   @action
   triggerConfirmChanges() {
     this.args.closeModal();
-    this.args.model.confirm();
+    if (this.args.model.confirm) {
+      this.args.model.confirm();
+    }
```
martin-brennan, line 66:
```suggestion
    this.args.model?.confirm();
```

assets/javascripts/discourse/components/modal/diff-modal.gjs
```diff
@@ -25,30 +86,46 @@ export default class ModalDiffModal extends Component {
...
+            class="btn-primary"
+            @label="discourse_ai.ai_helper.context_menu.loading"
+            @disabled={{true}}
+          />
+        {{else}}
+          <DButton
+            class="btn-primary confirm"
+            @action={{this.triggerConfirmChanges}}
+            @label="discourse_ai.ai_helper.context_menu.confirm"
+          />
```
martin-brennan, line 120: Up to you but you don't necessarily need to duplicate the confirmation button, you could do something like this:

```hbs
<DButton
  class="btn-primary confirm"
  @action={{this.triggerConfirmChanges}}
  @label={{if this.loading "discourse_ai.ai_helper.context_menu.loading" "discourse_ai.ai_helper.context_menu.confirm"}}
  @disabled={{this.loading}}
/>
```

assets/javascripts/initializers/ai-helper.js
```diff
@@ -0,0 +1,38 @@
...
+
+function initializeProofread(api) {
+  api.addComposerToolbarPopupMenuOption({
+    action: (toolbarEvent) => {
+      const modal = api.container.lookup("service:modal");
+
+      modal.show(ModalDiffModal, {
+        model: {
+          selected: toolbarEvent.selected,
+          toolbarEvent,
```
martin-brennan, line 12: Why split selected out here when you are passing in the whole `toolbarEvent` object?

spec/system/ai_helper/ai_proofreading_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+RSpec.describe "AI Composer Proofreading Features", type: :system, js: true do
+  fab!(:user) { Fabricate(:admin) }
+
+  before do
+    Group.find_by(id: Group::AUTO_GROUPS[:admins]).add(user)
```
martin-brennan, line 7: No need for this --- change `Fabricate(:admin)` to `Fabricate(:admin, refresh_auto_groups: true)` above :)

spec/system/ai_helper/ai_proofreading_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+RSpec.describe "AI Composer Proofreading Features", type: :system, js: true do
+  fab!(:user) { Fabricate(:admin) }
```
martin-brennan, line 4:
```suggestion
  fab!(:current_user) { Fabricate(:admin) }
```

convention in system specs, makes things clearer

spec/system/ai_helper/ai_proofreading_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  let(:composer) { PageObjects::Components::Composer.new }
+
+  it "proofreads selected text using the composer toolbar" do
+    visit "/new-topic"
+    composer.fill_content("hello worldd !")
+
+    # Select a portion of text
+    page.execute_script("document.querySelector('.d-editor-input').setSelectionRange(6, 12);")
```
martin-brennan, line 20: I thought we had a page object method for this but we didn't, added here:

https://github.com/discourse/discourse/pull/28539

spec/system/ai_helper/ai_proofreading_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    DiscourseAi::Completions::Llm.with_prepared_responses(["world"]) do
+      ai_toolbar = PageObjects::Components::SelectKit.new(".toolbar-popup-menu-options")
+      ai_toolbar.expand
+      ai_toolbar.select_row_by_name("Proofread Text")
+
+      find(".composer-ai-helper-modal .btn-primary.confirm").click
```
martin-brennan, line 27: You can use the generic modal page object for this:

```ruby
let(:modal) { PageObjects::Modals::Base.new }
modal.click_primary_button
```

assets/javascripts/discourse/components/modal/diff-modal.gjs
```diff
@@ -1,15 +1,76 @@
...
+      if (this.args.model.selected && this.args.model.toolbarEvent) {
+        this.loadDiff();
+      }
+    });
+  }
+
+  async loadDiff() {
+    this.loading = true;
+
+    const suggestion = await ajax("/discourse-ai/ai-helper/suggest", {
```
martin-brennan, line 35: Should wrap async ajax in try/catch with `popupAjaxError` for failures and reset this.loading in `finally`

assets/javascripts/discourse/components/modal/diff-modal.gjs
```diff
@@ -1,15 +1,76 @@
...
+  get selectedText() {
+    const selected = this.args.model.selected;
+    let selection;
+    if (selected.value === "") {
+      selection = selected.pre + selected.post;
+    } else {
+      selection = selected.value;
+    }
+
+    return selection;
```
martin-brennan, line 58:
```suggestion
    const selected = this.args.model.selected;
    
    if (selected.value === "") {
      return selected.pre + selected.post;
    }
    
    return selected.value;
```

martin-brennan (approved):
I wonder if this modal title could be clearer? Something like "Suggested changes" or "Suggested edits"

image

Other things are minor/code comments so approving since it works as described and tested :+1:

assets/javascripts/initializers/ai-helper.js
SamSaffron: yeah this modal is used in 2 spots, I can clean this up :+1:

spec/system/ai_helper/ai_proofreading_spec.rb
SamSaffron: will clean this up.

assets/javascripts/discourse/components/modal/diff-modal.gjs
renato: I believe the equivalent would be `this.args.model.confirm?.()`

spec/system/ai_helper/ai_proofreading_spec.rb
SamSaffron: did not seem to work I just left as is for now.

assets/javascripts/discourse/components/modal/diff-modal.gjs
SamSaffron: hmmm yeah it can work, I wanted to leave the modal though as is as much as possible for the other consumers, leaving this for now

assets/javascripts/discourse/components/modal/diff-modal.gjs
martin-brennan: Oh yeah you're right, good catch Renato!

SamSaffron: yeah I don't think this syntax is allowed? even if it was, it feels a bit too weird visually

martin-brennan: It is allowed, but yes in this case it's too weird, I was thinking of the case where `model` might be null when doing this comment. Here is an example of it working:

```javascript
let x = { y: { z: function () { console.log("alert"); } } }
x.y.z?.()
=> alert
```

renato: Yeah it's part of the optional chaining spec and widely supported.

The difference is that `this.args.model?.confirm()` will fail with a present model but absent `confirm` fn, while `this.args.model.confirm?.()` is exactly the same as the original code, checking only for the fn presence

I don't find it too weird (ok, ok, I even like it! 😄), I feel like parsing it is the same as any optional chaining

PR #773, opened 2024-08-27
DEV: Update ace-editor usage

CvX:
AceEditor is now a glimmer component (see: https://github.com/discourse/discourse/pull/28492) and it follows the "data down, actions up" pattern.


davidtaylorhq (approved):

romanrizzi:
@CvX - Is this backwards-compatible, or do we need to update `.discourse-compatibility`?

CvX:
It's not _strictly_ backward compatible 😅 but there already is a `< 3.4.0.beta1-dev` (i.e. the current core version) entry in .d-compat file, so I think this covers it?

PR #777, opened 2024-08-28
FEATURE: allow embedding based search without hyde

SamSaffron:
This allows callers of embedding based search to bypass hyde.

Hyde will expand the search term using an LLM, but if an LLM is
performing the search we can skip this expansion.

It also introduced some tests for the controller which we did not have

### DIFF
--- a/lib/embeddings/semantic_search.rb
+++ b/lib/embeddings/semantic_search.rb
@@ -30,6 +30,43 @@ module DiscourseAi
         Discourse.cache.read(embedding_key).present?
       end
 
+      def vector_rep
+        @vector_rep ||=
+          DiscourseAi::Embeddings::VectorRepresentations::Base.current_representation(
+            DiscourseAi::Embeddings::Strategies::Truncation.new,
+          )
+      end
+
+      def hyde_embedding(search_term)
+        digest = OpenSSL::Digest::SHA1.hexdigest(search_term)
+        hyde_key = build_hyde_key(digest, SiteSetting.ai_embeddings_semantic_search_hyde_model)
+
+        embedding_key =
+          build_embedding_key(
+            digest,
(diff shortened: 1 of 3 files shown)


lib/embeddings/semantic_search.rb
```diff
@@ -43,28 +44,42 @@ def search_for_topics(query, page = 1)
...
+                vector_rep.vector_from(hypothetical_post)
+              end
+        else
+          embedding_key = build_embedding_key(digest, "", SiteSetting.ai_embeddings_model)
+
+          search_embedding =
+            Discourse
+              .cache
+              .fetch(embedding_key, expires_in: 1.week) { vector_rep.vector_from(search_term) }
+        end
```
lis2, line 78: I think that this method is having too much responsibility. What would you say about extracting 2 private methods like `search_embedding_with_hyde` and `search_embedding_without_hyde`. 

Then:
```ruby
search_embedding = if hyde
  search_embedding_with_hyde
else
  search_embedding_without_hyde
end
```

lib/embeddings/semantic_search.rb
```diff
@@ -43,28 +44,42 @@ def search_for_topics(query, page = 1)
...
 
         candidate_topic_ids =
           vector_rep.asymmetric_topics_similarity_search(
-            hypothetical_post_embedding,
+            search_embedding,
```
lis2, line 82: Or even we can totally skip assignment and go with
```ruby
hyde ? search_embedding_with_hyde : search_embedding_without_hyde
```

lis2 (approved):

lib/embeddings/semantic_search.rb
SamSaffron: I will split it up a bit, no probs

PR #778, opened 2024-08-28
FEATURE: smooth streaming animation for summarization

keegangeorge:
This PR adds a smoother streaming animation to summarization, similar to what was added for AI bot in: https://github.com/discourse/discourse-ai/pull/413.

Streaming/animations here are difficult to test so no specs added here. But, we are using the same interface built in https://github.com/discourse/discourse-ai/pull/413, hence, there are already unit tests in place.


romanrizzi (approved):
Looks good to me!

PR #780, opened 2024-08-29
FEATURE: improve site setting search

SamSaffron:
This improves the site setting search so it performs a somewhat
fuzzy match.

Previously it did not handle seperators such as "space" and a
term such as "min_post_length" would not find "min_first_post_length"

A more liberal search algorithm makes it easier to the AI to
navigate settings.

### DIFF
--- a/spec/lib/modules/ai_bot/tools/search_settings_spec.rb
+++ b/spec/lib/modules/ai_bot/tools/search_settings_spec.rb
@@ -5,13 +5,28 @@ RSpec.describe DiscourseAi::AiBot::Tools::SearchSettings do
   let(:bot_user) { DiscourseAi::AiBot::EntryPoint.find_user_from_model(llm_model.name) }
   let(:llm) { DiscourseAi::Completions::Llm.proxy("custom:#{llm_model.id}") }
 
+  let(:fake_settings) do
+    [
+      { setting: "default_locale", description: "The default locale for the site", plugin: "core" },
+      { setting: "min_post_length", description: "The minimum length of a post", plugin: "core" },
+      {
+        setting: "ai_bot_enabled",
+        description: "Enable or disable the AI bot",
+        plugin: "discourse-ai",
+      },
+      { setting: "min_first_post_length", description: "First post length", plugin: "core" },
+    ]
+  end
+
   before do
     SiteSetting.ai_bot_enabled = true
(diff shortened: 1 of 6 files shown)


spec/lib/modules/ai_bot/tools/search_settings_spec.rb
```diff
@@ -5,13 +5,28 @@
...
   before do
     SiteSetting.ai_bot_enabled = true
     toggle_enabled_bots(bots: [llm_model])
   end
 
-  def search_settings(query)
-    described_class.new({ query: query }, bot_user: bot_user, llm: llm)
+  def search_settings(query, mock: true)
+    search = described_class.new({ query: query }, bot_user: bot_user, llm: llm)
+    search.all_settings = fake_settings if mock
```
martin-brennan, line 28: I think you need an `after` block to restore the original value of this, or reset it to nil so the memoization can kick in and re-initialize the settings in [search_settings.rb](https://github.com/discourse/discourse-ai/pull/780/files#diff-2f1ab10bdb1709b638c2b05ff7f7c0940fc62db89332de4c429b76570f294f90)

martin-brennan: Otherwise some test flakiness may occur

martin-brennan: An alternative approach is to do `DiscourseAi::AiBot::Tools::SearchSettings.any_instance.stubs(:site_settings).returns(mock)`, then that will automatically get cleaned up

martin-brennan: Or doing the above with `SiteSetting.all_settings` would avoid needing to add an `all_settings` method in that class entirely :)

martin-brennan (approved):
Approving with comments about stubbing all settings

spec/lib/modules/ai_bot/tools/search_settings_spec.rb
SamSaffron: no need, this is just defining an instance :) the instance and the override vanish after the test

martin-brennan: Oh wow duh :facepalm:

PR #787, opened 2024-08-30
FIX: coerce value before downcasing the hyde param

romanrizzi:
Fixes 
```
NoMethodError (undefined method `downcase' for nil)
```


SamSaffron:
Thanks 🙏

PR #788, opened 2024-08-30
FIX: Correctly display errors when parent module needs to be disabled first

romanrizzi:
Fixes:

```
NameError (undefined local variable or method `parent_module_name' for an instance of DiscourseAi::Configuration::LlmValidator)
```

### DIFF
--- a/spec/configuration/llm_validator_spec.rb
+++ b/spec/configuration/llm_validator_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+RSpec.describe DiscourseAi::Configuration::LlmValidator do
+  describe "#valid_value?" do
+    context "when the parent module is enabled and we try to reset the selected model" do
+      before do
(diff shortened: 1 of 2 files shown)


spec/configuration/llm_validator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+RSpec.describe DiscourseAi::Configuration::LlmValidator do
+  describe "#valid_value?" do
+    context "when the parent module is enabled and we try to reset the selected model" do
+      before do
+        assign_fake_provider_to(:ai_summarization_model)
+        SiteSetting.ai_summarization_enabled = true
+      end
+
+      it "returns false and display an error message" do
```
pmusaraj, line 11:
```suggestion
      it "returns false and displays an error message" do
```

pmusaraj (approved):

PR #791, opened 2024-09-03
FEATURE: more accurate and faster titles

SamSaffron:
Previously we waited 1 minute before automatically titling PMs

The new change introduces adding a title immediately after the the
llm replies

Prompt was also modified to include the LLM reply in title suggestion.

This helps situation like:

user: tell me a joke
llm: a very funy joke about horses

Then the title would be "A Funny Horse Joke"

Specs already covered some auto title logic, amended to also
catch the new message bus message we have been sending.


martin-brennan (approved):
Nice I like it, changing the [untitled bot pm] thing feels less pressing if the titling is instant :+1:

PR #793, opened 2024-09-04
FEATURE: disable rate limiting when skipping hyde

SamSaffron:
Embedding search is rate limited due to potentially expensive
hyde operation (which require LLM access).

Embedding generally is very cheap compared to it. (usually 100x cheaper)

This raises the limit to 100 per minute for embedding searches,
while keeping the old 4 per minute for HyDE powered search.

### DIFF
--- a/spec/requests/embeddings/embeddings_controller_spec.rb
+++ b/spec/requests/embeddings/embeddings_controller_spec.rb
@@ -60,6 +60,19 @@ describe DiscourseAi::Embeddings::EmbeddingsController do
       expect(response.parsed_body["topics"].map { |t| t["id"] }).to contain_exactly(topic.id)
     end
 
+    context "when rate limiting is enabled" do
+      before { RateLimiter.enable }
+
+      it "will not rate limit API for hyde search" do
+        10.times do |i|
+          query = "test #{SecureRandom.hex}"
+          stub_embedding(query)
+          get "/discourse-ai/embeddings/semantic-search.json?q=#{query}&hyde=false"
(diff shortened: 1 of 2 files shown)


spec/requests/embeddings/embeddings_controller_spec.rb
```diff
@@ -60,6 +60,19 @@ def create_api_key(user)
       expect(response.parsed_body["topics"].map { |t| t["id"] }).to contain_exactly(topic.id)
     end
 
+    context "when rate limiting is enabled" do
+      before { RateLimiter.enable }
+
+      it "will not rate limit API for hyde search" do
```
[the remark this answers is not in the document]
SamSaffron: true :)

spec/requests/embeddings/embeddings_controller_spec.rb
```diff
@@ -60,6 +60,19 @@ def create_api_key(user)
...
+        10.times do |i|
+          query = "test #{SecureRandom.hex}"
+          stub_embedding(query)
+          get "/discourse-ai/embeddings/semantic-search.json?q=#{query}&hyde=false"
+          expect(response.status).to eq(200)
+        end
```
tgxworld, line 72: Maybe we can stub the `MAX_HYDE_SEARCHES_PER_MINUTE ` constant so that we don't have to trigger 10 requests here?

SamSaffron: ill time the test and see if it is worth a follow up

PR #795, opened 2024-09-10
FIX: support case insensitive setting lookup

SamSaffron:


martin-brennan (approved):
Very insensitive of you Sam :P

PR #796, opened 2024-09-11
DEV: Move composer AI helper to toolbar

keegangeorge:
### 🔍 Overview
Previously we had moved the AI helper from the options menu to a selection menu that appears when selecting text in the composer. This had the benefit of making the AI helper a more discoverable feature. Now that some time has passed and the AI helper is more recognized, we will be moving it back to the composer toolbar. 

**This is better because:**
- It consistent with other behavior and ways of accessing tools in the composer
- It has an improved mobile experience
- It reduces unnecessary code and keeps things easier to migrate when we have composer V2.
- It allows for easily triggering AI helper for _all content_ by clicking the button instead of having to select everything.

#### Before:
Screenshot 2024-09-13 at 09 27 47
Screenshot 2024-09-13 at 09 27 39

#### After:
Screenshot 2024-09-13 at 09 28 13

### DIFF
--- a/assets/javascripts/initializers/ai-helper.js
+++ b/assets/javascripts/initializers/ai-helper.js
@@ -4,17 +4,18 @@ import AiComposerHelperMenu from "../discourse/components/ai-composer-helper-men
 import ModalDiffModal from "../discourse/components/modal/diff-modal";
 import { showComposerAiHelper } from "../discourse/lib/show-ai-helper";
 
-function showErrorToast(api) {
-  const toasts = api.container.lookup("service:toasts");
-  return toasts.error({
-    duration: 3000,
-    data: {
-      message: i18n("discourse_ai.ai_helper.no_content_error"),
-    },
-  });
-}
-
 function initializeAiHelperTrigger(api) {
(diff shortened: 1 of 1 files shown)


assets/javascripts/initializers/ai-helper.js
```diff
@@ -1,41 +1,85 @@
...
 import ModalDiffModal from "../discourse/components/modal/diff-modal";
+import { showComposerAiHelper } from "../discourse/lib/show-ai-helper";
 
-function initializeProofread(api) {
-  api.addComposerToolbarPopupMenuOption({
-    action: (toolbarEvent) => {
-      const modal = api.container.lookup("service:modal");
-      const composer = api.container.lookup("service:composer");
-      const toasts = api.container.lookup("service:toasts");
+function showErrorToast(api) {
```
pmusaraj, line 7: Super minor, can the `showErrorToast` helper live inside the initializer, that way we won't need to pass the whole `api` object to it every time it's invoked?

pmusaraj (approved):
Looks good to me!

PR #798, opened 2024-09-12
FEATURE: Make tool support polymorphic

SamSaffron:
Polymorphic RAG means that we will be able to access RAG fragments both from AiPersona and AiCustomTool

In turn this gives us support for richer RAG implementations. 

You can define a custom tool with custom rules for querying the RAG. 

So something like

- Query File1 for 3 results
- Query File10 for 20 results 

Becomes feasible in a tool. 

This allows people implementing a RAG extreme flexibility.

### DIFF
--- a/app/jobs/regular/digest_rag_upload.rb
+++ b/app/jobs/regular/digest_rag_upload.rb
@@ -9,17 +9,24 @@ module ::Jobs
     # TODO(roman): Add a way to automatically recover from errors, resulting in unindexed uploads.
     def execute(args)
       return if (upload = Upload.find_by(id: args[:upload_id])).nil?
-      return if (ai_persona = AiPersona.find_by(id: args[:ai_persona_id])).nil?
+
+      target_type = args[:target_type]
+      target_id = args[:target_id]
+
+      return if !target_type || !target_id
+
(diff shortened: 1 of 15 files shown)


app/jobs/regular/digest_rag_upload.rb
```diff
@@ -9,17 +9,24 @@ class DigestRagUpload < ::Jobs::Base
...
     def execute(args)
       return if (upload = Upload.find_by(id: args[:upload_id])).nil?
-      return if (ai_persona = AiPersona.find_by(id: args[:ai_persona_id])).nil?
+
+      target_type = args[:target_type]
+      target_id = args[:target_id]
+
+      return if !target_type || !target_id
+
+      target = target_type.constantize.find_by(id: target_id)
```
romanrizzi, line 18: Maybe `rescue nil` here to be safe? `constantize` could return `NameError`.

romanrizzi (approved):

app/jobs/regular/digest_rag_upload.rb
SamSaffron: the tricky thing here is that this would be incorrect usage, I don't want to eat the error of a typo... the "blank target_type" is just there for the transition period.

PR #802, opened 2024-09-16
FEATURE: RAG search within tools

SamSaffron:
This allows custom tools access to uploads and sophisticated searches using embedding.

It introduces: 

1. A shared front end for listing and uploading files (shared with personas)
2. Backend implementation of `index.search` function within a custom tool.

Custom tools now may search through uploaded files 

```
function invoke(params) {
   return index.search(params.query)
}
```

This means that RAG implementers now may preload tools with knowledge and have high fidelity over
the search. 

The search function support 

1. specifying max results
2. specifying a subset of files to search (from uploads) 

Also 

- Improved documentation for tools (when creating a tool a preamble explains all the functionality)
- uploads were a bit finicky, fixed an edge case where the UI would not show them as updated

### DIFF
--- a/app/controllers/discourse_ai/admin/ai_tools_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_tools_controller.rb
@@ -17,10 +17,11 @@ module DiscourseAi
       end
 
       def create
-        ai_tool = AiTool.new(ai_tool_params)
+        ai_tool = AiTool.new(ai_tool_params.except(:rag_uploads))
         ai_tool.created_by_id = current_user.id
 
         if ai_tool.save
+          RagDocumentFragment.link_target_and_uploads(ai_tool, attached_upload_ids)
           render_serialized(ai_tool, AiCustomToolSerializer, status: :created)
         else
           render_json_error ai_tool
@@ -28,7 +29,8 @@ module DiscourseAi
       end
 
       def update
-        if @ai_tool.update(ai_tool_params)
+        if @ai_tool.update(ai_tool_params.except(:rag_uploads))
+          RagDocumentFragment.update_target_uploads(@ai_tool, attached_upload_ids)
           render_serialized(@ai_tool, AiCustomToolSerializer)
         else
           render_json_error @ai_tool
@@ -71,6 +73,10 @@ module DiscourseAi
 
       private
 
+      def attached_upload_ids
+        ai_tool_params[:rag_uploads].to_a.map { |h| h[:id] }
+      end
+
       def find_ai_tool
         @ai_tool = AiTool.find(params[:id])
       end
@@ -81,6 +87,7 @@ module DiscourseAi
           :description,
           :script,
           :summary,
+          rag_uploads: [:id],
           parameters: [:name, :type, :description, :required, enum: []],
         )
       end
--- a/app/controllers/discourse_ai/admin/ai_tools_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_tools_controller.rb
@@ -87,6 +87,8 @@ module DiscourseAi
           :description,
           :script,
           :summary,
+          :rag_chunk_tokens,
+          :rag_chunk_overlap_tokens,
           rag_uploads: [:id],
           parameters: [:name, :type, :description, :required, enum: []],
         )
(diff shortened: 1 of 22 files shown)


assets/javascripts/discourse/components/rag-options.gjs
```diff
@@ -0,0 +1,70 @@
...
+      : I18n.t("discourse_ai.rag.options.show_indexing_options");
+  }
+
+  <template>
+    {{#if @model.rag_uploads}}
+      <a
+        href="#"
+        class="rag-options__indexing-options"
+        {{on "click" this.toggleIndexingOptions}}
+      >{{this.indexingOptionsText}}</a>
```
jjaffeux, line 31: Can we use a button here? It's a lot of code just because we try to use an anchor for something which is not an anchor

app/controllers/discourse_ai/admin/ai_tools_controller.rb
```diff
@@ -17,18 +17,20 @@ def show
...
+          RagDocumentFragment.link_target_and_uploads(ai_tool, attached_upload_ids)
           render_serialized(ai_tool, AiCustomToolSerializer, status: :created)
         else
           render_json_error ai_tool
         end
       end
 
       def update
-        if @ai_tool.update(ai_tool_params)
+        if @ai_tool.update(ai_tool_params.except(:rag_uploads))
```
jjaffeux, line 32: Im surprise a lot of this code is using non ! methods. same for save above.

What happens if `@ai.tool.update` fails?

app/models/ai_tool.rb
```diff
@@ -173,14 +257,16 @@ def self.presets
...
-#  created_by_id :integer          not null
-#  enabled       :boolean          default(TRUE), not null
-#  created_at    :datetime         not null
-#  updated_at    :datetime         not null
+#  id                       :bigint           not null, primary key
+#  name                     :string           not null
+#  description              :string           not null
+#  summary                  :string           not null
+#  parameters               :jsonb            not null
+#  script                   :text             not null
```
jjaffeux, line 265: should we limit the size of these fields?

spec/requests/admin/rag_document_fragments_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  describe "POST #upload_file" do
+    it "works" do
+      post "/admin/plugins/discourse-ai/rag-document-fragments/files/upload.json",
+           params: {
+             file: Rack::Test::UploadedFile.new(file_from_fixtures("spec.txt", "md")),
+           }
+
+      expect(response.status).to eq(200)
```
jjaffeux, line 30: This doesn't test much, maybe we should at least ensure something has been saved or whatever. I understand it's a controller test and are testing the status and errors, but I like to at least tet something in the happy path to ensure Im not just testing peanuts.

assets/javascripts/discourse/components/rag-uploader.gjs
```diff
@@ -12,18 +12,16 @@ import discourseDebounce from "discourse-common/lib/debounce";
 import I18n from "discourse-i18n";
 import RagUploadProgress from "./rag-upload-progress";
 
-export default class PersonaRagUploader extends Component.extend(
-  UppyUploadMixin
-) {
+export default class RagUploader extends Component.extend(UppyUploadMixin) {
```
jjaffeux, line 15: @davidtaylorhq FYI as you are working on this

jjaffeux (approved):
I don't have enough knowledge to have a very strong opinion, but I saw nothing stopping to merge this.

assets/javascripts/discourse/components/rag-options.gjs
SamSaffron: probably, will leave this to @awesomerobot to address in a followup, he probably wants to rethink some of the design here.

app/controllers/discourse_ai/admin/ai_tools_controller.rb
SamSaffron: yeah it works, it does a render json error that consumes the error.

spec/requests/admin/rag_document_fragments_controller_spec.rb
SamSaffron: looking at this one

SamSaffron: note its an old test that was there before

PR #807, opened 2024-09-18
DEV: Make indicator wave a reusable component

keegangeorge:
### 🔍 Overview
Previously we had some hardcoded markup with scss making a loading indicator wave. This code was being duplicated and used in both semantic search and summarization. We want to add the indicator wave to the AI helper diff modal as well and have the text flashing instead of the loading spinner. To ensure we do not repeat ourselves, in this PR we turn the summary indicator wave into a reusable template only component called: `AiIndicatorWave`. We then apply the usage of that component to semantic search, summarization, and the composer helper modal.

There are no visual changes seen for semantic search and summarization, but visually we will see changes for the composer helper modal's loading state. These visual changes are reflected in the screen captures below.

### 📸 Screen Recordings

_Before_

https://github.com/user-attachments/assets/a1b993a9-d418-4112-9b51-9bbb8ae92927

_After_

https://github.com/user-attachments/assets/8acbbf63-dc86-4205-944d-8f5b22fe94cb

### DIFF
--- a/assets/javascripts/discourse/components/ai-indicator-wave.gjs
+++ b/assets/javascripts/discourse/components/ai-indicator-wave.gjs
@@ -1,9 +1,8 @@
-import { array } from "@ember/helper";
-
+const indicatorDots = [".", ".", "."];
 const AiIndicatorWave = <template>
   {{#if @loading}}
     <span class="ai-indicator-wave">
-      {{#each (array "." "." ".") as |dot|}}
+      {{#each indicatorDots as |dot|}}
         <span class="ai-indicator-wave__dot">{{dot}}</span>
       {{/each}}
     </span>


assets/javascripts/discourse/components/ai-indicator-wave.gjs
```diff
@@ -0,0 +1,13 @@
+import { array } from "@ember/helper";
+
+const AiIndicatorWave = <template>
+  {{#if @loading}}
+    <span class="ai-indicator-wave">
+      {{#each (array "." "." ".") as |dot|}}
```
janzenisaac, line 6: 🏌️ but we could just define the array at the top level and then drop it into `#each`. That way we don't need the `array` helper

janzenisaac (approved):

janzenisaac (approved):

PR #810, opened 2024-09-22
DEV: remove ignore column syntax for the removed provider column in completion prompt model

nvh0412:
This provider column was removed 10 months ago in this PR https://github.com/discourse/discourse-ai/pull/312, there is no usage of this attribute in our codebase, so there is no potential danger to remove this ignored column.


SamSaffron:
thanks, looks good!

SamSaffron (approved):

PR #816, opened 2024-10-01
DEV: Update rate limits for image captioning

keegangeorge:
### 🔍 Overview
This PR updates the rate limits for AI helper so that image caption follows a specific rate limit of 20 requests per minute. This should help when uploading multiple files that need to be captioned. This PR also updates the UI so that it shows toast message with the extracted error message instead of having a blocking `popupAjaxError` error dialog.

### 📸 Screenshots
Screenshot 2024-10-01 at 16 08 09

### DIFF
--- a/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
+++ b/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
@@ -11,8 +11,14 @@ module DiscourseAi
       include SecureUploadEndpointHelpers
 
       RATE_LIMITS = {
-        "default" => { amount: 6, interval: 3.minutes },
-        "caption_image" => { amount: 20, interval: 1.minute }
+        "default" => {
+          amount: 6,
+          interval: 3.minutes,
+        },
+        "caption_image" => {
+          amount: 20,
+          interval: 1.minute,
+        },
       }.freeze
 
       def suggest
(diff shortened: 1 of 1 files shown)


app/controllers/discourse_ai/ai_helper/assistant_controller.rb
```diff
@@ -161,7 +161,11 @@ def get_post_param!
       end
 
       def rate_limiter_performed!
-        RateLimiter.new(current_user, "ai_assistant", 6, 3.minutes).performed!
+        if action_name == "caption_image"
+          RateLimiter.new(current_user, "ai_assistant_caption_image", 20, 1.minutes).performed!
```
pmusaraj, line 165:
```suggestion
          RateLimiter.new(current_user, "ai_assistant_caption_image", 20, 1.minute).performed!
```

Tiny detail. (Both work, but `1.minute` reads better.)

config/locales/client.en.yml
```diff
@@ -352,6 +352,7 @@ en:
             prompt: "This post contains non-captioned images. Would you like to enable automatic AI captions on image uploads? (This can be changed in your preferences later)"
             confirm: "Enable"
             cancel: "Don't ask again"
+          error: "There was an error generating a caption for the image."
```
pmusaraj, line 355: We should still surface the rate limit error. Better for users to know what happened, no? 

Also, does the switch to toasts mean that upload processing isn't blocked now?

pmusaraj (approved):

config/locales/client.en.yml
xfalcox: Yeah, if we now have toasts, we can show the actual "You have excedeed your AI caption limit. Wait 1 minute and try again" message.

xfalcox (approved):

PR #819, opened 2024-10-02
FEATURE: Tools for models from Ollama provider

nvh0412:
This pull request adds support for native tools and tool calls in the Ollama dialect, including the implementation of the `OllamaTools` class and its integration into existing methods. The changes also include test coverage for the new functionality.

NOTE: regarding the ollama API doc, we would have to pass `stream: false` if we want to use `tools` https://github.com/ollama/ollama/blob/main/docs/api.md#parameters-1, as a result, we can't utilise `@streaming_mode` here but have to do the following implementation:

- native tool support should be optional
- totally give up streaming to get the correct behavior when we use native tool with the Ollama provider

## Demo

Screenshot 2024-10-09 at 4 45 01 PM

Screenshot 2024-10-02 at 10 39 40 PM

https://github.com/user-attachments/assets/9a689b6b-95ff-4347-af09-4ba164786baa

### DIFF
--- a/app/models/llm_model.rb
+++ b/app/models/llm_model.rb
@@ -31,6 +31,7 @@ class LlmModel < ActiveRecord::Base
       },
       ollama: {
         disable_system_prompt: :checkbox,
+        enable_native_tool: :checkbox,
       },
     }
   end
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -275,7 +275,6 @@ de:
           region: "AWS-Bedrock-Region"
           organization: "Optionale OpenAI-Organisations-ID"
           disable_system_prompt: "Systemmeldung in Eingabeaufforderungen deaktivieren"
-          enable_native_tool: "Nativen Tool aktivieren"
       related_topics:
         title: "Verwandte Themen"
         pill: "Verwandt"
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -275,6 +275,7 @@ de:
           region: "AWS-Bedrock-Region"
           organization: "Optionale OpenAI-Organisations-ID"
           disable_system_prompt: "Systemmeldung in Eingabeaufforderungen deaktivieren"
+          enable_native_tool: "Nativen Tool aktivieren"
       related_topics:
         title: "Verwandte Themen"
         pill: "Verwandt"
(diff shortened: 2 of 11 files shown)


SamSaffron:
This is not ideal, I would say a couple of things here: 

1. native tool support should be optional, like we have other optional llm features (see llm_model.lookup_custom_param) 
2. I think this is too limiting cause you are forcing it so user can only use one tool, what if the llm wants to call a tool on the second turn? I think you need to totally give up streaming to get the correct behavior.

SamSaffron:
btw :wave: from Sydney!

nvh0412:
Thanks, @SamSaffron, I'm still in limbo actually, that's why I put that note in the description as it depends on your product vision as well. But if you said so, I'm on board with the idea that we can totally give up streaming to get the proper behaviour. It would also help us eliminate the trick, which, in my experience, is quite difficult to debug (in the case when we thought it was a streaming call but not).

SamSaffron:
I still stand behind thinking this should be optional, we get tons of mileage from XML tools, so it is nice to choose if you prefer XML tools and streaming vs native tools and not streaming.

nvh0412:
will do my homework about XML tools and cycle back this PR next week, cheers.

nvh0412:
Hey @SamSaffron, I’ve made the native tool for the Ollama provider optional. Let me know if you think the provider parameter key needs any adjustments.

SamSaffron:
Minor, can you only include English translations, we have an automatic process for translation so this will just cause confusion, plus Hebrew translation is way off anyway :)

SamSaffron:
copy is a bit confusing... let's go with "Enable native tool support" - a bit more verbose but copy is a lot clearer.

SamSaffron:
overall change looks good, only those minor bits need improving.

SamSaffron:
OH I think we are missing something here and we should add tests. 

If native tools are disabled you should still get tool support via xml tools, we should fall back to it.

SamSaffron (approved):

SamSaffron:
Thanks heaps! 

One thing I am wondering about, there are just so many providers these days with OpenAI "sort of compatible" endpoints. I wonder if we just go with a single "OpenAI Compatible" dialect / endpoint and then add enough options to make it work across the board, (eg: together.ai / grok / ollama ... etc...)

nvh0412:
Yes, the more I work with these providers, the more I notice they share the same API interface. It might be time to revisit and refactor their endpoints, dialects and tool classes. However, we'll need to add more tests before proceeding, I think.

PR #828, opened 2024-10-10
UX: Remove AdminPageSubheader style override

martin-brennan:
In this core commit https://github.com/discourse/discourse/pull/29149 we
are changing the subheader title to H2 and making the size smaller,
this style override is no longer needed.


martin-brennan:
test flake

PR #831, opened 2024-10-11
Proof of concept of Discord Bot integration

xfalcox:


lib/discord/bot/persona_replier.rb
```diff
@@ -0,0 +1,55 @@
...
+        @persona =
+          AiPersona
+            .all_personas
+            .find { |persona| persona.id == SiteSetting.ai_discord_search_persona.to_i }
+            .new
+        @bot =
+          DiscourseAi::AiBot::Bot.as(
+            Discourse.system_user,
+            persona: @persona,
+            model: AiPersona.find(@persona.id).default_llm, # TODO this is weird
```
romanrizzi, line 16: No need to do this. We guess the model using the persona when `model` is `nil`

plugin.rb
```diff
@@ -10,6 +10,7 @@
 
 gem "tokenizers", "0.4.4"
 gem "tiktoken_ruby", "0.0.9"
+gem "ed25519", "1.2.4" #TODO remove this as existing ssl gem should handle this
```
pmusaraj, line 13: Is this resolved now?

PR #834, opened 2024-10-14
FEATURE: Add a new setting to disable the automatic topic title feature of the AI bot

nvh0412:
I’m on the fence about this implementation, actually. Given that we’re about to implement a quota system to limit AI tokens, it makes sense to me to allow the Admin to prevent the AI bot from automatically updating the title of the inbox topic, as it does consume a significant number of tokens.

In this PR, we add a new setting with the default set to true to maintain backward compatibility with the current behavior. We can utilize this setting to prevent our Playground from automatically generating the title right after it finishes the reply_to method.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -87,6 +87,7 @@ en:
 
     ai_bot_enabled: "Enable the AI Bot module."
     ai_bot_enable_chat_warning: "Display a warning when PM chat is initiated. Can be overriden by editing the translation string: discourse_ai.ai_bot.pm_warning"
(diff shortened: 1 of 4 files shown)


SamSaffron:
I worry about having too many knobs sometimes. 

Maybe as a middle-ground here we just make this a hidden setting, so you can disable if you really need it. 

I have not seen this come up yet on meta.discourse.org , in the big scheme the title generator uses many orders of magnitude less tokens.

nvh0412:
Yeah, that's why I'm still here, on the fence. I think I can close this PR for now and revisit it when it’s actually needed.

PR #835, opened 2024-10-14
REFACTOR: Support of different summarization targets/prompts.

romanrizzi:
This change changes the internals of how we generate summaries to better support summarizing the same entities using different prompts and picking different parts of it (`strategies`). `FoldContent` is now the entry point of our summarization engine, which leverages these strategies to switch between different prompts and content sources, allowing us to generate different summary types for the same content.

In addition to this, we added a `summary_type` column to AiSummary, allowing us to have different summaries for the same target, allowing them to have other sources and lifecycles.

To showcase this change, I added a new strategy that lets us summarize a topic's OP and the last 20 replies into a single sentence called `TopicGist`. We plan to use it to give users a grasp of what's going on in a topic when they visit a topic list.


db/migrate/20240909180908_add_ai_summary_type_column.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddAiSummaryTypeColumn < ActiveRecord::Migration[7.1]
+  def change
+    add_column :ai_summaries, :summary_type, :string
```
xfalcox, line 4: I wonder if we should use an enum here :thinking:

romanrizzi: Much better, yes

xfalcox (approved):

PR #837, opened 2024-10-15
FEATURE: Generate topic gists for the hot topics list.

romanrizzi:
This PR builds on top of the changes introduced in #835, polishing the strategy for hot topics by improving the query and prompts used.

It also adds a job that processes N hot topics to create gists, where N is `SiteSetting#ai_summarize_max_hot_topics_gists_per_batch`. These summaries are included in the `TopicList` when the filter is `:hot`, and displayed in the UI through a connector.


xfalcox (approved):

lib/summarization/strategies/hot_topic_gists.rb
```diff
@@ -13,20 +13,21 @@ def targets_data
 
           op_post_number = 1
 
-          last_twenty_posts =
+          hot_topics_recent_cutoff = Time.zone.now - SiteSetting.hot_topics_recent_days.days
+
+          recent_hot_posts =
```
xfalcox, line 18: In some corner cases this may return nothing. I think in this case we should pick the last X, like it was done before, but now only when this set is empty.

lib/summarization/strategies/hot_topic_gists.rb
```diff
@@ -13,20 +13,21 @@ def targets_data
...
             Post
               .where(topic_id: target.id)
               .where("post_type = ?", Post.types[:regular])
               .where("NOT hidden")
-              .order("post_number DESC")
-              .limit(20)
+              .where("created_at >= ?", hot_topics_recent_cutoff)
               .pluck(:post_number)
```
xfalcox, line 25:
```suggestion

          # It may happen that a topic is hot without any recent posts
          # In that case, we'll just grab the last 20 posts
          # for an useful summary of the current state of the topic
          if recent_hot_posts.empty?
            recent_hot_posts =
              Post
                .where(topic_id: target.id)
                .where("post_type = ?", Post.types[:regular])
                .where("NOT hidden")
                .order("post_number DESC")
                .limit(20)
                .pluck(:post_number)
          end
```

PR #838, opened 2024-10-15
FIX: Remove missed AiPersona.allowed_chat

martin-brennan:
Followup bdf3b6268b2843f4b2ba8f8875e4d18a952ef326

I think this is the fix but not sure how to test it,
this is breaking the build


SamSaffron:
oh yikes,  not sure how I missed this.

SamSaffron (approved):

PR #842, opened 2024-10-18
FEATURE: Tool name validation

nvh0412:
Given that most LLM providers have their own function name formats (except for Ollama, as far as I have tested), they typically accept alphanumeric values (only including characters, numbers, and underscores). If we create a custom tool with the wrong name format, we won’t be able to invoke these tools. Other built-in tools we have won’t encounter this problem since their names are valid.

Ref: 

https://ai.google.dev/gemini-api/docs/function-calling#key-parameters-best-practices
https://meta.discourse.org/t/validate-the-name-field-for-ai-tools/330830

In this PR, we introduce the tool name validation in:

- The ActiveRecord class - the validation would be kicked in if we call the API directly.
- Use the tool_name to build the function payload to send to LLM instead of using the name.
- Add a validation on the AiPersona model which helps us prevent duplicate added tools to a Persona
- Add the unique constraint to the name field of the AiTool model.


app/services/discourse_ai/tool_name_checker.rb
```diff
@@ -0,0 +1,27 @@
...
+      if @tool_name.match? AiTool::ALPHANUMERIC_PATTERN
+        check_name_availability
+      else
+        { available: false, errors: [I18n.t("discourse_ai.tools.name.characters")] }
+      end
+    end
+
+    private
+
+    def check_name_availability
```
xfalcox, line 19: Should we also add a unique index to ensure uniqueness ?

nvh0412: Thanks, I'll double-check it later, but we did have that unique option on that column in the first place https://github.com/discourse/discourse-ai/blob/main/db/migrate/20240618080148_create_ai_tools.rb#L6, it's highly likely that we used the wrong option syntax for it, it should be

```ruby
t.string :name, index: { unique: true }
```

SamSaffron:
I think we may need a bit more of a refactor here, technically you may want 2 tools called "search" but then at the persona level pick the particular search you are after. 

So maybe we want an extra field here (tool_name tool name presented to LLM): 

```
name: search elastic docs 
tool_name: search
```

Then the "persona" save should ensure that the subset of tools you picked have distinct names. This also fixes "built-in" and "byo" clashes.


--- 

I am happy though with being more strict about what letters we allow for "tool_name" cause it will generally be safer to keep to a stricter amount of rules here.

config/locales/client.en.yml
```diff
@@ -211,9 +211,10 @@ en:
       tools:
         back: "Back"
         short_title: "Tools"
-        new: "New Tool"
         name: "Name"
         name_help: "The unique name of the tool as used by the language model"
+        tool_name: "Tool Name"
+        tool_name_help: "Tool name, only include numbers, letters, and underscores"
```
nvh0412, line 217: Let me know if you think there is a better copy for this tool_name_help and also the `name_help` copy as its value isn't unique actually.

app/models/ai_tool.rb
```diff
@@ -2,6 +2,7 @@
 
 class AiTool < ActiveRecord::Base
   validates :name, presence: true, length: { maximum: 100 }
+  validates :tool_name, presence: true, length: { maximum: 100 }, uniqueness: true
```
SamSaffron, line 5: tool_name should not be unique... it is a constraint on "persona" not on tool, the same tool name can exist multiple times with no problem

app/models/ai_tool.rb
```diff
@@ -202,7 +222,8 @@ def self.presets
       },
       {
         preset_id: "stock_quote",
-        name: "stock_quote",
+        name: "Stock Quote",
```
SamSaffron, line 225: Stock Quote (AlphaVantage) - feels like a better name.

SamSaffron:
its getting there, but we need to be clearer on what "name" means and what "tool_name" means. 

Name will show up in the Discourse UI and is the short identifier you will use to find the tool in various settings, it should be distinct (it is required) 

Tool Name is presented to the large language model. It is not distinct, but it is distinct per persona.  (persona validates on save)


I would say we don't allow nulls in either columns and during the migration just copy over old Name into Tool Name.

We probably want a distinct constraint on Name which means we need to append (1,2,3 etc) onto name during the migration if there are dupes.

db/migrate/20241020010245_add_tool_name_to_ai_tools.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class AddToolNameToAiTools < ActiveRecord::Migration[7.1]
+  def up
+    add_column :ai_tools, :tool_name, :string, null: false, default: "", if_not_exists: true
+
+    # Migrate existing name to tool_name
+    execute <<~SQL
+      UPDATE ai_tools
+      SET tool_name = regexp_replace(LOWER(name),'[^a-z0-9_]','', 'g');
+    SQL
```
nvh0412, line 11: re: We probably want a distinct constraint on Name which means we need to append (1,2,3 etc) onto name during the migration if there are dupes.

I reckon we should call it off and have another PR if we actually need it. Here, I only added the migration to migrate existing name to the new tool_name column. Let me know if you're ok with it.

SamSaffron:
I am not sure we need any of the "check_name" stuff though? we will just pick up issues on save. ''

the big thing missing though is the validation on save for ai_persona, that is where we want to make sure we did not add any duplicate tool names.

nvh0412:
I’m on board with you regarding the `check_name` issue, it’s now redundant if there’s no uniqueness validation on the tool_name. I’ll remove it shortly.

The purpose of this PR is to ensure we don’t have any invalid function names when invoking them via the LLM provider. As for the uniqueness tool validation on save for ai_persona, that feels like a separate matter to me.

But challenge accepted! I’ll take care of it—it’s actually quite fun.

lib/ai_bot/personas/general.rb
```diff
@@ -10,7 +10,6 @@ def tools
             Tools::Google,
             Tools::Image,
             Tools::Read,
-            Tools::Image,
```
nvh0412, line 13: gotcha moment.

nvh0412:
Hello @SamSaffron, would it be possible for you to review my latest changes in this PR once again? This would ensure that we're on the same page about what we've been discussing above.

app/models/ai_tool.rb
```diff
@@ -251,7 +272,8 @@ def self.presets
       },
       {
         preset_id: "image_generation",
-        name: "image_generation",
+        name: "Image Generation",
```
SamSaffron, line 275: Image Generation (Flux)

SamSaffron:
Its looking great to me... it is getting a bit late on a Friday, will merge it first thing next week.

SamSaffron:
was about to merge this but the validation is not looking right to me. 

The "name" stored in Persona should be: 

`custom:123` - for custom tools
`Search` - for buit in tools

The duplication check is against tool_name. 


So in the case of something like `['custom:1', 'custom:2']` we would load up the custom tool and check tool_name... if both of them are tool_name "search" then ... trouble... this will not work. Similarly ... don't allow a custom tool called "search" with built in "search". 

The UI should show Name everywhere.

We should have a test to cover this so it is clear how the validation works.

nvh0412:
Could we adjust `inner_name` (the first element in the `tool` array of the persona) of a Custom tool to include `tool_name`? This way, the validation logic can parse it and handle its validation logic. At the `ai_persona` level, `inner_name` seems to be the only reference point for meeting this requirement, lmk if you think there's a better place.

SamSaffron:
we can migrate the information on tools so it is clearer, but its a very painful refactor:

Eg: 

```
[{
   type: "custom",
   id: 123,
   force_tool: true,
   params: { ... ... }
}]
```

Storing name though is a no-go cause we would too easily drift (when you rename in the tools UI we would be stuck amending internal cached refs in persona.) 


I think the only easy way of validating easily is loading up the information from custom tool in a batch during validate.

nvh0412:
Even if we refactor the custom tool data structure saved in the Persona, as in your example, we would still need to load the custom tool data to retrieve their `tool_name`, wouldn’t we?

You're right that adjusting the inner_name to include tool_name is a no-go. I’ll load the custom tool data in batches and handle the validation that way. Does that work for you? However, I’m also open to a more robust solution in this PR since there’s no rush to merge it.

SamSaffron:
Yes sounds good!

nvh0412:
Screenshot 2024-10-30 at 6 36 04 PM

I think it would improve the user experience if we add the tool name to the selection box. This way, users can easily identify any duplicates.​

nvh0412:
In most Asian countries, we usually tidy up our homes to welcome the New Year. So here am I. Happy new year @SamSaffron

db/migrate/20241020010245_add_tool_name_to_ai_tools.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddToolNameToAiTools < ActiveRecord::Migration[7.1]
+  def up
+    add_column :ai_tools, :tool_name, :string, null: false, default: "", if_not_exists: true
```
SamSaffron, line 5: last one and I will merge it, this need a max length, sorry it is taking me so long here.

SamSaffron:
appreciate it @nvh0412 sorry it is taking me so long on this one

SamSaffron (approved):

SamSaffron (approved):

SamSaffron:
thanks. sorry for the delays here.

PR #847, opened 2024-10-21
DEV: `IMAGE_MARKDOWN_REGEX` import available in core

keegangeorge:
As of https://github.com/discourse/discourse/commit/ea1473e53260e8a458f8118b89e7b07b5cc50507, the regex for detecting image markdown is now available as an exported const in `discourse/lib/uploads`. We don't need to recreate it here as in our own utility file, instead we can simply import from core.

```js
import { IMAGE_MARKDOWN_REGEX } from "discourse/lib/uploads";
```

This PR removes the declaration in the plugin utilities file for `IMAGE_MARKDOWN_REGEX` and updates instances in the codebase to import from the core utility.


pmusaraj (approved):
This needs a compatibility file though, because it'll break if plugin is used with core earlier than the commit that introduces the import.

pmusaraj:
Yeah, that mostly covers this, but technically you should update that line to point to a commit in core from today, one that has the new import.

pmusaraj:
Yes, you're right, it's the plugin commit hash, so that does make sense for this. Thanks for fixing the typo though @keegangeorge.

PR #853, opened 2024-10-23
FEATURE: better logging for automation reports

SamSaffron:
A new feature_context json column was added to ai_api_audit_logs

This allows us to store rich json like context on any LLM request
made.

This new field now stores automation id and name.

Additionally allows llm_triage to specify maximum number of tokens

This means that you can limit the cost of llm triage by scanning only
first N tokens of a post.

### DIFF
--- a/lib/automation/report_runner.rb
+++ b/lib/automation/report_runner.rb
@@ -158,7 +158,7 @@ Follow the provided writing composition instructions carefully and precisely ste
           feature_context: {
             automation_id: @automation&.id,
             automation_name: @automation&.name,
-          }
+          },
         ) do |response|
           print response if Rails.env.development? && @debug_mode
           result << response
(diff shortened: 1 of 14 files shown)


lib/automation/report_runner.rb
```diff
@@ -153,6 +155,10 @@ def run!
           top_p: @top_p,
           user: Discourse.system_user,
           feature_name: "ai_report",
+          feature_context: {
+            automation_id: @automation&.id,
+            automation_name: @automation&.name,
+          },
```
lis2, line 161: Minor but maybe we `feature_context` should be empty hash and only merge ` automation_id` and `automation_name` when `@automation` exists?

lis2 (approved):

lib/automation/report_runner.rb
SamSaffron: I am so on the fence here ... automation being nil is kind of a usage bug so the odds of this happening is just tiny.

PR #860, opened 2024-10-23
FEATURE: Fast-track gist regeneration when a hot topic gets a new post

romanrizzi:


app/jobs/regular/update_hot_topic_gist.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      return if !TopicHotScore.where(topic: topic).exists?
+
+      summarizer = DiscourseAi::Summarization.topic_gist(topic)
+      gist = summarizer.existing_summary
+      return if gist.blank?
+      return if !gist.outdated
+
+      summarizer.delete_cached_summaries!
+      summarizer.summarize(Discourse.system_user)
```
xfalcox, line 23: We are using this in two places now, should we add a "re-sumarize" method that does this?

xfalcox:
From what I understood here, this will trigger on every new post and then update the gist if the topic has a hot score even if it was years ago?

Maybe we can check for a recent hot score on the topic *before* enqueuing the job?

romanrizzi:
@xfalcox - Good point, but I think what we want is a min score? We have a gravity setting that bring scores down over time

xfalcox:
> @xfalcox - Good point, but I think what we want is a min score? We have a gravity setting that bring scores down over time

Since this is the fast track and anything we miss will be auto-healed by the other job, IMO we can be pretty selective and only fast-track hot-scores updated recently. I'd go with updated in the last day to start.

romanrizzi:
LGTM 👍

PR #876, opened 2024-10-29
FEATURE: new endpoint for directly accessing a persona

SamSaffron:
The new `/admin/plugins/discourse-ai/ai-personas/stream-reply.json` was added.

This endpoint streams data direct from a persona and can be used
to access a persona from remote systems leaving a paper trail in
PMs about the conversation that happened

This endpoint is only accessible to admins.

### DIFF
--- a/app/controllers/discourse_ai/admin/ai_personas_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_personas_controller.rb
@@ -74,7 +74,191 @@ module DiscourseAi
         end
       end
 
+      class << self
+        POOL_SIZE = 10
+        def thread_pool
+          @thread_pool ||=
+            Concurrent::CachedThreadPool.new(min_threads: 0, max_threads: POOL_SIZE, idletime: 30)
+        end
+
+        def queue_reply(&block)
+          # think about a better way to handle cross thread connections
+          if Rails.env.test?
+            block.call
+            return
+          end
+
+          db = RailsMultisite::ConnectionManagement.current_db
+          thread_pool.post do
+            RailsMultisite::ConnectionManagement.with_connection(db) { block.call }
+          end
+        end
+      end
+
+      CRLF = "\r\n"
+
+      def stream_reply
+        persona =
+          AiPersona.find_by(name: params[:persona_name]) ||
+            AiPersona.find_by(id: params[:persona_id])
+        return render_json_error(I18n.t("discourse_ai.errors.persona_not_found")) if persona.nil?
+
+        return render_json_error(I18n.t("discourse_ai.errors.persona_disabled")) if !persona.enabled
+
+        if !persona.default_llm.present?
+          return render_json_error(I18n.t("discourse_ai.errors.no_default_llm"))
+        end
+
+        if params[:query].blank?
+          return render_json_error(I18n.t("discourse_ai.errors.no_query_specified"))
+        end
+
+        if !persona.user_id
+          return render_json_error(I18n.t("discourse_ai.errors.no_user_for_persona"))
+        end
+
+        if !params[:username] && !params[:user_unique_id]
+          return render_json_error(I18n.t("discourse_ai.errors.no_user_specified"))
+        end
+
+        user = nil
+
+        if params[:username]
+          user = User.find_by(username_lower: params[:username].downcase)
+          return render_json_error(I18n.t("discourse_ai.errors.user_not_found")) if user.nil?
+        elsif params[:user_unique_id]
+          user = stage_user
+        end
+
+        raise Discourse::NotFound if user.nil?
+
+        topic_id = params[:topic_id].to_i
+        topic = nil
+        post = nil
+
+        if topic_id > 0
+          topic = Topic.find(topic_id)
+
+          raise Discourse::NotFound if topic.nil?
+
+          if topic.topic_allowed_users.where(user_id: user.id).empty?
+            return render_json_error(I18n.t("discourse_ai.errors.user_not_allowed"))
+          end
+
+          post =
+            PostCreator.create!(
+              user,
+              topic_id: topic_id,
+              raw: params[:query],
+              skip_validations: true,
+            )
+        else
+          post =
+            PostCreator.create!(
+              user,
+              title: I18n.t("discourse_ai.ai_bot.default_pm_prefix"),
+              raw: params[:question],
+              archetype: Archetype.private_message,
+              target_usernames: "#{user.username},#{persona.user.username}",
+              skip_validations: true,
+            )
+
+          topic = post.topic
+        end
+
+        hijack = request.env["rack.hijack"]
+        io = hijack.call
+
+        user = current_user
+
+        self.class.queue_reply do
+          begin
+            io.write "HTTP/1.1 200 OK"
+            io.write CRLF
+            io.write "Content-Type: application/json"
+            io.write CRLF
+            io.write "Transfer-Encoding: chunked"
+            io.write CRLF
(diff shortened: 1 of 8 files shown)


SamSaffron:
Example usage:

```
curl -X POST "http://l.discourse:9292/admin/plugins/discourse-ai/ai-personas/stream-reply.json" \
     -H "Content-Type: application/json" \
     -H "Api-Key: SOME_KEY" \
     -H "Api-Username: sam" \
     -d '{
       "persona_name": "DJ Random",
       "query": "how are you?",
       "user_unique_id": "site:test.com:user_id:1",
       "preferred_username": "test_user"
     }' \

{"topic_id":290251,"bot_user_id":-1200,"persona_id":4}

{"partial":"Well"}

{"partial":","}

{"partial":" that"}

{"partial":" was"}

{"partial":" unexpected"}

{"partial":"!"}

{"partial":" It"}

{"partial":" seems"}

{"partial":" even"}

{"partial":" I"}

{"partial":" can"}

{"partial":" be"}

{"partial":" a"}

{"partial":" bit"}

{"partial":" too"}

{"partial":" random"}

{"partial":" sometimes"}

{"partial":"."}

{"partial":" Let's"}

{"partial":" try"}

{"partial":" that"}

{"partial":" again"}

{"partial":":"}

{"partial":" \n\n"}

{"partial":"I'm"}

{"partial":" feeling"}

{"partial":" as"}

{"partial":" random"}

{"partial":" as"}

{"partial":" ever"}

{"partial":","}

{"partial":" ready"}

{"partial":" to"}

{"partial":" pick"}

{"partial":" choices"}

{"partial":" with"}

{"partial":" style"}

{"partial":"!"}

{"partial":" Just"}

{"partial":" let"}

{"partial":" me"}

{"partial":" know"}

{"partial":" what"}

{"partial":" you"}

{"partial":" need"}

{"partial":"."}

{"partial":" 😊"}
```

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,200 @@ def destroy
...
+
+      def stream_reply
+        persona =
+          AiPersona.find_by(name: params[:persona_name]) ||
+            AiPersona.find_by(id: params[:persona_id])
+        return render_json_error(I18n.t("discourse_ai.errors.persona_not_found")) if persona.nil?
+
+        return render_json_error(I18n.t("discourse_ai.errors.persona_disabled")) if !persona.enabled
+
+        if !persona.default_llm.present?
```
Grubba27, line 112:
```suggestion
        if persona.default_llm.blank?
```

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,200 @@ def destroy
...
+            rescue StandardError => e
+              Discourse.warn_exception(e, message: "Discourse AI: Unable to stream reply")
+            end
+          end
+        end
+      end
+
+      CRLF = "\r\n"
+
+      def stream_reply
```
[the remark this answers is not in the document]
SamSaffron: will move stuff around, sure.

config/locales/server.en.yml
```diff
@@ -347,3 +348,13 @@ en:
...
+
+    errors:
+      no_query_specified: The query parameter is required, please specify it.
+      no_user_for_persona: The persona specified does not have a user associated with it.
+      persona_not_found: The persona specified does not exist. Check the persona_name or persona_id params.
+      no_user_specified: The username or the user_unique_id parameter is required, please specify it.
+      user_not_found: The user specified does not exist. Check the username param.
+      persona_disabled: The persona specified is disabled. Check the persona_name or persona_id params.
+      no_default_llm: The persona must have a default_llm defined.
+      user_not_allowed: The user is not allowed to participate in the topic.
```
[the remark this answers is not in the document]
SamSaffron: yeah one of those is not even used. moved the other one

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,205 @@ def destroy
...
```
martin-brennan, line 104: I see Keegan mentioned a similar thing, but this is an extremely fat controller action, I really don't feel like we should be doing this very often, if ever. It doesn't feel very maintainable or easy to read, and could easily expand even further in size. Much of this is a perfect case for using a service.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,205 @@ def destroy
...
+            # multi-threaded code that exhibits various limitations in rails
+            p e if Rails.env.development?
+            Discourse.warn_exception(e, message: "Discourse AI: Unable to stream reply")
+          ensure
+            io.close
+          end
+        end
+      end
+
+      def stage_user
```
martin-brennan, line 253: Another thing that could be in a service, or at least a reusable class in core. Nothing except the custom field is AI-specific here

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,205 @@ def destroy
...
       private
 
+      AI_STREAM_CONVERSATION_UNIQUE_ID = "ai-stream-conversation-unique-id"
+
+      # keeping this in a static method so we don't capture ENV and other bits
+      # this allows us to release memory earlier
+      def self.queue_streamed_reply(io, persona, user, topic, post)
+        schedule_block do
+          begin
+            io.write "HTTP/1.1 200 OK"
```
martin-brennan, line 190: I guess it's for a good reason (performance?), but is there any way not to do this writing of headers and data completely manually?

app/controllers/discourse_ai/admin/ai_personas_controller.rb
```diff
@@ -74,8 +74,205 @@ def destroy
         end
       end
 
+      class << self
```
martin-brennan, line 77: Feels like stuff that should be generic/in core/an `ActiveSupport::Concern` that we could use elsewhere too?

app/controllers/discourse_ai/admin/ai_personas_controller.rb
martin-brennan: The reasons why it's perfect for a service:

* It has model lookups
* It has several parameter validations and error messages
* It has a clear action and result from that action, which could also have more errors

Not saying the reply streaming stuff should be in one, but everything before it easily could be

martin-brennan (commented):
Not sure if further refactors are planned here. I get that we are trying to move fast, but I think there is time to go back and improve things as well.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: rails has streaming, but I don't want to stream from the foreground thread, I can make wrapper classes for this but at the end the same code will run

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: I will move it out to a service at some point definitely do not want a concern here , this is a very sharp knife I only want people that absolutely understand repercussions to use it.

app/controllers/discourse_ai/admin/ai_personas_controller.rb
SamSaffron: I have been waiting a bit on service docs and service stability, stuff has been changing recently

PR #889, opened 2024-11-03
FIX: encode parameters returned from LLMs correctly 

SamSaffron:
Fixes encoding of params on LLM function calls. 

Previously we would improperly return results if a function parameter returned an HTML tag.

Additionally adds some missing HTTP verbs to tool calls.


jjaffeux (approved):
lgtm

PR #891, opened 2024-11-04
FIX: we must properly encode objects prior to escaping

SamSaffron:
(in cases of arrays escapeHTML will not work)


martin-brennan (approved):

SamSaffron:
will follow up with a test later, just need to land this.

martin-brennan:
Approving now since its urgent, tests can be added later :+1:

PR #892, opened 2024-11-04
FEATURE: Automatically backfill regular summaries.

romanrizzi:
This change introduces a job to summarize topics and cache the results automatically. We provide a setting to control how many topics we'll backfill per hour and what the topic's minimum word count is to qualify.

We'll prioritize topics without summary over outdated ones.


app/jobs/scheduled/summaries_backfill.rb
```diff
@@ -0,0 +1,42 @@
...
+  class SummariesBackfill < ::Jobs::Scheduled
+    every 5.minutes
+
+    def execute(_args)
+      return if !SiteSetting.discourse_ai_enabled
+      return if !SiteSetting.ai_summarization_enabled
+      return if SiteSetting.ai_summary_backfill_maximum_topics_per_hour.zero?
+
+      # Backfill up to 150 summaries per 5-minute run.
+      limit_per_job = 150
```
xfalcox, line 13: Shouldn't this be `ai_summary_backfill_maximum_topics_per_hour / 12` ?

romanrizzi: Isn't that worse when max topics per hour is small?

xfalcox: Should be the same right? every 5 minutes is 12 calls an hour.

xfalcox (approved):

PR #896, opened 2024-11-05
FEATURE: Extend summary backfill to also generate gists

romanrizzi:


xfalcox:
This will only generate gists for topics who need summaries won't it? A topic with a summary won't get gists this way.

romanrizzi:
> This will only generate gists for topics who need summaries won't it? A topic with a summary won't get gists this way.

Actually, I think to revisit this logic. The join will duplicate a topic for each type of summary.

xfalcox (approved):

PR #906, opened 2024-11-11
DEV: Move semantic search out of connector to component

markvanlan:
This PR allows the semantic search component to be rendered by another plugin, rather than being defined in the connector, which limits it's usage.


SamSaffron:
@markvanlan where are we on this?

PR #913, opened 2024-11-13
FEATURE: Order by emotion on /filter

xfalcox:


lib/sentiment/emotion_filter_order.rb
```diff
@@ -0,0 +1,75 @@
...
+          admiration
+          optimism
+          fear
+          love
+          excitement
+          curiosity
+          amusement
+          surprise
+          gratitude
+          pride
```
pmusaraj, line 35: Sort these lines alphabetically?

spec/lib/modules/sentiment/emotion_filter_order_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+      gratitude
+      pride
+    ]
+
+    filters = DiscoursePluginRegistry.custom_filter_mappings.reduce(Hash.new, :merge)
+
+    emotions.each { |emotion| expect(filters).to include("order:emotion_#{emotion}") }
+  end
+
+  it "filters topics by emotion" do
```
pmusaraj, line 130: Hmm, I am confused a bit about ordering vs filtering here. The admin report for a given emotion should point to a filtered list of topics where there is a match for that emotion, right? For that, would we need a filter for sometihng like `emotion:disgust`? 

We certainly want `order:emotion_disgust`, no doubt, but I think we also need the filtering?

pmusaraj: Yeah, not a fan of the double keywords. Not a fan of `order:X` filtering out results either because it goes against what order does in other places. I don't have a definitive answer, maybe in the short-run we go with the `order:` keyword only and accept it is exceptional. 

Regarding the cutoff, I'm happy with your suggestion. The main thing is that we need to use the same threshold that we use in the admin report: 

image

The graph here will likely change to accomodate the additional emotion dimensions, but, if it will still show a count per emotion, we need to respect that same count in the filtered view.

pmusaraj (approved):
Let's try it!

PR #918, opened 2024-11-18
fixed "fools" to "tools" typo in client.en.yml

tobiaseigen:


SamSaffron:
thanks, fixed this the other day should be good now!

PR #919, opened 2024-11-19
FEATURE: Add support for Mistral models

SamSaffron:
Adds support for mistral models (pixtral and mistral large now have presets) 

Also corrects token accounting in AWS bedrock models

### DIFF
--- a/lib/completions/endpoints/mistral.rb
+++ b/lib/completions/endpoints/mistral.rb
@@ -7,6 +7,10 @@ module DiscourseAi
         def self.can_contact?(model_provider)
           model_provider == "mistral"
         end
+
+        def provider_id
+          AiApiAuditLog::Provider::Mistral
+        end
       end
     end
   end
--- a/lib/completions/endpoints/mistral.rb
+++ b/lib/completions/endpoints/mistral.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module Completions
+    module Endpoints
+      class Mistral < OpenAi
+        def self.can_contact?(model_provider)
+          model_provider == "mistral"
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


lib/completions/endpoints/mistral.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module Completions
+    module Endpoints
+      class Mistral < OpenAi
```
tyb-talks, line 6: Should the `provider_id` be overridden? As-is, this currently returns the one for OpenAi.

tyb-talks (approved):
Minor question on provider_id for the endpoint, but otherwise looks ok.

lib/completions/endpoints/mistral.rb
SamSaffron: oh sure ... good catch will fix

tyb-talks (changes requested):
I just spotted something I missed earlier in the OpenAi endpoint:

```
        def prepare_payload(prompt, model_params, dialect)
          payload = default_options.merge(model_params).merge(messages: prompt)

          if @streaming_mode
            payload[:stream] = true

            # Usage is not available in Azure yet.
            # We'll fallback to guess this using the tokenizer.
            payload[:stream_options] = { include_usage: true } if llm_model.provider == "open_ai"
          end
```

tyb-talks (approved):
😅 actually it's fine, it just omits those params if it's not from OpenAi. 🤔 If the logic in the inherited endpoint changes more specifically for OpenAi though, it would also affect this Mistral endpoint. Still fine for now.

SamSaffron:
Yeah this is safe by "side effect" cause mistral just sends this info anyway.

SamSaffron:
thanks for the review @tyb-talks :hugs:

PR #932, opened 2024-11-20
DEV: improve artifact presentation

SamSaffron:
1. Keep source in a "details" block after rendered so it does
not overwhelm users

2. Ensure artifacts are never indexed by robots

3. Cache break our CSS that changed recently


SamSaffron:
will add a test and merge, thx

PR #937, opened 2024-11-20
REFACTOR: Tidy-up embedding endpoints config.

romanrizzi:
Two changes worth mentioning:

1. `.instance` returns a fully configured embedding endpoint ready to use. 
2. All endpoints respond to the same method and have the same signature - `perform!(text)`. It's no longer a class method.

This makes it easier to reuse them when generating embeddings in bulk.

### DIFF
--- a/lib/embeddings/vector_representations/all_mpnet_base_v2.rb
+++ b/lib/embeddings/vector_representations/all_mpnet_base_v2.rb
@@ -24,12 +24,7 @@ module DiscourseAi
         end
 
         def vector_from(text, asymetric: false)
-          DiscourseAi::Inference::DiscourseClassifier.perform!(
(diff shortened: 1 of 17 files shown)


xfalcox:
Why the divide between `HuggingFaceText` and `TeiEmbeddings` ? They are the same service, this separation makes it hella confusing.

romanrizzi:
> Why the divide between `HuggingFaceText` and `TeiEmbeddings` ? They are the same service, this separation makes it hella confusing.
> 
> Also `TeiEmbeddings` is another case of "Text **Embeddings** Inference **Embeddings**"

Fair enough. I'll merge it back but leave `rerank` and `classify` as class methods.

PR #938, opened 2024-11-21
FIX: cancel functionality regressed

SamSaffron:
The cancel messaging was not floating correctly to the HTTP call


martin-brennan (approved):
Don't really understand it but approving since it's a big bug :+1:

SamSaffron:
I am trying to write a test, if I can one done with it now I will merge ... it is just very hard to test.

PR #941, opened 2024-11-21
DEV: Backfill embeddings concurrently.

romanrizzi:
We are adding a new method for generating and storing embeddings in bulk, which relies on `Concurrent::Promises::Future`. Generating an embedding consists of three steps:

- Prepare text
- HTTP call to retrieve the vector
- Save to DB.

Second one is independently executed on whatever thread the pool gives us.

We are bringing a custom thread pool instead of the global executor since we want control over how many threads we spawn to limit concurrency. We also avoid firing thousands of HTTP requests when working with large batches.

Ran benchmarks locally with the following script:

```ruby
def benchmark_embeddings
  ActiveRecord::Base.logger.level = 1 # or Logger::INFO
  truncation = DiscourseAi::Embeddings::Strategies::Truncation.new
  vector_rep = DiscourseAi::Embeddings::VectorRepresentations::Base.current_representation(truncation)

  Benchmark.bmbm do |x|
    x.report("With concurrent (90 topics)") do 
      DB.exec("DELETE FROM ai_topic_embeddings")
      vector_rep.gen_bulk_reprensentations(Topic.includes(:tags, :posts).all)
    end

    x.report("Without concurrent (90 topics)") do
      DB.exec("DELETE FROM ai_topic_embeddings")
      Topic.includes(:tags, :posts).all.each { |t| vector_rep.generate_representation_from(t) }
    end
  end
end
```

Results:

```
                                     user     system      total        real
With concurrent (90 topics)      2.418655   0.084879   2.503534 (  3.716947)
Without concurrent (90 topics)   5.097356   0.268728   5.366084 ( 63.132123)
```

### DIFF
--- a/lib/embeddings/vector_representations/base.rb
+++ b/lib/embeddings/vector_representations/base.rb
@@ -50,8 +50,38 @@ module DiscourseAi
           raise NotImplementedError
         end
 
+        def gen_bulk_reprensentations(relation)
+          http_pool_size = 100
+          pool =
+            Concurrent::CachedThreadPool.new(
+              min_threads: 0,
+              max_threads: http_pool_size,
+              idletime: 30,
+            )
+
+          embedding_gen = inference_client
+          promised_embeddings =
+            relation.map do |record|
+              materials = { target: record, text: prepare_text(record) }
+
+              Concurrent::Promises
+                .fulfilled_future(materials, pool)
+                .then_on(pool) do |w_prepared_text|
+                  w_prepared_text.merge(
+                    embedding: embedding_gen.perform!(w_prepared_text[:text]),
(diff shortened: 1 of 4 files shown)


lib/embeddings/vector_representations/base.rb
```diff
@@ -50,8 +50,37 @@ def vector_from(text, asymetric: false)
           raise NotImplementedError
         end
 
+        def gen_bulk_reprensentations(relation)
+          raw_inputs = relation.map { |record| { target: record, text: prepare_text(record) } }
+
+          pool_size = 10
```
romanrizzi, line 56: Set this to `10` because I feared we would grab too many DB connections.

xfalcox (approved):

PR #942, opened 2024-11-21
FIX: automatically bust cache for share ai assets

SamSaffron:
CDNs can be configured to strip query params in Discourse
hosting. This is generally safe, but in this case we had
no way of busting the cache using the path.

New design properly caches and properly breaks busts the
cache if asset changes so we don't need to worry about versions


davidtaylorhq (approved):

martin-brennan (approved):
LGTM

PR #954, opened 2024-11-25
FIX: Related topics missing after scrolling post stream

keegangeorge:
Reported on Meta: https://meta.discourse.org/t/related-topics-missing-after-rereading-a-topic-with-more-than-20-posts/306147/6


xfalcox:
Continues in #976

PR #956, opened 2024-11-26
UX: Applying more admin UI guidelines

martin-brennan:
This commit applies further admin UI guidelines, now that they have been more
fleshed out in core, to the AI admin UI:

* Tools
* LLMs
* Personas

The changes include but are not limited to:

* Applying the table CSS classes, for desktop and mobile
* Adding a description and learn more link for each tab
* Adding an empty list placeholder with CTA using `AdminConfigAreaEmptyList`
* Replacing custom headings with `AdminPageSubheader`

**Desktop**

image
image
image

**Mobile**

image
image
image


spec/system/llms/ai_llm_spec.rb
```diff
@@ -76,14 +76,14 @@
     it "shows the provider as CDCK in the UI" do
       visit "/admin/plugins/discourse-ai/ai-llms"
       expect(page).to have_css(
-        "[data-llm-id='cdck-hosted'] .column-provider",
```
tyb-talks, line 79: Noticed that we are treating the table wrapper classes as extraneous (e.g. `admin-row__detail`, `d-admin-row__controls`) during the assertions. Is this to ensure the tests aren't too fragile? Assertions on those would enforce that these elements keep to classes that align with the admin UI guidelines.

tyb-talks (approved):
Generally lgtm, had a question on the tr/td class assertions.

PR #958, opened 2024-11-26
FEATURE: Calculate gists from non hot topics too

xfalcox:
We want to fast track non hot topics gists too, as gists are also available on /latest now.


lib/summarization/entry_point.rb
```diff
@@ -35,20 +35,12 @@ def inject_into(plugin)
...
         # jobs if the feature is disabled.
         plugin.on(:post_created) do |post|
           if SiteSetting.discourse_ai_enabled && SiteSetting.ai_summarization_enabled &&
-               SiteSetting.ai_summarize_max_hot_topics_gists_per_batch > 0 && post.topic
-            hot_score = TopicHotScore.find_by(topic: post.topic)
-
-            if hot_score.present? && hot_score.updated_at > 1.day.ago
-              Jobs.enqueue(:update_hot_topic_gist, topic_id: post&.topic_id)
-            end
+               SiteSetting.ai_summary_gists_enabled > 0 && post.topic
```
romanrizzi, line 42:
```suggestion
               SiteSetting.ai_summary_gists_enabled && post.topic
```

config/locales/server.en.yml
```diff
@@ -86,7 +86,7 @@ en:
     ai_summarization_model: "Model to use for summarization."
     ai_custom_summarization_allowed_groups: "Groups allowed to use create new summaries."
     ai_pm_summarization_allowed_groups: "Groups allowed to create and view summaries in PMs."
-    ai_summarize_max_hot_topics_gists_per_batch: "After updating topics in the hot list, we'll generate brief summaries of the first N ones. (Disabled when 0)"
+    ai_summary_gists_enabled: "After updating topics in the hot list, we'll generate brief summaries of the first N ones. (Disabled when 0)"
```
romanrizzi, line 89: We need to update the copy

romanrizzi (changes requested):
We should also skip gist generation on `Jobs::SummariesBackfill` when `ai_summary_gists_enabled` is false.

romanrizzi (approved):

PR #961, opened 2024-11-26
FIX: Shutdown embeddings thread pool after processing

romanrizzi:


lib/embeddings/vector_representations/base.rb
```diff
@@ -77,10 +77,12 @@ def gen_bulk_reprensentations(relation)
               end
               .compact
 
-          Concurrent::Promises
-            .zip(*promised_embeddings)
-            .value!
-            .each { |e| save_to_db(e[:target], e[:embedding], e[:digest]) }
+          promises = Concurrent::Promises.zip(*promised_embeddings).value!
+
+          pool.shutdown
```
xfalcox, line 82: I believe we could do all this after storing all in the DB ?

xfalcox (approved):

PR #967, opened 2024-11-27
DEV: Prevent `detect_text_locale` from appearing in menus

keegangeorge:
## :mag: Overview
With the recent changes to allow Discourse AI in the translator plugin (https://github.com/discourse/discourse-ai/pull/946), `detect_text_locale` was needed as a `CompletionPrompt`. However, it is leaking into composer/post helper menus. This PR ensures we don't not show it in those menus.

### Before
Screenshot 2024-11-27 at 14 21 29

### After
Screenshot 2024-11-27 at 14 21 38

### DIFF
--- a/lib/ai_helper/assistant.rb
+++ b/lib/ai_helper/assistant.rb
@@ -233,6 +233,8 @@ module DiscourseAi
           %w[post]
         when "illustrate_post"
           %w[composer]
+        when "detect_text_locale"
+          %w[]
         else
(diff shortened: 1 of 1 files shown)


xfalcox (approved):

SamSaffron:
I think it is fine, but we should have some sort of test for this stuff, even just on the server. 

Then as we add / change options we will not be in for a surprise. I think it is worth a followup PR.

PR #972, opened 2024-11-28
DEV: Remove the short-lived prompt for locale detection

nattsw:
It was originally introduced in https://github.com/discourse/discourse-ai/pull/946 for https://github.com/discourse/discourse-translator/pull/181. 

But since then, discourse-translator will be using its own custom prompt, so there is no need for this seeded prompt any more. This keeps things neater.


romanrizzi (approved):

SamSaffron:
test fails are legit @nattsw ... our tests expect a certain number of prompts

PR #975, opened 2024-11-28
UX: Fix styling of AI Helper on quick edit

xfalcox:


romanrizzi (approved):

pmusaraj (approved):
Approving as a quick fix, I suspect we'll need to standardize this at some point.

PR #982, opened 2024-11-29
FEATURE: Backfill posts sentiment.

romanrizzi:
It adds a scheduled job to backfill posts' sentiment, similar to our existing rake task, but with two settings to control the batch size and posts' max-age.


lib/sentiment/post_classification.rb
```diff
@@ -3,6 +3,33 @@
 module DiscourseAi
   module Sentiment
     class PostClassification
+      def self.backfill_query(from_post_id: nil, max_age_days: nil)
+        public_categories = Category.where(read_restricted: false).pluck(:id)
```
xfalcox, line 7: Should we drop limiting it to public categories?

xfalcox (approved):

PR #984, opened 2024-11-30
FEATURE: Add support for mxbai-embed-xsmall-v1 embedding model

cduk:
Add mxbai-embed-xsmall-v1 as an embedding model.


xfalcox:
As we talked in Meta, this needs migrations to address indexes and an unique ID.

It will be addressed by configurable embedding soon, so let's wait on that.

PR #1020, opened 2024-12-10
FEATURE: Add periodic problem checks for each LLM in use

keegangeorge:
### 🔍 Overview
This feature adds a periodic problem check which periodically checks for issues with LLMs that are in use. Periodically, we will run a test to see if the in use LLMs are still operational. If it is not, the LLM with the problem is surfaced to the admin so they can easily go and update the configuration.

### 📸 Screenshots
Dashboard Problem Check AI

### DIFF
--- a/app/services/problem_check/ai_llm_status.rb
+++ b/app/services/problem_check/ai_llm_status.rb
@@ -8,6 +8,10 @@ class ProblemCheck::AiLlmStatus < ProblemCheck
     llm_errors
   end
 
+  def base_path
+    Discourse.base_path
+  end
+
   private
 
   def llm_errors
@@ -24,7 +28,7 @@ class ProblemCheck::AiLlmStatus < ProblemCheck
     rescue => e
       error_message = parse_error_message(e.message)
       message =
-        "#{I18n.t("dashboard.problem.ai_llm_status", { model_name: model.display_name, model_id: model.id })}"
+        "#{I18n.t("dashboard.problem.ai_llm_status", { base_path: base_path, model_name: model.display_name, model_id: model.id })}"
 
       Problem.new(
         message,
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -455,4 +455,4 @@ en:
       prompt_message_length: The message %{idx} is over the 1000 character limit.
   dashboard:
     problem:
-      ai_llm_status: "The LLM model: %{model_name} is encountering issues. Please check the <a href='/admin/plugins/discourse-ai/ai-llms/%{model_id}'>model's configuration page</a>."
+      ai_llm_status: "The LLM model: %{model_name} is encountering issues. Please check the <a href='%{base_path}/admin/plugins/discourse-ai/ai-llms/%{model_id}'>model's configuration page</a>."
--- a/spec/services/problem_check/ai_llm_status_spec.rb
+++ b/spec/services/problem_check/ai_llm_status_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe ProblemCheck::AiLlmStatus do
       it "returns a problem with an LLM model" do
         stub_request(:post, post_url).to_return(status: 403, body: error_response, headers: {})
         message =
-          "#{I18n.t("dashboard.problem.ai_llm_status", { model_name: llm_model.display_name, model_id: llm_model.id })}"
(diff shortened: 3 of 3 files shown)


app/services/problem_check/ai_llm_status.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class ProblemCheck::AiLlmStatus < ProblemCheck
+  self.priority = "high"
+  self.perform_every = 1.hour
```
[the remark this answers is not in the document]
romanrizzi: I don't think it's very expensive, but maybe it's too much? 3/6 hours might be a better target.

config/locales/server.en.yml
```diff
@@ -453,3 +453,6 @@ en:
       no_default_llm: The persona must have a default_llm defined.
       user_not_allowed: The user is not allowed to participate in the topic.
       prompt_message_length: The message %{idx} is over the 1000 character limit.
+  dashboard:
+    problem:
+      ai_llm_status: "The LLM model: %{model_name} is encountering issues. Please check the <a href='/admin/plugins/discourse-ai/ai-llms/%{model_id}'>model's configuration</a>."
```
pmusaraj, line 458:
```suggestion
      ai_llm_status: "The LLM model: %{model_name} is encountering issues. Please check the <a href='/admin/plugins/discourse-ai/ai-llms/%{model_id}'>model's configuration page</a>."
```

app/services/problem_check/ai_llm_status.rb
```diff
@@ -0,0 +1,56 @@
...
+    return [] if !SiteSetting.discourse_ai_enabled
+
+    LlmModel.in_use.find_each.filter_map do |model|
+      try_validate(model) { validator.run_test(model) }
+    end
+  end
+
+  def try_validate(model, &blk)
+    begin
+      raise({ message: "Forced error for testing" }.to_json) if Rails.env.test?
```
pmusaraj, line 23: Hmm... this means you can't test the use case for "model is enabled and working fine" though.

pmusaraj: Does the dummy model have an auth mechanism? One way I've seen this tested in other places (login methods for example) is by mocking the author response, if it doesn't return a 200, then the problem check is raised.

pmusaraj: In other words, if the credentials are wrong, will the problem check be raised here?

romanrizzi: You will have to config a real model and stub an HTTP request to simulate the error case. Set one of the module's site setting to that model and run the test.

config/locales/server.en.yml
```diff
@@ -453,3 +453,6 @@ en:
       no_default_llm: The persona must have a default_llm defined.
       user_not_allowed: The user is not allowed to participate in the topic.
       prompt_message_length: The message %{idx} is over the 1000 character limit.
+  dashboard:
+    problem:
+      ai_llm_status: "The LLM model: %{model_name} is encountering issues. Please check the <a href='%{base_path}/admin/plugins/discourse-ai/ai-llms/%{model_id}'>model's configuration page</a>."
```
gschlager, line 458: BTW, it's actually recommended to build the URL in Ruby, and have a single %{url} parameter. If the URL ever changes, translators don't need to update their translations.

romanrizzi (approved):

config/locales/server.en.yml
davidtaylorhq: I wonder if it's the way we cache problem checks? I ran into it earlier in the week when working on encrypt: (internal `/t/128444/7`)

app/services/problem_check/ai_llm_status.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def try_validate(model, &blk)
+    begin
+      blk.call
+      nil
+    rescue => e
+      error_message = parse_error_message(e.message)
+      message =
+        "#{I18n.t("dashboard.problem.ai_llm_status", { base_path: base_path, model_name: model.display_name, model_id: model.id })}"
```
Drenmi, line 31: Sorry. Late to the party. This shouldn't really need to be wrapped in an interpolation string? 🙂

PR #1021, opened 2024-12-11
UX: No admin header for edit personas tools or llms

lis2:
In this PR, we added functionality to hide the admin header for edit/new actions - https://github.com/discourse/discourse/pull/30175

To make it work properly, we have to rename `show` to `edit` which is also a more accurate name.

Screenshot 2024-12-11 at 2 37 26 PM
Screenshot 2024-12-11 at 2 37 36 PM


assets/javascripts/discourse/admin-discourse-ai-plugin-route-map.js
```diff
@@ -6,17 +6,17 @@ export default {
   map() {
     this.route("discourse-ai-personas", { path: "ai-personas" }, function () {
       this.route("new");
-      this.route("show", { path: "/:id" });
+      this.route("edit", { path: "/:id" });
```
martin-brennan, line 9: Can you also make the paths `/:id/edit`? Also something Ted and I discussed in passing, I think it's probably good to standardize

martin-brennan (approved):
Nice!

PR #1049, opened 2025-01-06
FEATURE: configurable embeddings

romanrizzi:
Adds a way to configure embeddings similar to what we already have for other concepts like LLMs, tools, personas, etc.

It hides many old settings and adds a new one called "ai_embeddings_selected_model". We include a data migration to seed the model using these old settings.

It also removes the `DiscourseClassifier` service.

Screenshot 2025-01-13 at 11 33 39 AM

---

Screenshot 2025-01-13 at 11 33 45 AM

---

Screenshot 2025-01-13 at 11 33 51 AM


app/models/embedding_definition.rb
```diff
@@ -0,0 +1,139 @@
...
+
+  def strategy
+    @strategy ||= DiscourseAi::Embeddings::Strategies::Truncation.new
+  end
+
+  def cloudflare_client
+    DiscourseAi::Inference::CloudflareWorkersAi.new(endpoint_url, api_key)
+  end
+
+  def discourse_client
```
xfalcox, line 94: Let's take this opportunity and deprecate the discourse api

config/settings.yml
xfalcox, line 1: Time to kill / hide settings like

ai_openai_embeddings_url
ai_hugging_face_tei_endpoint
ai_hugging_face_tei_api_key
ai_cloudflare_workers_account_id
ai_cloudflare_workers_api_token
ai_gemini_api_key

romanrizzi: ah yeah, missed those. Will hide them

db/migrate/20241230153300_new_embeddings_tables.rb
```diff
@@ -0,0 +1,68 @@
...
+      t.bigint :topic_id, null: false
+      t.bigint :model_id, null: false
+      t.integer :model_version, null: false
+      t.integer :strategy_id, null: false
+      t.integer :strategy_version, null: false
+      t.text :digest, null: false
+      t.column :embeddings, "halfvec", null: false
+      t.timestamps
+
+      t.index %i[model_id strategy_id topic_id],
```
xfalcox, line 17: Where is the index for perf ? 

https://github.com/discourse/discourse-ai/blob/main/db/migrate/20241008054440_create_binary_indexes_for_embeddings.rb

db/migrate/20241230153300_new_embeddings_tables.rb
```diff
@@ -0,0 +1,68 @@
...
+      t.text :digest, null: false
+      t.column :embeddings, "halfvec", null: false
+      t.timestamps
+
+      t.index %i[model_id strategy_id rag_document_fragment_id],
+              unique: true,
+              name: "index_ai_fragments_embeddings_on_model_strategy_fragment"
+    end
+
+    # Copy data from old tables to new tables
```
xfalcox, line 52: Don't we need to redo this again in a post migrate migration for embeddings that were created mid deploy ?

db/migrate/20241230153300_new_embeddings_tables.rb
romanrizzi: I'm replicating stuff from the old table so no particular reason. I imagine is because of `unique: true`

db/migrate/20241230153300_new_embeddings_tables.rb
romanrizzi: I believe the batch job will take care of those. Will confirm.

xfalcox (approved):

PR #1053, opened 2025-01-08
FIX: Update spam controller action should consider seeded LLM properly

keegangeorge:
The seeded LLM setting: `SiteSetting.ai_spam_detection_model_allowed_seeded_models` returns a _string_ with IDs separated by pipes. running `_map` on it will return an array with strings. We were previously checking for the id with custom prefix identifier, but instead we should be checking the stringified ID.


SamSaffron:
lets add a quick test here please

xfalcox (approved):

PR #1063, opened 2025-01-14
FIX: Split backfill into separate migrations to use independent transactions

romanrizzi:


romanrizzi:
@davidtaylorhq - CI failed because the post migration that drops the table ran before the new ones. I modified the timestamps to just after the new tables creation. I think this won't be an issue because of how migrations are applied during deploy, right?

davidtaylorhq:
It depends on the deployment. Some environments (i.e. most self-hosters) will run all migrations in numerical order, regardless of pre/post. Other environments will run pre-migrations first, then post-migrations.

So, best strategy is to make the current post-migration a no-op, and introduce a new one with a higher number

PR #1095, opened 2025-01-28
DEV: Ween off deprecated plugin nav mode option

Drenmi:
### What is this change?

We have removed this flag in core. All plugins now use the "top mode" for their navigation. A backwards-compatible change has been made in core while we remove the usage from plugins.


.discourse-compatibility
```diff
@@ -1,3 +1,4 @@
+< 3.4.0.beta4-dev: a53719ab8eb071459f215227421b3ea4987e5f87
```
Drenmi, line 1: Technically covered by the below line, but I think this is correct, in case of the PR adding the below being reverted, etc.

martin-brennan (approved):

PR #1119, opened 2025-02-10
FIX: Check post action creator result when flagging spam

martin-brennan:
Currently in core re-flagging something that is already flagged as spam
is not supported, long term we may want to support this but in the meantime
we should not be silencing/hiding if the PostActionCreator fails
when flagging things as spam.

### DIFF
--- a/lib/ai_moderation/spam_scanner.rb
+++ b/lib/ai_moderation/spam_scanner.rb
@@ -394,22 +394,27 @@ module DiscourseAi
             queue_for_review: true,
           ).perform
 
-        log.update!(reviewable: result.reviewable)
+        # Currently in core re-flagging something that is already flagged as spam
+        # is not supported, long term we may want to support this but in the meantime
+        # we should not be silencing/hiding if the PostActionCreator fails.
+        if result.success?
+          log.update!(reviewable: result.reviewable)
 
-        reason = I18n.t("discourse_ai.spam_detection.silence_reason", url: url)
-        silencer =
-          UserSilencer.new(
-            post.user,
-            flagging_user,
-            message: :too_many_spam_flags,
-            post_id: post.id,
-            reason: reason,
-            keep_posts: true,
-          )
-        silencer.silence
+          reason = I18n.t("discourse_ai.spam_detection.silence_reason", url: url)
+          silencer =
+            UserSilencer.new(
+              post.user,
+              flagging_user,
+              message: :too_many_spam_flags,
+              post_id: post.id,
+              reason: reason,
+              keep_posts: true,
+            )
+          silencer.silence
 
-        # silencer will not hide tl1 posts, so we do this here
-        hide_post(post)
+          # silencer will not hide tl1 posts, so we do this here
+          hide_post(post)
+        end
       end
 
       def self.hide_post(post)
--- a/lib/ai_moderation/spam_scanner.rb
+++ b/lib/ai_moderation/spam_scanner.rb
@@ -414,6 +414,11 @@ module DiscourseAi
 
           # silencer will not hide tl1 posts, so we do this here
           hide_post(post)
+        else
+          log.update!(
+            error:
+              "unable to flag post as spam, post action failed for post #{post.id} with error: '#{result.errors.full_messages.join(", ").truncate(3000)}'",
+          )
         end
       end
 
--- a/lib/automation/llm_triage.rb
+++ b/lib/automation/llm_triage.rb
@@ -88,15 +88,18 @@ module DiscourseAi
                 .sub("%%AUTOMATION_NAME%%", automation&.name.to_s)
 
             if flag_type == :spam || flag_type == :spam_silence
-              PostActionCreator.new(
-                Discourse.system_user,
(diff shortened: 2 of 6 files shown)


spec/lib/modules/ai_moderation/spam_scanner_spec.rb
```diff
@@ -288,6 +288,31 @@
...
+        post_action_type_id: PostActionType.types[:spam],
+      )
+
+      described_class.new_post(post)
+
+      prompt = nil
+      DiscourseAi::Completions::Llm.with_prepared_responses(["spam"]) do |_, _, _prompts|
+        # force a rebake so we actually scan
+        post.rebake!
+        prompt = _prompts.first
```
Drenmi, line 307: Doesn't look like this assignment is used? 🤔

spec/lib/modules/automation/llm_triage_spec.rb
```diff
@@ -128,6 +128,28 @@ def triage(**args)
...
+        model: "custom:#{llm_model.id}",
+        system_prompt: "test %%POST%%",
+        search_for_text: "bad",
+        flag_post: true,
+        flag_type: :spam_silence,
+        automation: nil,
+      )
+    end
+
+    expect(post.user.silenced?).to eq(false)
```
Drenmi, line 150:
```suggestion
    expect(post.user.).not_to be_silenced
```

Drenmi (approved):
LGTM! Some small nits. Can't answer the question you asked from Sam. 🚀

lib/automation/llm_triage.rb
```diff
@@ -88,15 +88,18 @@ def self.handle(
...
-              PostActionCreator.new(
-                Discourse.system_user,
-                post,
-                PostActionType.types[:spam],
-                message: score_reason,
-                queue_for_review: true,
-              ).perform
+              result =
+                PostActionCreator.new(
+                  Discourse.system_user,
```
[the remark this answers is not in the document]
SamSaffron: I am not sure, technically it probably should use a triage user, we should create one on demand for it, cause it could be used for stuff other than spam

If you feel like it, change it, but if you prefer to leave it is fine. 

For example "if you notice a highly argumentative post that is violating our community guidelines, flag" ... not spam... just triage doing triage.

lib/ai_moderation/spam_scanner.rb
```diff
@@ -394,22 +394,27 @@ def self.handle_spam(post, log)
...
-            message: :too_many_spam_flags,
-            post_id: post.id,
-            reason: reason,
-            keep_posts: true,
-          )
-        silencer.silence
+        # Currently in core re-flagging something that is already flagged as spam
+        # is not supported, long term we may want to support this but in the meantime
+        # we should not be silencing/hiding if the PostActionCreator fails.
+        if result.success?
```
SamSaffron, line 400: maybe we log something somewhere so we know it happened, do we have room in the ai table to add information that we tried to flag but it did not work?

SamSaffron (approved):

SamSaffron:
approving, this is already better than what we have, but an audit trail that we tried to flag and had no luck would be nice for future debugging.

lib/ai_moderation/spam_scanner.rb
martin-brennan: Good point, will see what I can add

lib/automation/llm_triage.rb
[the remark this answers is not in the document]
martin-brennan: Ok will leave it for now

lib/ai_moderation/spam_scanner.rb
```diff
@@ -394,22 +394,32 @@ def self.handle_spam(post, log)
...
-            message: :too_many_spam_flags,
-            post_id: post.id,
-            reason: reason,
-            keep_posts: true,
+          # silencer will not hide tl1 posts, so we do this here
+          hide_post(post)
+        else
+          log.update!(
+            error:
+              "unable to flag post as spam, post action failed for post #{post.id} with error: '#{result.errors.full_messages.join(", ")}'",
```
SamSaffron, line 420: this will error out when errors are longer than 3000 :) we probably need to truncate.

SamSaffron:
nice I like this approach.. do we want error as text or JSON ? It probably does not matter that much but JSON does have some advantages given it is structured but the downside is that you can not easily limit length.

lib/ai_moderation/spam_scanner.rb
martin-brennan: Yeah was wondering about that, I don't _think_ the errors will be that long, but will truncate just in case :)

martin-brennan:
@SamSaffron I think the text is fine. Core only gives an error message, and not a super useful one at that since it's user readable for example "Oops! You already performed this action. Can you try refreshing the page?". I think it will be enough of a clue for debugging, maybe in future `PostActionCreator` could give a failed reason enum or something

SamSaffron:
no worries, I am good to have this merged :hugs:

PR #1127, opened 2025-02-16
FEATURE: Native PDF support

SamSaffron:
This amends it so we use PDF Reader gem to extract text from PDFs


config/settings.yml
```diff
@@ -355,6 +355,6 @@ discourse_ai:
     hidden: true
     type: list
 
-  ai_rag_pdf_images_enabled:
```
tgxworld, line 358: Just noting here that a DB migration to clean up the site setting is not required since the site setting was only introduced 3 days ago.

lib/utils/image_to_text.rb
```diff
@@ -127,6 +143,8 @@ def extract_text_from_page(page)
   end
 
   def extract_text_with_tesseract(page)
+    # return nil if we can not find tessaract binary
+    return nil if !Discourse::Utils.which("tesseract")
```
tgxworld, line 147:
```suggestion
    return nil if !self.class.tesseract_installed?
```

I think we can just use the new method you defined in this PR. Also not sure if `Discourse::Utils.which` is valid because I can't find the method definition anywhere.

lib/utils/pdf_to_images.rb
```diff
@@ -31,6 +29,7 @@ def extract_pages
 
       raise Discourse::InvalidParameters.new("Failed to download PDF") if pdf_path.nil?
 
+      temp_dir = Dir.mktmpdir("discourse-pdf-#{SecureRandom.hex(8)}")
```
tgxworld, line 32: Now that this is moved into the `begin..ensure` context, we need to change `FileUtils.rm_rf(temp_dir)` to `FileUtils.rm_rf(temp_dir) if temp_dir`

lib/utils/pdf_to_text.rb
```diff
@@ -0,0 +1,110 @@
...
+    @user = user
+    @llm_model = llm_model
+  end
+
+  def extract_text
+    pdf_path =
+      if upload.local?
+        Discourse.store.path_for(upload)
+      else
+        Discourse.store.download_safe(upload, max_file_size_kb: MAX_PDF_SIZE)&.path
```
tgxworld, line 48: The `MAX_PDF_SIZE ` constant is not defined in this class so this will blow up when S3 uploads are enabled. Should we add a test to cover this change here?

tgxworld, line 48: It doesn't seem like we are cleaning up this file after downloading it.

lib/utils/pdf_to_text.rb
```diff
@@ -0,0 +1,110 @@
...
+        "-auto-orient",
+        "-quality",
+        "85",
+        output_path,
+      ]
+
+      Discourse::Utils.execute_command(
+        *command,
+        failure_message: "Failed to convert PDF page #{page_number} to image",
+        timeout: 30,
```
tgxworld, line 95: Just confirming we don't call this method during a web request since it may cause a unicorn worker to timeout.

spec/lib/utils/pdf_to_text_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      pdf_to_text.extract_text { |page| pages << page }
+
+      expect(pages).to eq(["Page 1", "Page 2"])
+    end
+  end
+
+  context "when improving PDF extraction with LLM" do
+    it "can properly simulate a file" do
+      if ENV["CI"]
+        skip "This test requires imagemagick is installed with ghostscript support - which is not available in CI"
```
tgxworld, line 23: I'm alittle confused here because CI runs tests with the `discourse/discourse_test:release` image that contains the same imagemagick binary as production. If this feature doesn't work in CI, it would mean that it won't work in production as well.

config/settings.yml
SamSaffron: yeah not worth dealing with this imo, hidden setting, almost nobody changed it

lib/utils/image_to_text.rb
SamSaffron: yeah good catch, oops

lib/utils/pdf_to_images.rb
SamSaffron: good point will fix

lib/utils/pdf_to_text.rb
SamSaffron: oh yikes, will fix

lib/utils/pdf_to_text.rb
SamSaffron: does not run in unicorn only ever runs in sidekiq

spec/lib/utils/pdf_to_text_spec.rb
SamSaffron: I tried and it was missing gs... which is odd, I guess we can double check here, was sure @xfalcox added it.

SamSaffron:
@tgxworld pushed the fixes except for upload cleanup, not sure there is a good pattern for it in core

spec/lib/utils/pdf_to_text_spec.rb
xfalcox: @tgxworld looks like this regressed on https://github.com/discourse/discourse_docker/pull/889. Can you help us there?

xfalcox:
Needs a fix for regression introduced at https://github.com/discourse/discourse_docker/pull/889 making we miss gs

SamSaffron:
I think it is fine to merge this now cause the pdf OCR using gs was behind a hidden setting and remains like so after the merge of this change.

xfalcox (approved):

PR #1137, opened 2025-02-19
FIX: hardcoded require for evals

keegangeorge:
The require for `discourse/config/environment` should point to the local user's core Discourse environment instead of the hardcoded path for Sam's env.


romanrizzi (approved):

SamSaffron:
oops sorry :crying_cat_face:

PR #1139, opened 2025-02-20
FEATURE: Experimental search results from an AI Persona.

romanrizzi:
When a user searches discourse, we'll send the query to an AI Persona to provide additional context and enrich the results. The feature depends on the user being a member of a group to which the persona has access.

Screenshot 2025-02-20 at 11 20 32 AM
Screenshot 2025-02-20 at 11 21 01 AM
Screenshot 2025-02-20 at 11 21 26 AM
Screenshot 2025-02-20 at 11 21 32 AM


app/jobs/regular/stream_discover_reply.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Jobs
+  class StreamDiscoverReply < ::Jobs::Base
+    sidekiq_options retry: false
+
+    BUFFER_SIZE_WORDS = 50
```
xfalcox, line 7: where is this used?

xfalcox (approved):

PR #1142, opened 2025-02-21
DEV: Extract configs to a yml file and allow local config

nattsw:
Eval configs were hard-coded in `DiscourseAi::Evals::Llm`. This PR moves them into a config file.

Evaluators can now also define their own config locally with `config/eval-llms.local.yml`, and will have the choice to define the `api_key` attribute, or `api_key_env` if an env var is preferred.

Example file:
```
llms:
  discourse_llm:
    display_name: Qwen/Qwen2.5-32B-Instruct
    name: Qwen/Qwen2.5-32B-Instruct
    tokenizer: DiscourseAi::Tokenizer::OpenAiTokenizer
    api_key: <FILL IN KEY HERE>
    provider: open_router
    url: <FILL IN URL HERE>
    max_prompt_tokens: 128000
    vision_enabled: false
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 .env
 evals/log
 evals/cases
+config/eval-llms.local.yml
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@ For more information, please see: https://meta.discourse.org/t/discourse-ai/2592
 
 ### Evals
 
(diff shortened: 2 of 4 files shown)


SamSaffron (approved):

SamSaffron:
nice change :hugs:

PR #1149, opened 2025-02-24
FIX: legacy reasoning models not working, missing provider params

SamSaffron:
1. Legacy reasoning models (o1-preview / o1-mini) do not support developer or system messages, do not use them.
2. LLM editor form not showing all provider params due to missing remap

### DIFF
--- a/assets/javascripts/discourse/components/ai-llm-editor-form.gjs
+++ b/assets/javascripts/discourse/components/ai-llm-editor-form.gjs
@@ -61,7 +61,7 @@ export default class AiLlmEditorForm extends Component {
       provider: model.provider,
       enabled_chat_bot: model.enabled_chat_bot,
       vision_enabled: model.vision_enabled,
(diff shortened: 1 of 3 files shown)


tgxworld:
@SamSaffron I'm not familiar with the plugin and its constraints but I'm curious why we opted not to test the fix here.

SamSaffron:
The UI needs a system test, I agree with that, but the the backend is pretty fluid ad open ai break stuff all the time, so I think the backend fix is OK without a test. 

will see if I can add 1 system test for llm editing... we have none I guess.

tgxworld (approved):
Approving first as the fix looks good to me.

SamSaffron:
@tgxworld added a system test, oddly enough we covered a lot here just missed the edit case.

tgxworld:
> @tgxworld added a system test, oddly enough we covered a lot here just missed the edit case.

Looks good to me 👍

PR #1161, opened 2025-03-03
DEV: Updates to sentiment analysis reports

keegangeorge:
## ✨ What's new
**This PR includes a variety of updates to the Sentiment Analysis report:**
- [X] Conditionally showing sentiment reports based on `sentiment_enabled` setting
- [X] Sentiment reports should only be visible in sidebar if data is in the reports
- [X] Fix infinite loading of posts in drill down
- [x] Fix markdown emojis showing not showing as emoji representation
- [x] Drill down of posts should have URL
- [x] ~~Different doughnut sizing based on post count~~ [reverting and will address in follow-up (see: `/t/146786/47`)]
- [X] Hide non-functional export button 
- [X] Sticky drill down filter nav

### ← Before
before-z

### → After
after-z

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -25,9 +25,6 @@ discourse_ai:
   ai_sentiment_backfill_post_max_age_days:
     default: 60
     hidden: true
-  ai_sentiment_reports_enabled:
-    default: true
-    client: true
 
   ai_openai_dall_e_3_url: "https://api.openai.com/v1/images/generations"
   ai_openai_embeddings_url:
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -25,6 +25,9 @@ discourse_ai:
   ai_sentiment_backfill_post_max_age_days:
     default: 60
     hidden: true
+  ai_sentiment_reports_enabled:
```
xfalcox, line 28: why?

xfalcox (approved):

PR #1167, opened 2025-03-05
DEV: Add Ai summary persona siteSettings WIP

rngus2344:
## Background

Meta topic that brings up this change:
https://meta.discourse.org/t/ai-summarization-formatting-and-bugs/351163

## Changes 

This PR adds hidden site settings to the AI plugin:

- ai_summary_persona_id 
- ai_summary_consolidator_persona_id 

Then in the Ai summary methods, we see if those site settings are filled in. If so, we grab the prompt from the persona. If not, we use what is there now.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -358,3 +358,12 @@ discourse_ai:
   ai_rag_images_enabled:
     default: false
     hidden: true
+
+  ai_summary_persona_id:
+    default: ""
+    client: true
+    hidden: true
+  ai_summary_consolidator_persona_id:
+    default: ""
+    client: true
+    hidden: true
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -361,9 +361,7 @@ discourse_ai:
 
   ai_summary_persona_id:
     default: ""
-    client: true
     hidden: true
   ai_summary_consolidator_persona_id:
     default: ""
-    client: true
     hidden: true
--- a/lib/summarization/strategies/topic_summary.rb
+++ b/lib/summarization/strategies/topic_summary.rb
@@ -32,35 +32,41 @@ module DiscourseAi
           end
         end
 
+        def summary_prompt(topic_id:, persona_id:, default_prompt:)
+          prompt_text = default_prompt
+          if persona_id
+            persona = AiPersona.find_by(id: persona_id)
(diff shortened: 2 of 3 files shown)


config/settings.yml
```diff
@@ -358,3 +358,12 @@ discourse_ai:
   ai_rag_images_enabled:
     default: false
     hidden: true
+
+  ai_summary_persona_id:
+    default: ""
+    client: true
```
markvanlan, line 364: These don't need to be `client: true` (because the browser doesn't need to access them)

lib/summarization/strategies/topic_summary.rb
```diff
@@ -38,21 +38,29 @@ def summary_extension_prompt(summary, contents)
...
-            - Maintain the original language of the text being summarized.
-            - Aim for summaries to be 400 words or less.
-            - Each new post is formatted as "<POST_NUMBER>) <USERNAME> <MESSAGE>"
-            - Cite specific noteworthy posts using the format [DESCRIPTION](#{resource_path}/POST_NUMBER)
-              - Example: links to the 3rd and 6th posts by sam: sam ([#3](#{resource_path}/3), [#6](#{resource_path}/6))
-              - Example: link to the 6th post by jane: [agreed with](#{resource_path}/6)
-              - Example: link to the 13th post by joe: [joe](#{resource_path}/13)
-            - When formatting usernames either use @USERNAME or [USERNAME](#{resource_path}/POST_NUMBER)
-          TEXT
+          if SiteSetting.ai_summary_consolidator_persona_id
```
markvanlan, line 41: I think it would be good to abstract this logic into separate methods:
```ruby
def summary_prompt(topic_id:, persona_id:, default_prompt:)
  prompt_text = default_prompt
  if persona_id
    persona = AiPersona.find_by(id: persona_id)
    prompt_text = persona.system_prompt if persona
  end
  DiscourseAi::Completions::Prompt.new(prompt_text, topic_id: topic_id)
end

# In the two methods
prompt = summary_prompt(topic_id: target.id, persona_id:  SiteSetting.SETTING_NAME, default_prompt: <<~TEXT)
  put default text here
TEXT
 
...


```

rngus2344:
I might consolidate the two spec some.

PR #1182, opened 2025-03-10
DEV: Improve title suggester suggestions when editing topic

keegangeorge:
## :mag: Overview
This update ensures topic title suggestions when suggesting from edit topic take into account the whole topic for more accurate title suggestions.

### ← Before
Screenshot 2025-03-11 at 08 38 47

### → After
Screenshot 2025-03-11 at 08 36 18

### DIFF
--- a/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
+++ b/app/controllers/discourse_ai/ai_helper/assistant_controller.rb
@@ -64,7 +64,13 @@ module DiscourseAi
         raise Discourse::InvalidParameters.new(:mode) if !prompt
 
         hijack do
-          render json: DiscourseAi::AiHelper::Assistant.new.generate_and_send_prompt(prompt, input, current_user), status: 200
+          render json:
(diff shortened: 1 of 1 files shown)


romanrizzi (approved):

app/controllers/discourse_ai/ai_helper/assistant_controller.rb
```diff
@@ -53,19 +53,32 @@ def suggest
...
       def suggest_title
-        input = get_text_param!
+        assistant = DiscourseAi::AiHelper::Assistant.new
+
+        if params[:topic_id]
+          topic = Topic.find_by(id: params[:topic_id])
+
+          topic_content =
+            topic
+              .posts
```
xfalcox, line 63: In the summarization lib we have some logic to deal with mega topics that may be useful here to adopt or reuse.

Like if the topic has 10000 replies, we take the best 50 and so on...

xfalcox (approved):

PR #1183, opened 2025-03-11
DEV: fix Zeitwerk issue with SpamScanner class

lis2:
When spam scanner is enabled and code is reloaded, developer experience this error:

```
NameError at /posts
===================

uninitialized constant DiscourseAi::AiModeration::EntryPoint::SpamScanner

> To access an interactive console with this error, point your browser to: /__better_errors

plugins/discourse-ai/lib/ai_moderation/entry_point.rb, line 7
```

It is because when we call `SpamScanner` it is searched within parent `DiscourseAi::AiModeration::EntryPoint` namespace.

We can help a bit Zeitwerk by calling SpamScanner more explicitly.

### DIFF
--- a/lib/ai_moderation/entry_point.rb
+++ b/lib/ai_moderation/entry_point.rb
@@ -4,12 +4,18 @@ module DiscourseAi
   module AiModeration
     class EntryPoint
       def inject_into(plugin)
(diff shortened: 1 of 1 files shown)


martin-brennan (approved):
Thank you @lis2 , I appreciate it :pray:

PR #1189, opened 2025-03-12
DEV: Use existing topic embeddings when suggesting tags/categories on edit

keegangeorge:
When editing a topic (instead of creating one) and using the
tag/category suggestion buttons. We want to use existing topic
embeddings instead of creating new ones.

### DIFF
--- a/lib/ai_helper/semantic_categorizer.rb
+++ b/lib/ai_helper/semantic_categorizer.rb
@@ -11,7 +11,7 @@ module DiscourseAi
       end
 
       def categories
-        return [] if @text.blank? && !@topic_id
+        return [] if @text.blank? && @topic_id.nil?
         return [] if !DiscourseAi::Embeddings.enabled?
 
         candidates = nearest_neighbors
@@ -56,7 +56,7 @@ module DiscourseAi
       end
 
       def tags
-        return [] if @text.blank? && !@topic_id
+        return [] if @text.blank? && @topic_id.nil?
         return [] if !DiscourseAi::Embeddings.enabled?
 
         candidates = nearest_neighbors(limit: 100)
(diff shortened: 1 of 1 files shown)


lib/ai_helper/semantic_categorizer.rb
```diff
@@ -2,15 +2,16 @@
...
-        @text = input[:text]
+        @text = opts[:text]
         @vector = DiscourseAi::Embeddings::Vector.instance
         @schema = DiscourseAi::Embeddings::Schema.for(Topic)
+        @topic_id = opts[:topic_id]
       end
 
       def categories
-        return [] if @text.blank?
+        return [] if @text.blank? && !@topic_id
```
xfalcox, line 14: Nitpick but isn't `@topic_id.nil?` more readable than `!@topic_id`  ?

lib/ai_helper/semantic_categorizer.rb
```diff
@@ -100,7 +101,17 @@ def tags
       private
 
       def nearest_neighbors(limit: 50)
-        raw_vector = @vector.vector_from(@text)
+        if @topic_id
+          table_name = DiscourseAi::Embeddings::Schema::TOPICS_TABLE
+          embeddings =
+            DB
+              .query("SELECT embeddings::text FROM #{table_name} WHERE topic_id=#{@topic_id}")
```
xfalcox, line 108:
```suggestion
              .query("SELECT embeddings FROM #{table_name} WHERE topic_id = ?", topic_id)
```

How about that?

romanrizzi: Why not use `Schema#find_by_target`?

```ruby
target = Topic.new(id: @topic_id)
@schema.find_by_target(target)&.embeddings
```

xfalcox (approved):

PR #1196, opened 2025-03-14
UX: try AI search to side on large screens

awesomerobot:
Wanted to give this alternate layout a try, it's behind a media query so it will still work the same way it does now on smaller screens. 

This allows the topic results to still appear at the top, and removes the need to expand the AI result. 

image

Smaller screens don't change: 

image


brrusselburg (approved):
Oh thank god 😅 I support this wholeheartedly

PR #1200, opened 2025-03-18
WIP: Adding Enum button back to AI tools parameters form

martin-brennan:


assets/javascripts/discourse/admin/models/ai-tool.js
```diff
@@ -32,8 +32,10 @@ export default class AiTool extends RestModel {
 
         if (parameter.enum && parameter.enum.length) {
           parameter.enum = new TrackedArray(parameter.enum);
+          parameter.isEnum = true;
         } else {
           parameter.enum = null;
+          parameter.isEnum = false;
```
SamSaffron, line 38: I think we need this to be opaque to rest model cause we don't want to send it to the server

PR #1204, opened 2025-03-19
DEV: migrates tools form to form-kit

jjaffeux:
This PR is a retry of: https://github.com/discourse/discourse-ai/pull/1135, where we migrate `AiTools` form to FormKit. The previous PR accidentally removed code related to setting enum values, and as a result was reverted. This update includes enums correctly along with the previous updates.

### DIFF
--- a/assets/stylesheets/modules/ai-bot/common/ai-tools.scss
+++ b/assets/stylesheets/modules/ai-bot/common/ai-tools.scss
@@ -13,10 +13,12 @@
 
 .ai-tool-parameter__enum-values {
   margin-block: 1rem;
+
   .form-kit__container-content {
     display: grid;
     grid-template-columns: 1fr auto;
     position: relative;
+
     .form-kit__button.btn-icon-text {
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/ai-tool-editor-form.gjs
```diff
@@ -0,0 +1,359 @@
...
+          as |field|
+        >
+          <field.Custom>
+            <RagUploader
+              @target={{@editingModel}}
+              @updateUploads={{fn this.updateUploads form.addItemToCollection}}
+              @onRemove={{fn this.removeUpload form}}
+              @allowImages={{@settings.rag_images_enabled}}
+            />
+            <RagOptions
```
romanrizzi, line 327: @keegangeorge - I created an FK version of this component called `RagOptionsFK`, can we combine them now that this form uses form kit too?

xfalcox (approved):

PR #1206, opened 2025-03-20
DEV: add `guardian` argument to `TopicsFilter` callback

ZogStriP:
To match the updates in discourse/discourse#31908.

Note: no need to update the `.discourse-compatibility` as it's already pinned for the current beta.


ZogStriP:
Failure is expected until discourse/discourse#31908 gets merged.

xfalcox (approved):

PR #1210, opened 2025-03-23
FIX: bump system prompt limit

jjaffeux:
300 was too low, bumped it to 100_000


SamSaffron (approved):

jjaffeux:
Screenshot 2025-03-23 at 22 48 37
Screenshot 2025-03-23 at 22 48 40

PR #1223, opened 2025-03-26
DEV: Remove no longer needed click outside logic

keegangeorge:
The older design approach for sentiment analysis report needed the click outside logic. However, when the design was changed this logic was accidentally left behind. It is potentially causing some negative performance impacts. This PR removes the old unnecessary logic.


jjaffeux:
you need to remove hash too

jjaffeux (approved):
pre-approve; you just have to fix linting

PR #1246, opened 2025-04-04
FEATURE: allow to send LLM reports to groups

SamSaffron:
Previously we only allowed targeting users and emails with the ai llm report, we now also allow groups

### DIFF
--- a/lib/automation/report_runner.rb
+++ b/lib/automation/report_runner.rb
@@ -57,7 +57,15 @@ module DiscourseAi
         automation: nil
       )
         @sender = User.find_by(username: sender_username)
-        @receivers = User.where(username: receivers)
+        receivers_without_emails = receivers&.reject { |r| r.include? "@" }
+        if receivers_without_emails.present?
+          @group_receivers = Group.where(name: receivers_without_emails)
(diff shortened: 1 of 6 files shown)


lib/automation/report_runner.rb
```diff
@@ -57,7 +57,15 @@ def initialize(
         automation: nil
       )
         @sender = User.find_by(username: sender_username)
-        @receivers = User.where(username: receivers)
+        receivers_without_emails = receivers&.reject { |r| r.include? "@" }
+        if receivers_without_emails.present?
+          @group_receivers = Group.where(name: receivers_without_emails)
+          receivers_without_emails -= @group_receivers.pluck(:name)
+          @receivers = User.where(username: receivers_without_emails)
```
lis2, line 64: Edge case, but are you ok that if there is username same as group name, group will take precedence?

lis2 (approved):

lib/automation/report_runner.rb
SamSaffron: I think it is fine cause we try to share a namespace, you can not name a group the same as a user.

PR #1249, opened 2025-04-07
FIX: search discovery quirks

keegangeorge:
### :mag: Overview
This update fixes a few small quirks with Discobot discoveries:
- [X] subsequent searches should work correctly
- [X] discovery timeout should maintain UI
- [X] key up/down traversal of regular search results should still work

### :camera_flash: Screenshots
Screenshot 2025-04-07 at 12 25 03

### DIFF
--- a/assets/javascripts/discourse/components/ai-search-discoveries.gjs
+++ b/assets/javascripts/discourse/components/ai-search-discoveries.gjs
@@ -77,10 +77,8 @@ export default class AiSearchDiscoveries extends Component {
           return;
         }
 
-        this.search.handleArrowUpOrDown(event);
-
         if (this.discobotDiscoveries.lastQuery === this.query) {
-          return false;
+          return true;
         }
 
         if (event.key === "Enter" && this.query) {


awesomerobot (approved):

assets/javascripts/discourse/components/ai-search-discoveries.gjs
```diff
@@ -73,10 +73,12 @@ export default class AiSearchDiscoveries extends Component {
 
     withPluginApi((api) => {
       api.addSearchMenuOnKeyDownCallback((searchMenu, event) => {
-        if (!searchMenu || this.discobotDiscoveries.loadingDiscoveries) {
+        if (!searchMenu) {
           return;
         }
 
+        this.search.handleArrowUpOrDown(event);
```
pmusaraj, line 80: If I read the core api correctly, this shouldn't be necessary if the callbacks below all return `true`. We only need this because returning `false` somewhere here will skip the default core event handling (for arrows, for Enter, for Escape).

https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/components/search-menu/search-term.gjs#L59-L68

We could do `return true` on each case where we return here, haven't tried it, but I suspect that should allow core's handling to work as expected.

assets/javascripts/discourse/components/ai-search-discoveries.gjs
```diff
@@ -158,6 +160,11 @@ export default class AiSearchDiscoveries extends Component {
     if (!personas) {
       return false;
     }
+
+    if (this.discobotDiscoveries.discoveryTimedOut) {
+      return false;
```
pmusaraj, line 165: what happens if we `return true` here?

pmusaraj (approved):

PR #1252, opened 2025-04-08
DEV: Use structured responses for summaries

romanrizzi:
System personas:

Screenshot 2025-04-14 at 2 38 28 PM

---

Regular personas:

Screenshot 2025-04-14 at 2 38 46 PM

---

Schema editor:

Screenshot 2025-04-14 at 2 40 36 PM

### DIFF
--- a/spec/lib/completions/endpoints/anthropic_spec.rb
+++ b/spec/lib/completions/endpoints/anthropic_spec.rb
@@ -837,15 +837,15 @@ RSpec.describe DiscourseAi::Completions::Endpoints::Anthropic do
         },
       ).to_return(status: 200, body: body)
 
-      result = +""
+      structured_output = nil
       llm.generate(
         prompt,
         user: Discourse.system_user,
         feature_name: "testing",
         response_format: schema,
-      ) { |partial, cancel| result << partial }
+      ) { |partial, cancel| structured_output = partial }
 
-      expect(result).to eq("\"key\":\"Hello!\"}")
+      expect(structured_output.full_output).to eq({ key: "Hello!" })
 
       expected_body = {
         model: "claude-3-opus-20240229",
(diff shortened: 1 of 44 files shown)


spec/lib/completions/endpoints/anthropic_spec.rb
```diff
@@ -769,4 +769,92 @@
...
+
+      result = +""
+      llm.generate(
+        prompt,
+        user: Discourse.system_user,
+        feature_name: "testing",
+        response_format: schema,
+      ) { |partial, cancel| result << partial }
+
+      expect(result).to eq("\"key\":\"Hello!\"}")
```
SamSaffron, line 848: this is a bit surprising imo... it should be:  {key: "Hello!"} or perhaps a special object that inherits off hash eg: StructuredOutput , then nobody needs to worry about parsing and we can enforce structure at the boundary.

xfalcox: What about we use `JSON.parse('result', object_class: OpenStruct)` so we always have a proper object to work with?

romanrizzi: I don't think that will work. During streaming, we only get chunks of the JSON so we cannot use `JSON#parse`. 

I believe Sam's proposal is to have a class that handles the buffering and parsing while decoding the response. This class reads the chunk, keeps track of which key/value are we at, and updates the state:

```
Chunk 1: "{\"message\": \"Line 1\\"
Partial: {"message"=>"Line 1"}

Chunk 2: "nLine 2\", \"status\": \"o"
Partial: {"message"=>"Line 1\nLine 2", "status"=>"o"}

Chunk 3: "k\"}"
Partial: {"message"=>"Line 1\nLine 2", "status"=>"ok"}
```

By taking care of it internally, the burden of doing this doesn't fall on the modules using structured outputs, like in this case, `FoldContent`.

PR #1257, opened 2025-04-11
FIX: invalid context construction for responders

SamSaffron:
Previous to this fix we assumed the name field contained usernames
when in fact it was stored in the id field.

This fixes the context contruction and also adds some basic user
information to the context to assist responders in understanding
the cast of chars


SamSaffron:
Agree, will improve the story around this code next, but I want to get the urgent fix in first.

PR #1268, opened 2025-04-17
FEATURE: Add spending metrics to AI usage

keegangeorge:
## :mag: Overview
This update adds metrics for estimated spending in AI usage. To make use of it, admins must add cost details to the LLM config page (input, output, and cached input costs per 1M tokens). After doing so, the metrics will appear in the AI usage dashboard as the AI plugin is used.

## 📸 Screenshots

llm-config-page
tile-spending
table-spending

### DIFF
--- a/app/controllers/discourse_ai/admin/ai_llms_controller.rb
+++ b/app/controllers/discourse_ai/admin/ai_llms_controller.rb
@@ -157,6 +157,7 @@ module DiscourseAi
             :provider,
             :tokenizer,
             :max_prompt_tokens,
+            :max_output_tokens,
             :api_key,
             :enabled_chat_bot,
             :vision_enabled,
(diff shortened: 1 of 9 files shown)


xfalcox:
Can you share a screenshot of how it looks when the costs are nil ? Both the form and the usage page

xfalcox:
@awesomerobot do you want to do a pass on the form?

xfalcox:
> > Can you share a screenshot of how it looks when the costs are nil ? Both the form and the usage page
> 
> @xfalcox The form just has blank input. And the usage page shows $0, but I have this tooltip: Screenshot 2025-04-17 at 11 56 15

I wonder if we should hide the total box and the columns when it's all zeroes...

On one hand it showcases the feature so people know that they can fill that, on the other hand it may be just noise if no ones fills it.

xfalcox (approved):

PR #1276, opened 2025-04-23
FEATURE: Allow for persona & llm selection in bot conversations page

markvanlan:
This PR
1. Abstracts the persona/llm dropdowns to a component to be used in both the composer connector AND the conversations page. (and adds them to the conversations page)
2. Removes reliance on the composer for creating bot PMs on the new page. Now we simply make an ajax request and transition to the new post. This did require a new topic custom field, so that when we navigate, we don't have to rely on the "are any of the posts bot posts" logic -- that doesn't work when you transition immediately, as the bot hasn't posted yet. I don't think having this topic custom field is a bad thing.
3. Simplify and improve how the update the topic titles in the sidebar on the left. No more updating DOM directly.

Screenshot 2025-04-23 at 10 50 09 AM

### DIFF
--- a/assets/javascripts/discourse/components/ai-persona-llm-selector.gjs
+++ b/assets/javascripts/discourse/components/ai-persona-llm-selector.gjs
@@ -4,6 +4,7 @@ import { hash } from "@ember/helper";
 import { next } from "@ember/runloop";
 import { service } from "@ember/service";
 import KeyValueStore from "discourse/lib/key-value-store";
+import { i18n } from "discourse-i18n";
 import DropdownSelectBox from "select-kit/components/dropdown-select-box";
 
 export default class AiPersonaLlmSelector extends Component {
@@ -88,7 +89,7 @@ export default class AiPersonaLlmSelector extends Component {
   }
 
   get filterable() {
-    return this.botOptions.length > 4;
+    return this.botOptions.length > 8;
   }
 
   get value() {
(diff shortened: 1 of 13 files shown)


lib/ai_bot/entry_point.rb
```diff
@@ -102,6 +105,14 @@ def inject_into(plugin)
           Rails.root.join("plugins", "discourse-ai", "db", "fixtures", "ai_bot"),
         )
 
+        plugin.add_to_serializer(
+          :topic_view,
+          :is_bot_pm,
+          include_condition: -> do
+            object.personal_message && object.topic.custom_fields[TOPIC_AI_BOT_PM_FIELD]
```
xfalcox, line 112: What about existing PMs?

markvanlan: @xfalcox Doesn't mess with them.. they have an existing bot PM. This would leave room in the future for us to somehow backfill.. but this is really to address the case of 
* PM is sent
* Navigate to PM
* There is no response yet.. we don't have any clue it's a bot PM (and the mechanisms to determine can have false-positive like the case with auto-responder)

I didn't remove the conditionals for checking if there is a bot PM, so old PMs still function.

assets/javascripts/discourse/controllers/discourse-ai-bot-conversations.js
```diff
@@ -1,48 +1,28 @@
...
+import { alias } from "@ember/object/computed";
 import { service } from "@ember/service";
-import { tracked } from "@ember-compat/tracked-built-ins";
 
 export default class DiscourseAiBotConversations extends Controller {
   @service aiBotConversationsHiddenSubmit;
   @service currentUser;
 
-  @tracked selectedPersona = this.personaOptions[0].username;
+  @alias("aiBotConversationsHiddenSubmit.loading") loading;
```
jjaffeux, line 10:
```
get loading() {
  return this.aiBotConversationsHiddenSubmit?.loading;
}
```

assets/javascripts/discourse/components/ai-persona-llm-selector.gjs
```diff
@@ -0,0 +1,181 @@
+import Component from "@glimmer/component";
```
jjaffeux, line 1: what are the big change sin this file?

jjaffeux: Nothing just copy/pasted into its own component?

markvanlan: Simply abstracting out the "setting" of values -- they're passed in now. instead of `this.composer.metaData["personaId"] = id` we have `this.args.setPersonaId(id)`

lib/ai_bot/entry_point.rb
xfalcox: So we are adding a permanent custom field just to address the temporary case of detection during the first transition? I don't really agree littering the database permanently is a good strategy in this case.

Can't this be handled by a JS state, URL param, or anything else that has a more appropriate lifetime ?

assets/javascripts/discourse/components/ai-persona-llm-selector.gjs
```diff
@@ -0,0 +1,192 @@
...
+      }
+
+      return enabledPersonas.map((persona) => {
+        return {
+          id: persona.id,
+          name: persona.name,
+          description: persona.description,
+        };
+      });
+    }
```
megothss, line 85:
```suggestion
    if (!this.currentUser.ai_enabled_personas) {
      return;
    }

    let enabledPersonas = this.currentUser.ai_enabled_personas;

    if (!this.hasLlmSelector) {
      enabledPersonas = enabledPersonas.filter((persona) => persona.username);
    }

    return enabledPersonas.map((persona) => {
      return {
        id: persona.id,
        name: persona.name,
        description: persona.description,
      };
    });
```

xfalcox (approved):

janzenisaac (approved):

PR #1280, opened 2025-04-24
FIX: eat all leading spaces llms provide when they stream them

SamSaffron:


SamSaffron:
Chat errors out if you try to create a message with a blank body.

jjaffeux (approved):

PR #1282, opened 2025-04-24
DEV: Migration to backfill bot PM custom field

markvanlan:
In the last commit, I introduced a topic_custom_field to determine if a PM is indeed a bot PM.

This commit adds a migration to backfill any PM that is between 1 real user, and 1 bot. The correct topic_custom_field is added for these, so they will appear on the bot conversation sidebar properly.

We can also drop the joining to topic_users in the controller for sidebar conversations, and the `isPostFromAiBot` logic from the sidebar.

### DIFF
--- a/app/controllers/discourse_ai/ai_bot/conversations_controller.rb
+++ b/app/controllers/discourse_ai/ai_bot/conversations_controller.rb
@@ -15,8 +15,8 @@ module DiscourseAi
           Topic
             .private_messages_for_user(current_user)
             .where(user: current_user) # Only show PMs where the current user is the author
-            .joins(:topic_users)
-            .where(topic_users: { user_id: bot_user_ids })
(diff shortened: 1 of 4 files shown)


app/controllers/discourse_ai/ai_bot/conversations_controller.rb
```diff
@@ -15,8 +15,8 @@ def index
           Topic
             .private_messages_for_user(current_user)
             .where(user: current_user) # Only show PMs where the current user is the author
-            .joins(:topic_users)
-            .where(topic_users: { user_id: bot_user_ids })
+            .joins("INNER JOIN topic_custom_fields tcf ON tcf.topic_id = topics.id")
+            .where("tcf.name = ? AND tcf.value = 't'", DiscourseAi::AiBot::TOPIC_AI_BOT_PM_FIELD)
```
xfalcox, line 19: can you move the `where` conditions to the `join`?

xfalcox (approved):

PR #1293, opened 2025-04-29
FEATURE: add OpenAI image generation and editing capabilities

SamSaffron:
This commit enhances the AI image generation functionality by adding support for:

1. OpenAI's GPT-based image generation model (gpt-image-1)
2. Image editing capabilities through the OpenAI API
3. A new "Designer" persona specialized in image generation and editing
4. Two new AI tools: CreateImage and EditImage

Technical changes include:
- Renaming `ai_openai_dall_e_3_url` to `ai_openai_image_generation_url` with a migration
- Adding `ai_openai_image_edit_url` setting for the image edit API endpoint
- Refactoring image generation code to handle both DALL-E and the newer GPT models
- Supporting multipart/form-data for image editing requests

### DIFF
--- a/lib/inference/open_ai_image_generator.rb
+++ b/lib/inference/open_ai_image_generator.rb
@@ -161,9 +161,7 @@ module ::DiscourseAi
           end
         end
 
-        threads_complete = false
-        threads_complete = threads.all? { |t| t.join(2) } until threads_complete
-
+        threads.each(&:join)
         threads.filter_map(&:value)
       end
 
@@ -343,25 +341,8 @@ module ::DiscourseAi
               end
             image_data = File.read(image_path)
             image_filename = File.basename(image.url)
-          elsif image.is_a?(File) || image.is_a?(Tempfile)
-            image_data = File.read(image.path)
-            image_filename = File.basename(image.path)
-          elsif image.is_a?(String) && File.exist?(image)
-            image_data = File.read(image)
-            image_filename = File.basename(image)
-          elsif image.is_a?(String) && image.start_with?("http")
-            # Download image from URL
-            image_temp =
-              FileHelper.download(
-                image,
-                max_file_size: 25.megabytes,
-                tmp_file_name: "edit_image_download",
-                follow_redirect: true,
-              )
-            image_data = File.read(image_temp)
-            image_filename = File.basename(image)
           else
-            raise "Unsupported image format. Must be Upload, File, path string, or URL."
+            raise "Unsupported image format. Must be an Upload"
           end
 
           body << "--#{boundary}\r\n"
--- a/lib/inference/open_ai_image_generator.rb
+++ b/lib/inference/open_ai_image_generator.rb
@@ -5,12 +5,240 @@ module ::DiscourseAi
     class OpenAiImageGenerator
       TIMEOUT = 60
 
-      def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, api_url: nil)
+      def self.create_uploads!(
+        prompts,
+        model:,
+        size: nil,
+        api_key: nil,
+        api_url: nil,
+        user_id:,
+        for_private_message: false,
+        n: 1,
+        quality: nil,
+        style: nil,
+        background: nil,
+        moderation: "low",
+        output_compression: nil,
+        output_format: nil,
+        title: nil
+      )
+        # Get the API responses in parallel threads
+        api_responses =
+          generate_images_in_threads(
+            prompts,
+            model: model,
+            size: size,
+            api_key: api_key,
+            api_url: api_url,
+            n: n,
+            quality: quality,
+            style: style,
+            background: background,
+            moderation: moderation,
+            output_compression: output_compression,
+            output_format: output_format,
+          )
+
+        create_uploads_from_responses(api_responses, user_id, for_private_message, title)
+      end
+
+      # Method for image editing that returns Upload objects
+      def self.create_edited_upload!(
+        images,
+        prompt,
+        mask: nil,
+        model: "gpt-image-1",
+        size: "auto",
+        api_key: nil,
+        api_url: nil,
+        user_id:,
+        for_private_message: false,
+        n: 1,
+        quality: nil
+      )
+        api_response =
+          edit_images(
+            images,
+            prompt,
+            mask: mask,
+            model: model,
+            size: size,
+            api_key: api_key,
+            api_url: api_url,
+            n: n,
+            quality: quality,
+          )
+
+        # Create uploads in the main thread
+        create_uploads_from_responses([api_response], user_id, for_private_message).first
+      end
+
+      # Common method to create uploads from API responses
+      def self.create_uploads_from_responses(
+        api_responses,
+        user_id,
+        for_private_message,
+        title = nil
+      )
+        all_uploads = []
+
+        api_responses.each do |response|
+          next unless response
+
+          response[:data].each_with_index do |image, index|
+            Tempfile.create("ai_image_#{index}.png") do |file|
+              file.binmode
+              file.write(Base64.decode64(image[:b64_json]))
+              file.rewind
+
+              upload =
+                UploadCreator.new(
+                  file,
+                  title || "image.png",
+                  for_private_message: for_private_message,
+                ).create_for(user_id)
+
+              all_uploads << {
+                # Use revised_prompt if available (DALL-E 3), otherwise use original prompt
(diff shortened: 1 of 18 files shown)


SamSaffron:
odd, failure is a flake... having a think about how to fix

lib/inference/open_ai_image_generator.rb
```diff
@@ -5,12 +5,239 @@ module Inference
...
+              retry if attempts < 3
+              Discourse.warn_exception(e, message: "Failed to generate image for prompt #{prompt}")
+              puts "Error generating image for prompt: #{prompt} #{e}" if Rails.env.development?
+              nil
+            end
+          end
+        end
+
+        threads_complete = false
+        threads_complete = threads.all? { |t| t.join(2) } until threads_complete
```
tgxworld, line 165:
```suggestion
        ThreadsWait.all_waits(*threads)
```

I think we can lean on the Ruby stdlib here. Provided in Ruby via https://github.com/ruby/thwait

lib/personas/designer.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+
+      def system_prompt
+        <<~PROMPT
+            You are designerbot and you are here to help people generate and edit images.
+
+            - A good prompt needs to be detailed and specific.
+            - You can specify subject, medium (e.g. oil on canvas), artist (person who drew it or photographed it)
+            - You can specify details about lighting or time of day.
+            - You can specify a particular website you would like to emulate (artstation or deviantart)
```
tgxworld, line 21: I'm not sure if there is a legal basis for my worry but I wonder if we want to avoid publicly naming specific websites here.

lib/inference/open_ai_image_generator.rb
SamSaffron: but we would need to add a gem here for this? the 2 second wait is a bit odd... I think we can simplify to a .join maybe

lib/personas/designer.rb
SamSaffron: Its too late, this stuff is in training data of all the models

lib/inference/open_ai_image_generator.rb
```diff
@@ -45,6 +288,185 @@ def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, ap
...
+          image_data = nil
+          image_filename = nil
+
+          # Handle different image input types
+          if image.is_a?(Upload)
+            image_path =
+              if image.local?
+                Discourse.store.path_for(image)
+              else
+                Discourse.store.download_safe(image, max_file_size_kb: MAX_IMAGE_SIZE)&.path
```
tgxworld, line 342: This creates a tempfile which we need to clean up after reading.

lib/inference/open_ai_image_generator.rb
SamSaffron: https://github.com/ruby/thwait/blob/master/lib/thwait.rb

lib/inference/open_ai_image_generator.rb
```diff
@@ -45,6 +288,185 @@ def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, ap
...
+            image_filename = File.basename(image.url)
+          elsif image.is_a?(File) || image.is_a?(Tempfile)
+            image_data = File.read(image.path)
+            image_filename = File.basename(image.path)
+          elsif image.is_a?(String) && File.exist?(image)
+            image_data = File.read(image)
+            image_filename = File.basename(image)
+          elsif image.is_a?(String) && image.start_with?("http")
+            # Download image from URL
+            image_temp =
```
tgxworld, line 354: Same as https://github.com/discourse/discourse-ai/pull/1293/files#r2065599634

lib/inference/open_ai_image_generator.rb
SamSaffron: I think this is mainly useful for signalling reading the code, not seeing what it gains if we don't need to signal

tgxworld: > I think we can simplify to a .join maybe

`threads.each(&:join)` may be simpler?

lib/inference/open_ai_image_generator.rb
SamSaffron: looking at core, not seeing us doing any cleanup when we download safe?

lib/inference/open_ai_image_generator.rb
SamSaffron: yeah switching to that.

lib/personas/tools/edit_image.rb
```diff
@@ -0,0 +1,87 @@
...
+          return { prompt: prompt, error: "No valid images provided" } if image_urls.blank?
+
+          uploads =
+            image_urls
+              .map do |url|
+                sha1 = Upload.sha1_from_short_url(url)
+                Upload.find_by(sha1: sha1)
+              end
+              .compact
+              .take(10)
```
tgxworld, line 59:
```suggestion
          sha1s =
            image_urls
              .map do |url|
                Upload.sha1_from_short_url(url)
              end
              .compact

          uploads = Upload.where(sha1: sha1s).order(created_by: :asc).limit(10)
```
Some code golfing to prevent an N+1 queries problem and also prevent over fetching.

lib/inference/open_ai_image_generator.rb
SamSaffron: medium straight forward cause we need extra tracking to do that, I don't mind nuking it early, but it complicates the code

spec/lib/personas/tools/create_image_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      expect(JSON.parse(info)).to eq(
+        {
+          "prompts" => [
+            {
+              "prompt" => "a watercolor painting of flowers",
+              "url" => "upload://pv9zsrM93Jz3U8xELTJCPYU2DD0.png",
+            },
+            {
+              "prompt" => "a watercolor painting of flowers",
+              "url" => "upload://pv9zsrM93Jz3U8xELTJCPYU2DD0.png",
```
tgxworld, line 51: Sorry as this is confusing to me. Why do we return two prompts with the same `prompt` and `url` attributes?

lib/personas/tools/edit_image.rb
SamSaffron: yeah seems fine, thanks

spec/lib/personas/tools/create_image_spec.rb
SamSaffron: it got remapped by the model

lib/inference/open_ai_image_generator.rb
SamSaffron: ok I cleaned it up in my last commit

lib/inference/open_ai_image_generator.rb
```diff
@@ -45,6 +286,168 @@ def self.perform!(prompt, model: "dall-e-3", size: "1024x1024", api_key: nil, ap
...
+
+      def self.perform_edit_api_call!(
+        images,
+        prompt,
+        mask: nil,
+        model: "gpt-image-1",
+        size: "auto",
+        api_key:,
+        api_url:,
+        n: 1,
```
tgxworld, line 300: I wonder if there is a more descriptive name we can use for this kwarg. Not sure what `n` stands for here without having to read through method.

SamSaffron: yeah its tricky, I map to what open ai call the param, we can remap internally, but generally we keep it at 1 anyway so maybe we can even remove for now.

lib/personas/tools/edit_image.rb
```diff
@@ -0,0 +1,87 @@
...
+          return { prompt: prompt, error: "No valid images provided" } if image_urls.blank?
+
+          sha1s =
+            image_urls
+              .map do |url|
+                Upload.sha1_from_short_url(url)
+              end
+              .compact
+
+          uploads = Upload.where(sha1: sha1s).order(created_by: :asc).limit(10)
```
tgxworld, line 59:
```suggestion
          uploads = Upload.where(sha1: sha1s).order(created_at: :asc).limit(10)
```

Sorry typo here in my previous suggestion.

tgxworld (approved):
Looks good to me 👍 Thank you for the changes.

lib/personas/tools/edit_image.rb
SamSaffron: also needed a to_a for later on, all sorted, thanks!

PR #1297, opened 2025-04-29
UX: Empty state for AI conversations sidebar & btn changes

markvanlan:
This PM adds an empty state when the user doesn't have any PM history. It ALSO retains the new conversation button in the sidebar so it no longer jumps. The button is disabled, icon, and text are all updated.

Relies on Core to render empty state: https://github.com/discourse/discourse/pull/32518

### Empty state:
Screenshot 2025-04-29 at 3 28 23 PM

### Button retained:
Screenshot 2025-04-29 at 3 27 30 PM


SamSaffron:
greying the button is visually confusing, I think we just leave it as clilckable and does nothing. 

"Enter Question" does not feel right though as text. "+ New Question" is much more natural.

PR #1306, opened 2025-05-01
FIX: better LLM feedback for image generation failures 

SamSaffron:
When image edit or creation fails for any sort of reason we should properly inform the LLM of what happened. 

This allows it to self correct.

### DIFF
--- a/lib/inference/open_ai_image_generator.rb
+++ b/lib/inference/open_ai_image_generator.rb
@@ -42,8 +42,7 @@ module ::DiscourseAi
 
         raise api_responses[0] if api_responses.all? { |resp| resp.is_a?(StandardError) }
 
-        api_responses =
-          api_responses.filter { |response| !response.is_a?(StandardError) }
+        api_responses = api_responses.filter { |response| !response.is_a?(StandardError) }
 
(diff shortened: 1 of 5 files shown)


lib/inference/open_ai_image_generator.rb
```diff
@@ -40,6 +40,11 @@ def self.create_uploads!(
             output_format: output_format,
           )
 
+        raise api_responses[0] if api_responses.all? { |resp| resp.is_a?(StandardError) }
+
+        api_responses =
+          api_responses.filter_map { |response| response.is_a?(StandardError) ? nil : response }
```
lis2, line 46:
```suggestion
        api_responses =
          api_responses.filter { |response| !response.is_a?(StandardError) }
```
Do you need those `nil` leftover from StandardError?

lis2 (approved):

SamSaffron:
Filter map gets rid of them :)

lib/inference/open_ai_image_generator.rb
SamSaffron: agree this is simpler ... filter_map removed them, but no point mapping

PR #1309, opened 2025-05-05
FEATURE: enforce jpg/png for all images

SamSaffron:
This ensures webp / gif are converted to png prior to sending
to LLMs given webp and gif are not evenly supported.

png/jpg is universally supported and are the only supported format.

longer term we need to add support for audio/video/pdf which is supported by some models.


spec/lib/completions/upload_encoder_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
jjaffeux, line 1: we should also have webp and jpeg

lib/completions/upload_encoder.rb
```diff
@@ -10,6 +10,14 @@ def self.encode(upload_ids:, max_pixels:)
           next if upload.blank?
           next if upload.width.to_i == 0 || upload.height.to_i == 0
 
+          desired_extension = upload.extension
+          desired_extension = "png" if upload.extension == "gif"
+          desired_extension = "png" if upload.extension == "webp"
+          desired_extension = "jpeg" if upload.extension == "jpg"
+
+          # this keeps it very simple format wise given everyone supports png and jpg
+          next if !%w[jpeg png].include?(desired_extension)
```
jjaffeux, line 19: should we log domething here?

jjaffeux (approved):

lib/completions/upload_encoder.rb
SamSaffron: long term maybe, for now I think skipping obscure image formats is ok

spec/lib/completions/upload_encoder_spec.rb
SamSaffron: no worries, added

ZogStriP (approved):

PR #1313, opened 2025-05-06
FEATURE: forum researcher persona for deep research

SamSaffron:
This commit introduces a new Forum Researcher persona specialized in deep forum content analysis along with comprehensive improvements to our AI infrastructure.

Key additions:
- New Forum Researcher persona with advanced filtering and analysis capabilities
- Robust filtering system supporting tags, categories, dates, users, and keywords
- LLM formatter to efficiently process and chunk research results

Infrastructure improvements:
- Implemented CancelManager class to centrally manage AI completion cancellations
- Replaced callback-based cancellation with a more robust pattern
- Added systematic cancellation monitoring with callbacks

Other improvements:
- Added configurable `default_enabled` flag to control which personas are enabled by default
- Updated translation strings for the new researcher functionality
- Added comprehensive specs for the new components

This change makes our AI platform more stable while adding powerful research capabilities that can analyze forum trends and surface relevant content.

### DIFF
--- a/assets/javascripts/discourse/lib/ai-streamer/progress-handlers.js
+++ b/assets/javascripts/discourse/lib/ai-streamer/progress-handlers.js
@@ -2,7 +2,7 @@ import { later } from "@ember/runloop";
 import PostUpdater from "./updaters/post-updater";
 
 const PROGRESS_INTERVAL = 40;
-const GIVE_UP_INTERVAL = 60000;
+const GIVE_UP_INTERVAL = 600000; // 10 minutes which is our max thinking time for now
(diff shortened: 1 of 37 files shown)


SamSaffron:
I actually regret bubbling it from the block cause it forces you to keep repeating it which is a bit confusing. Plus if you are waiting a minute for the first few bytes from the LLM you need to bubble it right away which means you say "here is no text, but in case you need to cancel you can use this" 

having it passed in gives us so much more consistency (now that endpoint/base and open ai image generation both share the pattern)

PR #1314, opened 2025-05-06
FIX: handle unexpected errors when browsing web

SamSaffron:


pento (approved):

SamSaffron:
I do wonder if we should blanket guard all tools, mixed ... maybe

PR #1318, opened 2025-05-06
FIX: Return structured output on non-streaming mode

romanrizzi:
Non-streaming mode should always return a `StructureOutput` when a response format is given. Multi-turn responses aren't supported so `response_data` is always a string.


janzenisaac:
The AI Homepage spec that is failing should be addressed separately. Going to approve :)

PR #1330, opened 2025-05-08
FEATURE: add inferred concepts system

xfalcox:
This commit adds a new inferred concepts system that:
- Creates a model for storing concept labels that can be applied to topics
- Provides AI personas for finding new concepts and matching existing ones
- Adds jobs for generating concepts from popular topics
- Includes a scheduled job that automatically processes engaging topics

### DIFF
--- a/lib/inferred_concepts/applier.rb
+++ b/lib/inferred_concepts/applier.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module InferredConcepts
+    class Applier
+      # Associates the provided concepts with a topic
+      # topic: a Topic instance
+      # concepts: an array of InferredConcept instances
+      def self.apply_to_topic(topic, concepts)
+        return if topic.blank? || concepts.blank?
+        
+        concepts.each do |concept|
+          # Use the join table to associate the concept with the topic
+          # Avoid duplicates by using find_or_create_by
+          ActiveRecord::Base.connection.execute(<<~SQL)
+            INSERT INTO topics_inferred_concepts (topic_id, inferred_concept_id, created_at, updated_at)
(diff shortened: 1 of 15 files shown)


lib/inferred_concepts/applier.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module InferredConcepts
+    class Applier
```
romanrizzi, line 5: Methods from this class that need a persona could be instance methods, and the persona could be passed in an initializer. DI makes it easier to test.

lib/personas/concept_matcher.rb
```diff
@@ -0,0 +1,43 @@
...
+            <o>
+              {"matching_concepts": ["concept1", "concept3", "concept5"]}
+            </o>
+
+          Only include concepts from the provided list that match the content. If no concepts match, return an empty array.
+        PROMPT
+      end
+
+      def response_format
+        [{ "key" => "matching_concepts", "type" => "array" }]
```
romanrizzi, line 39: We should add `array` as an option in the persona's form

romanrizzi (commented):
Looks very good. It'd be great to add some tests

lib/personas/bot_context.rb
```diff
@@ -68,7 +70,15 @@ def initialize(
...
-      TEMPLATE_PARAMS = %w[time site_url site_title site_description participants resource_url]
+      TEMPLATE_PARAMS = %w[
+        time
+        site_url
+        site_title
+        site_description
+        participants
+        resource_url
+        inferred_concepts
+      ]
```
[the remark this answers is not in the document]
romanrizzi: I think it's freezed by default, no?

PR #1331, opened 2025-05-09
UX: don't show AI suggestions in composer when inputs are disabled

awesomerobot:
Reported here: https://meta.discourse.org/t/disable-ai-title-generator-when-pm-title-cannot-be-edited/307595

This hides the composer AI suggestion buttons when the user can't edit the title/category/tags

### DIFF
--- a/assets/stylesheets/modules/ai-helper/common/ai-helper.scss
+++ b/assets/stylesheets/modules/ai-helper/common/ai-helper.scss
@@ -229,13 +229,13 @@
     padding-right: 2em;
   }
 
-  .category-chooser {
+  .category-chooser:not(.is-disabled) {
     .select-kit-header-wrapper {
       padding-right: 1.5em;
     }
   }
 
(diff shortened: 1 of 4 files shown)


assets/stylesheets/modules/ai-helper/common/ai-helper.scss
```diff
@@ -314,7 +314,7 @@
   }
 
   .showing-ai-suggestions {
-    #reply-title {
+    #reply-title:not([disabled]) {
```
pmusaraj, line 317: Do we need the CSS changes now that the elements aren't rendered when they are marked as disabled? I may be misunderstanding...

awesomerobot: ooh yes, I was in a weird state locally where `showing-ai-suggestions` was present because I didn't refresh... this should already be handled without these additions

PR #1336, opened 2025-05-13
DEV: output raw response in dev mode

pmusaraj:
Makes it easier to debug issues, for example if rate limits are hit.


SamSaffron:
mixed on this, it can end up making console really noisy, I think an extra ENV var for this ... 

Maybe DISCOURSE_AI_DEBUG_LEVEL : low, medium, high?

pmusaraj:
Ok, no worries. I don't work on the plugin enough to really insist here, fine to close this if it's not helpful to the team.

PR #1342, opened 2025-05-15
FEATURE: hashtag and mention autocomplete for first bot message

SamSaffron:
Also removes controller which is a deprecated pattern


SamSaffron:
I think failure is related to https://github.com/discourse/discourse/pull/32760

dbattersby (approved):
Looks good to me, the apply hashtag and mentions follows similar patterns to how it is implemented in chat.

PR #1346, opened 2025-05-18
FEATURE: allow passing in data attributes to an artifact

SamSaffron:
Also allow artifact access to current username

Usage inside artifact is:

1. await window.discourseArtifactReady;
2. access data via window.discourseArtifactData;

```
<div class="ai-artifact" data-ai-artifact-id="482" data-hello="world"></div>
```

image


SamSaffron:
thanks, yeah leaking usernames I think is ok in this case. you need to be pretty trusted to make artifacts anyway.

PR #1347, opened 2025-05-19
DEV: rewrites artifact spec with capybara waiters

jjaffeux:
Generally speaking we never want to do:

```
expect(element.text).to eq("foo")
```

As these are rspec matchers and do not add further Capybara-style waiting specifically for the text content to become present.

### DIFF
--- a/spec/system/ai_bot/artifact_spec.rb
+++ b/spec/system/ai_bot/artifact_spec.rb
@@ -48,30 +48,26 @@ RSpec.describe "AI Artifact with Data Attributes", type: :system do
     find(".ai-artifact__click-to-run button").click
 
     artifact_element_selector = ".ai-artifact[data-ai-artifact-id='#{ai_artifact.id}']"
-    iframe_selector = "#{artifact_element_selector} iframe"
+    artifact_element = find(artifact_element_selector)
 
-    expect(page).to have_css(iframe_selector)
-
-    iframe_element = find(iframe_selector)
-    expect(iframe_element["data-custom-message"]).to eq("hello-from-post")
-    expect(iframe_element["data-post-author-id"]).to eq(author.id.to_s)
+    expect(artifact_element).to have_css("iframe[data-custom-message='hello-from-post']")
+    expect(artifact_element).to have_css("iframe[data-post-author-id='#{author.id}']")
 
     # note: artifacts are within nested iframes for security reasons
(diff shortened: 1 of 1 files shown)


spec/system/ai_bot/artifact_spec.rb
```diff
@@ -48,30 +48,26 @@
     find(".ai-artifact__click-to-run button").click
 
     artifact_element_selector = ".ai-artifact[data-ai-artifact-id='#{ai_artifact.id}']"
-    iframe_selector = "#{artifact_element_selector} iframe"
+    artifact_element = find(artifact_element_selector)
 
-    expect(page).to have_css(iframe_selector)
```
jjaffeux, line 53: We are doing a find at tnetx step so I don't think that's necessary, we will fail on the find if it doesn’t exist

spec/system/ai_bot/artifact_spec.rb
```diff
@@ -48,30 +48,26 @@
...
-
-    iframe_element = find(iframe_selector)
-    expect(iframe_element["data-custom-message"]).to eq("hello-from-post")
-    expect(iframe_element["data-post-author-id"]).to eq(author.id.to_s)
+    expect(artifact_element).to have_css("iframe[data-custom-message='hello-from-post']")
+    expect(artifact_element).to have_css("iframe[data-post-author-id='#{author.id}']")
```
jjaffeux, line 54: You can't check data attributes on the element itself in capybara, so the best way is to do this.

spec/system/ai_bot/artifact_spec.rb
```diff
@@ -48,30 +48,26 @@
...
-        data_display_element = find("#data-display")
-
-        expect(data_display_element.text).not_to be_empty
-        expect(data_display_element.text).not_to eq("Waiting for data...")
-        expect(data_display_element.text).not_to include("Error:")
+      within_frame(inner_iframe) do
+        data_selector = "#data-display"
+        expect(page).to have_selector(data_selector, text: /.+/)
+        expect(page).to have_no_selector(data_selector, text: "Waiting for data...")
+        expect(page).to have_no_selector(data_selector, text: "Error:")
```
jjaffeux, line 64: We don't want to use rspec matcher on dom elements, they are not waiting correctly

SamSaffron (approved):

PR #1350, opened 2025-05-19
DEV: cancel manager should bypass webmock

SamSaffron:
Webmock can be a bit flaky under certain use cases.

### DIFF
--- a/spec/lib/completions/cancel_manager_spec.rb
+++ b/spec/lib/completions/cancel_manager_spec.rb
@@ -3,6 +3,8 @@
 describe DiscourseAi::Completions::CancelManager do
   fab!(:model) { Fabricate(:anthropic_model, name: "test-model") }
 
+  before { WebMock.allow_net_connect! }
+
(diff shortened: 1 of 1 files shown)


spec/lib/completions/cancel_manager_spec.rb
```diff
@@ -3,6 +3,8 @@
 describe DiscourseAi::Completions::CancelManager do
   fab!(:model) { Fabricate(:anthropic_model, name: "test-model") }
 
+  before { WebMock.allow_net_connect! }
```
tgxworld, line 6: I’m not following this as it means that we are allowing outgoing http connections in the test environment. Not sure if I am missing something here.

SamSaffron: yes, read downwards, it creates a web server see the current flake, there is a bug in webmock somewhere where it defines local incorrectly.

tgxworld (approved):

PR #1358, opened 2025-05-22
FIX: stabilize diff algorithm for streaming

SamSaffron:
Previous to this fix we would diff the entire body of text, this
could lead to situations where a diff presented to a user was
wildly off matching areas in the text that should not have been
tested.

New algorithm only checks a portion of the string, this ensures
that during streaming there is no chance for wild mistakes


SamSaffron:
Agree we can expand this

PR #1359, opened 2025-05-22
FIX: ensures stream update object is scoped to its initial topic

jjaffeux:
Before this commit you could end up in this situation where a `post-updater` is constructed for a specific topic, but the user changes topic mid steam and it ends up updating the same post number but in a  different topic as we were only checking for `post_number` and not the combination of `topic_id` + `post_number`.

No test as there's only a unit test at the moment which wouldn't confirm the topic logic.

### DIFF
--- a/assets/javascripts/discourse/lib/ai-streamer/updaters/post-updater.js
+++ b/assets/javascripts/discourse/lib/ai-streamer/updaters/post-updater.js
@@ -18,7 +18,7 @@ export default class PostUpdater extends StreamUpdater {
     this.post = postStream.findLoadedPost(postId);
     const topicId = postStream.topic.id;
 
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/lib/ai-streamer/updaters/post-updater.js
```diff
@@ -57,6 +59,10 @@ export default class PostUpdater extends StreamUpdater {
   }
 
   async setCooked(value) {
+    if (!this.postElement) {
```
jjaffeux, line 62: do not attempt to find cooked on a non existing element

SamSaffron (approved):

SamSaffron:
Thanks heaps

PR #1363, opened 2025-05-23
this code should no longer need

SamSaffron:


SamSaffron:
indeed, agree we need it for slow chunks.

PR #1367, opened 2025-05-24
FIX: apply diffs more consistently

SamSaffron:
1. Do escaping direct in diff streamer, that way HTML tags and other unsafe chars can be displayed and fixed
2. Add safeguard to ensure streaming always stops when it was terminated elsewhere

### DIFF
--- a/assets/javascripts/discourse/components/modal/diff-modal.gjs
+++ b/assets/javascripts/discourse/components/modal/diff-modal.gjs
@@ -24,8 +24,8 @@ export default class ModalDiffModal extends Component {
 
   @tracked loading = false;
   @tracked finalResult = "";
-  @tracked selectedText = escapeExpression(this.args.model.selectedText);
-  @tracked diffStreamer = new DiffStreamer(this.selectedText);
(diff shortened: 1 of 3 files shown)


assets/javascripts/discourse/components/modal/diff-modal.gjs
```diff
@@ -45,11 +45,16 @@ export default class ModalDiffModal extends Component {
...
-    return this.selectedText;
+    return this.escapedSelectedText;
   }
 
   get isStreaming() {
-    return this.diffStreamer.isStreaming || this.smoothStreamer.isStreaming;
+    // diffStreamer stops "streaming" when it is finished with a chunk
+    return (
+      this.diffStreamer.isStreaming ||
+      !this.diffStreamer.isDone ||
```
jjaffeux, line 55: I feel like that when you are done you should not be streaming anymore. So setting isDone to true should also set isStreaming to false, otherwise we will plague the code with this double check.

SamSaffron:
Yeah agree probably being over protective

PR #1398, opened 2025-06-03
FEATURE: Use different personas to power AI helper features.

romanrizzi:
You can now edit each AI helper prompt individually through personas, limit access to specific groups, set different LLMs, etc.


romanrizzi:
Thanks @keegangeorge - Addressed the feature and added the deprecation message. However, this class won't be around much longer. Will drop it in two weeks unless there are any issues.

PR #1401, opened 2025-06-04
FEATURE: add context and llm controls to researcher, fix username filter

SamSaffron:
Adds context length controls to researcher (max tokens per post and batch)
Allow picking LLM for researcher
Fix bug where unicode usernames were not working
Fix documentation of OR logic

### DIFF
--- a/lib/utils/research/filter.rb
+++ b/lib/utils/research/filter.rb
@@ -153,12 +153,12 @@ module DiscourseAi
           end
         end
 
-        register_filter(/\A\@(\w+)\z/i) do |relation, username, filter|
-          user = User.find_by(username_lower: username.downcase)
-          if user
-            relation.where("posts.user_id = ?", user.id)
+        register_filter(/\Ausernames?:(.+)\z/i) do |relation, username, filter|
+          user_ids = User.where(username_lower: username.split(",").map(&:downcase)).pluck(:id)
+          if user_ids.empty?
+            relation.where("1 = 0")
           else
-            relation.where("1 = 0") # No results if user doesn't exist
+            relation.where("posts.user_id IN (?)", user_ids)
           end
         end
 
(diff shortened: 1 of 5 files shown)


spec/lib/personas/tools/researcher_spec.rb
```diff
@@ -21,6 +21,54 @@
 
   before { SiteSetting.ai_bot_enabled = true }
 
+  it "uses custom researcher_llm and applies token limits correctly" do
+    # Create a second LLM model to test the researcher_llm option
+    secondary_llm_model = Fabricate(:llm_model, name: "secondary_model")
+
+    # Create test content with long text to test token truncation
+    topic = Fabricate(:topic, category: category, tags: [tag_research])
+    long_content = "zz " * 100 # This will exceed our token limit
```
tgxworld, line 30: `100` is abit of a magic number here. I wonder if we can generate the long content here based on a constant so that updating the constant in the future doesn't break this test.

tgxworld: Actually is this limit based on https://github.com/discourse/discourse-ai/pull/1401/files#diff-a4d7be1d2a5b82a80a1a4693975f7ac231a577ced032e4d941f332292fb56339R47? If so, then I think this is fine.

lib/utils/research/filter.rb
```diff
@@ -153,12 +153,12 @@ def self.word_to_date(str)
...
-          user = User.find_by(username_lower: username.downcase)
-          if user
-            relation.where("posts.user_id = ?", user.id)
+        register_filter(/\Ausernames?:(.+)\z/i) do |relation, username, filter|
+          user_ids = User.where(username_lower: username.split(",").map(&:downcase)).pluck(:id)
+          if user_ids.empty?
+            relation.where("1 = 0")
           else
-            relation.where("1 = 0") # No results if user doesn't exist
+            relation.where("posts.user_id IN (?)", user_ids)
```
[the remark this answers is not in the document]
SamSaffron: I started off with that and it failed, so I ended up simplifying, I think the odds of getting a problem here memory wise is very low given llm is streaming in usernames very unlikely the list will get too long.

PR #1405, opened 2025-06-04
UX: add features to persona list and other style updates

awesomerobot:
This updates the persona list layout to include:

* Features used
* Default LLM
* Filterable list (by name, description, LLM, or features)
* optional card view 

Before:
image

After:
image

image

### DIFF
--- a/app/serializers/localized_ai_persona_serializer.rb
+++ b/app/serializers/localized_ai_persona_serializer.rb
@@ -56,6 +56,8 @@ class LocalizedAiPersonaSerializer < ApplicationSerializer
   end
 
   def features
-    object.features.map { |feature| { id: feature.module_id, name: feature.module_name } }
+    object.features.map do |feature|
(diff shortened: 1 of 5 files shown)


app/serializers/localized_ai_persona_serializer.rb
```diff
@@ -3,6 +3,11 @@
 class LocalizedAiPersonaSerializer < ApplicationSerializer
   root "ai_persona"
 
+  def initialize(object, options = {})
+    @features_by_persona_id = options.delete(:features_by_persona_id)
+    super(object, options)
+  end
```
SamSaffron, line 9: yeah, this is a bit of a weird pattern I can refine it today so we don't need this override.

romanrizzi:
@awesomerobot - FYI. Pushed a fix to display the feature name instead of the module:

Screenshot 2025-06-11 at 11 43 21 AM
Screenshot 2025-06-11 at 11 46 08 AM

Modules like summarization, which have different features could use the same persona.

PR #1417, opened 2025-06-09
FEATURE: persistent key-value storage for AI Artifacts

SamSaffron:
Introduces a persistent, user-scoped key-value storage system for
AI Artifacts, enabling them to be stateful and interactive. This
transforms artifacts from static content into mini-applications that can
save user input, preferences, and other data.

The core components of this feature are:

1.  **Model and API**:
    - A new `AiArtifactKeyValue` model and corresponding database table to
      store data associated with a user and an artifact.
    - A new `ArtifactKeyValuesController` provides a RESTful API for
      CRUD operations (`index`, `set`, `destroy`) on the key-value data.
    - Permissions are enforced: users can only modify their own data but
      can view public data from other users.

2.  **Secure JavaScript Bridge**:
    - A `postMessage` communication bridge is established between the
      sandboxed artifact `iframe` and the parent Discourse window.
    - A JavaScript API is exposed to the artifact as `window.discourseArtifact`
      with async methods: `get(key)`, `set(key, value, options)`,
      `delete(key)`, and `index(filter)`.
    - The parent window handles these requests, makes authenticated calls to the
      new controller, and returns the results to the iframe. This ensures
      security by keeping untrusted JS isolated.

3.  **AI Tool Integration**:
    - The `create_artifact` tool is updated with a `requires_storage`
      boolean parameter.
    - If an artifact requires storage, its metadata is flagged, and the
      system prompt for the code-generating AI is augmented with detailed
      documentation for the new storage API.

4.  **Configuration**:
    - Adds hidden site settings `ai_artifact_kv_value_max_length` and
      `ai_artifact_max_keys_per_user_per_artifact` for throttling.

This also includes a minor fix to use `jsonb_set` when updating
artifact metadata, ensuring other metadata fields are preserved.

### DIFF
--- a/ai_artifact_key_value_api_spec.rb
+++ b/ai_artifact_key_value_api_spec.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+RSpec.describe "AI Artifact Key-Value API", type: :system, js: true do
+  fab!(:user)
+  fab!(:private_message_topic) { Fabricate(:private_message_topic, user: user) }
+  fab!(:private_message_post) { Fabricate(:post, topic: private_message_topic, user: user) }
+  fab!(:artifact) do
+    Fabricate(
+      :ai_artifact,
+      post: private_message_post,
+      metadata: {
+        public: true,
+      },
+      html: '<div id="log">Artifact Loaded</div>',
+      js: <<~JS,
+        const logElement = document.getElementById('log');
+
+        window.addEventListener('load', async function() {
+          try {
+            logElement.innerHTML = "TESTING KEY-VALUE API...";
+            const log = [];
+            await window.discourseArtifact.set('test_key', 'test_value');
+            log.push('Set operation completed');
+            logElement.innerHTML = log.join('<br>');
+
+            const value = await window.discourseArtifact.get('test_key');
+            log.push('Got value:' + value);
+
+            await window.discourseArtifact.delete('test_key');
+            log.push('Delete operation completed');
+
+            const deletedValue = await window.discourseArtifact.get('test_key');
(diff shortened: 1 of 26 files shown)


ai_artifact_key_value_api_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    sign_in(user)
+  end
+
+  it "provides working key-value API in artifact JavaScript" do
+    visit "/discourse-ai/ai-bot/artifacts/#{artifact.id}"
+
+    within_frame(find("iframe")) do
+      expect(page).to have_selector("#log", wait: 10)
+      expect(page).to have_selector("#log[data-test-complete='true']", wait: 15)
+      expect(page).not_to have_selector("#log[data-test-error]")
```
jjaffeux, line 58:
```suggestion
      expect(page).to have_no_selector("#log[data-test-error]")
```

app/models/ai_artifact_key_value.rb
```diff
@@ -0,0 +1,56 @@
...
+  attribute :public, :boolean, default: false
+  validates :ai_artifact, presence: true
+  validates :user, presence: true
+  validates :key, uniqueness: { scope: %i[ai_artifact_id user_id] }
+
+  validate :validate_max_keys_per_user_per_artifact
+
+  private
+
+  def validate_max_keys_per_user_per_artifact
```
jjaffeux, line 22: we should at least test this one in a model spec

SamSaffron: will add something

xfalcox (approved):

PR #1430, opened 2025-06-12
FIX: various bugs in AI interface

SamSaffron:
1. Improves logic so when we save llm/person/tool we don't have /new routes in our state stack. Breaking expected behavior of back button
2. Fixes implementation of enum in llm tool options
3. Improves seeding logic to avoid having information hidden from end users in the UI (show examples and structured output)

### DIFF
--- a/assets/javascripts/discourse/components/ai-tool-editor-form.gjs
+++ b/assets/javascripts/discourse/components/ai-tool-editor-form.gjs
@@ -91,7 +91,7 @@ export default class AiToolEditorForm extends Component {
         this.args.tools.pushObject(this.args.model);
       }
 
-      this.router.replaceWith(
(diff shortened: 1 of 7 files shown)


assets/javascripts/discourse/components/ai-tool-editor-form.gjs
```diff
@@ -91,7 +91,7 @@ export default class AiToolEditorForm extends Component {
         this.args.tools.pushObject(this.args.model);
       }
 
-      this.router.transitionTo(
+      this.router.replaceWith(
```
tgxworld, line 94:
```suggestion
      await this.router.replaceWith(
```

tgxworld (approved):

PR #1431, opened 2025-06-12
FEATURE: hybrid artifact security mode

SamSaffron:
`In hybrid mode ai artifacts can optionally automatically run.

This is useful for cases where you may want to embed a survey and so on.

Additionally, artifacts now allow for better fidelity around display:

Eg: `<div class="ai-artifact" data-ai-artifact-id="501" data-ai-artifact-height="300px" data-ai-artifact-autorun data-ai-artifact-seamless></div>`

User can supply height and seamless mode to be seamlessly rendered with no box shadow and show full screen button.

### DIFF
--- a/assets/javascripts/discourse/components/ai-artifact.gjs
+++ b/assets/javascripts/discourse/components/ai-artifact.gjs
@@ -51,7 +51,21 @@ export default class AiArtifactComponent extends Component {
     if (this.showingArtifact) {
       return false;
     }
-    return this.siteSettings.ai_artifact_security === "strict";
+
+    if (this.siteSettings.ai_artifact_security === "strict") {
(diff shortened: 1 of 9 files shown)


assets/javascripts/discourse/components/ai-artifact.gjs
```diff
@@ -98,11 +113,38 @@ export default class AiArtifactComponent extends Component {
     }
   }
 
+  get heightStyle() {
+    if (this.args.artifactHeight) {
+      let height = parseInt(this.args.artifactHeight, 10);
```
jjaffeux, line 118: I wonder if there's any payload that can tryp parseInt here, nothing that I tried worked

PR #1434, opened 2025-06-15
DEV: makes ai menu helper a standalone menu

jjaffeux:
The current menu was rendering inside the post text toolbar (on desktop). This is not ideal as the post text toolbar rendering is conditioned on the presence of text selection, when you click a button on the toolbar, by design of the web browsers you will lose your text selection, making all of this super tricky.

This commit makes desktop and mobile behave in the same way by rendering their own menu and capturing the quote state when we render the post text selection toolbar, this allows us to reason a much simpler way about the AI helper.

This commit also removes what appears to be an unused file and corrects which was seemingly copy/paste mistakes.


assets/javascripts/discourse/lib/virtual-element-from-caret-coords.js
```diff
@@ -1,45 +0,0 @@
-class VirtualElementFromCaretCoords {
```
jjaffeux, line 1: @keegangeorge didn't find where it was used

assets/javascripts/discourse/connectors/post-text-buttons/ai-post-helper-trigger.gjs
```diff
@@ -133,65 +115,44 @@ export default class AiPostHelperTrigger extends Component {
...
-    this.highlightSelectedText();
-    if (this.site.mobileView) {
-      this.currentMenu.close();
-
-      await this.menu.show(virtualElementFromTextRange(), {
-        identifier: "ai-post-helper-menu",
-        component: AiPostHelperMenu,
-        inline: true,
-        interactive: true,
-        placement: this.shouldRenderUnder ? "bottom-start" : "top-start",
```
jjaffeux, line 145: `this.shouldRenderUnder` doesn't exist here, it's seems like it's a copy pasta error

assets/javascripts/discourse/components/ai-post-helper-menu.gjs
```diff
@@ -338,7 +320,7 @@ export default class AiPostHelperMenu extends Component {
       (and this.site.mobileView (eq this.menuState this.MENU_STATES.options))
     }}
       <div class="ai-post-helper-menu__selected-text">
-        {{@data.selectedText}}
+        {{@data.quoteState.buffer}}
```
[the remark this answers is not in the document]
jjaffeux: feel free to push directly to the PR

assets/javascripts/discourse/components/ai-post-helper-menu.gjs
```diff
@@ -168,7 +151,14 @@ export default class AiPostHelperMenu extends Component {
...
     const channel = `/discourse-ai/ai-helper/stream_suggestion/${this.args.data.quoteState.postId}`;
-    this.messageBus.subscribe(channel, this._updateResult);
+    this.messageBus.subscribe(
+      channel,
+      (data, id) => {
+        this._lastMessageIds[channel] = id;
+        this._updateResult(data, id);
+      },
+      this._lastMessageIds[channel]
+    );
```
jjaffeux, line 161: not that can't work like, you need to get this from the server

PR #1445, opened 2025-06-18
FIX: unable to scroll AI bot persona selector

keegangeorge:
## :mag: Overview
This update fixes a UX issue on the AI bot conversations page where the persona selector dropdown doesn't scroll when there are many items and the viewport is small.

No tests as it's tricky to test scrolling.

### ← Before
before

### → After
after


pmusaraj (approved):
Is this safe on mobile too?

PR #1449, opened 2025-06-20
DEV: re-implement bulk sentiment classifier

SamSaffron:
New implementation uses core concurrent job queue, it is more
robust and predictable than the one shipped in Concurrent.

Additionally:

- Trickles through updates during bulk classification
- Reports errors if we fail during a bulk classification


tgxworld (approved):
Looks good to me 👍

PR #1454, opened 2025-06-20
DEV: Update linting setup and reapply gjs-codemod

CvX:


SamSaffron:
@CvX when should we merge?

PR #1455, opened 2025-06-20
FIX: A typo in bot filtration in ai-bot-header-icon

CvX:


romanrizzi (approved):

CvX:
@romanrizzi looks like there's a spec asserting the previous behavior. shall I remove the line entirely instead?

romanrizzi:
@CvX - Thanks for fixing the typo. The spec failure was legit because we want to display the icon when there's a persona with a default LLM set. This kinda worked thanks to the typo (but will fail when trying to create a topic).

PR #1467, opened 2025-06-26
FEATURE: Update OpenAI tokenizer to GPT-4o and later one

xfalcox:


lib/tokenizer/open_ai_gpt4o_tokenizer.rb
```diff
@@ -1,13 +0,0 @@
-# frozen_string_literal: true
-
-module DiscourseAi
-  module Tokenizer
-    class OpenAiGpt4oTokenizer < OpenAiTokenizer
```
romanrizzi, line 5: Should we update any record that might be pointing to this tokenizer?

romanrizzi: Scratch that. Looks like it's not used anywhere.

romanrizzi (approved):

PR #1471, opened 2025-06-27
UX: add filter to features page, update styles

awesomerobot:
This makes the features list filterable and searchable, and does some light restyling on the feature boxes (including fixing some overflow issues)

Before:
image

After: 
image

### DIFF
--- a/assets/javascripts/discourse/components/ai-features-list.gjs
+++ b/assets/javascripts/discourse/components/ai-features-list.gjs
@@ -44,14 +44,19 @@ class ExpandableList extends Component {
     this.isExpanded = !this.isExpanded;
   }
 
+  @action
+  isLastItem(index) {
+    return index === this.visibleItems.length - 1;
+  }
+
   <template>
     {{#each this.visibleItems as |item index|}}
-      {{yield item index}}
+      {{yield item index this.isLastItem}}
     {{/each}}
 
     {{#if this.hasMore}}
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/components/ai-features-list.gjs
```diff
@@ -61,11 +66,11 @@ class ExpandableList extends Component {
...
-    return this.args.modules.sort((a, b) => {
-      const nameA = i18n(`discourse_ai.features.${a.module_name}.name`);
-      const nameB = i18n(`discourse_ai.features.${b.module_name}.name`);
-      return nameA.localeCompare(nameB);
-    });
+    if (!this.args.modules || !this.args.modules.length) {
+      return [];
+    }
+
+    return this.args.modules.sortBy("module_name");
```
SamSaffron, line 73: one thing to keep mindful of is that sortBy is going to go eventually. @CvX what is the clean and quick way of doing this in a future safe way?

SamSaffron (approved):

SamSaffron:
looks so much better!

assets/javascripts/discourse/components/ai-features-list.gjs
awesomerobot: seems like this should do it https://github.com/discourse/discourse-ai/pull/1476

PR #1472, opened 2025-06-28
Fix: convert array content to string before LLM prompt to avoid TypeE…

gptv:
### Context
When using the AI helper with Gemini (or other LLMs) to analyze posts containing images, the `content` payload is an array combining text and image metadata. This causes the following error:

### Root Cause
The `generate_prompt` method constructs `context.messages` assuming `content` is always a string. When it's an array, string operations like `.gsub` or concatenation fail.

### Fix
This PR adds a simple guard clause to normalize any array content to a string by joining its elements before passing the prompt to the LLM.

### Reproduction
- Create a post with text and an image attachment
- Run the AI helper with the default "creative" model
- Observe the TypeError
- After this patch, the prompt is processed correctly.

### DIFF
--- a/lib/ai_helper/assistant.rb
+++ b/lib/ai_helper/assistant.rb
@@ -125,6 +125,11 @@ module DiscourseAi
             custom_instructions: custom_locale_instructions(user, force_default_locale),
           )
         context = attach_user_context(context, user, force_default_locale: force_default_locale)
+        context.messages.each do |msg|
+           if msg[:content].is_a?(Array)
+             msg[:content] = msg[:content].map(&:to_s).join("\n")
+           end
+        end
 
         helper_response = +""
 


SamSaffron:
@romanrizzi thoughts?

SamSaffron:
thanks heaps for the PR!

gptv:
Thanks for the prompt feedback, @romanrizzi!

🔎 Full back-trace

TypeError (no implicit conversion of Array into String)
/plugins/discourse-ai/gems/.../tokenizer.rb:14:in `_encode'
/plugins/discourse-ai/lib/tokenizer/basic_tokenizer.rb:23:in `tokenize'
/plugins/discourse-ai/lib/personas/question_consolidator.rb:37:in `revised_prompt'
/plugins/discourse-ai/lib/personas/question_consolidator.rb:29:in `reverse_each'
/plugins/discourse-ai/lib/personas/bot.rb:58:in `reply'
/plugins/discourse-ai/app/jobs/regular/create_ai_reply.rb:18:in `execute'
…

Logged msg[:content] right before the tokenizer:

["<input>Please summarise …</input>", { upload_id: 12345 }]
# => Array

QuestionConsolidator.revised_prompt passes this Array directly to Tokenizer.encode, which expects a String, hence the error. Personas that don’t include an image hash avoid this path, so the bug is sporadic.

🌱 Fix direction
	•	Keep the Array in generate_image_caption (no join there).
	•	Add a guard only where the Array reaches revised_prompt, e.g.:

prompt = prompt.is_a?(Array) ? prompt.map(&:to_s).join("\n") : prompt

gptv:
Confirmed – the trace I captured was indeed on the **AI Bot** path (RAG + upload), not the AI Helper prompt flow. I’ve just tested with your patch in #1475 applied and the error no longer occurs. Looks like your fix fully covers the issue. 🎉

Given that, I’ll close this PR to avoid redundancy. Thanks for the clarification and the quick fix!

PR #1478, opened 2025-07-01
FIX: cross talk when in ai helper

SamSaffron:
Previous to this change we reused channels for proofreading progress and
ai helper progress

The new changeset ensures each POST to stream progress gets a dedicated
message bus channel

This fixes a class of issues where the wrong information could be displayed
to end users on subsequent proofreading or helper calls


nattsw (approved):

nattsw:
Approved with comment, but do check test passing before merging. 👌

PR #1483, opened 2025-07-02
FIX: Ignore captions and quotes when detecting locale and update prompts

nattsw:
A more deterministic way of making sure the LLM detects the correct language (instead of relying on prompt to LLM to ignore it) is to take the cooked and remove unwanted elements.

In this PR 
- we remove quotes, image captions, etc. and only take the remaining text, falling back to the unadulterated cooked
- and update prompts related to detection and translation

/152465/12

### DIFF
--- a/lib/translation/post_detection_text.rb
+++ b/lib/translation/post_detection_text.rb
@@ -8,7 +8,7 @@ module DiscourseAi
         cooked = post.cooked
         return if cooked.blank?
 
-        doc = Nokogiri::HTML.fragment(cooked)
+        doc = Nokogiri::HTML5.fragment(cooked)
         original = doc.text.strip
 
         # quotes and blockquotes
--- a/lib/translation/post_detection_text.rb
+++ b/lib/translation/post_detection_text.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module DiscourseAi
+  module Translation
+    class PostDetectionText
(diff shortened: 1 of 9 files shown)


lib/translation/post_detection_text.rb
```diff
@@ -0,0 +1,40 @@
...
+        # oneboxes (external content)
+        doc.css("aside.onebox").remove
+        # code blocks
+        doc.css("code, pre").remove
+        # hashtags
+        doc.css("a.hashtag-cooked").remove
+        # emoji
+        doc.css("img.emoji").remove
+        # mentions
+        doc.css("a.mention").remove
```
jjaffeux, line 30: I would group stuff in constants, to have only two remove calls.

I don't think the comments are adding much: `# emoji` => `img.emoji`. But if you still think they are important you can format like this:

```
NECESSARY_SELECTORS = [
  "aside.onebox", # oneboxes (external content)
  "code, pre", # code blocks
  ...
]
```

and then call: `doc.css(*NECESSARY_SELECTORS).remove`

jjaffeux (approved):

lib/translation/post_detection_text.rb
jjaffeux: I would also add that necessary/preferred is not very clear. I don't really know why they are necessary or preferred, this is where we would need a comment IMO.

PR #1485, opened 2025-07-03
FIX: correctly translate automation name

jjaffeux:
The key is `persona`, not `name`.

Before:
Screenshot 2025-07-04 at 01 11 37

After the fix:
Screenshot 2025-07-04 at 01 10 21

### DIFF
--- a/assets/javascripts/discourse/components/ai-llms-list-editor.gjs
+++ b/assets/javascripts/discourse/components/ai-llms-list-editor.gjs
@@ -112,17 +112,9 @@ export default class AiLlmsListEditor extends Component {
   }
 
   localizeUsage(usage) {
-    if (usage.type === "ai_persona") {
-      return i18n("discourse_ai.llms.usage.ai_persona", {
-        persona: usage.name,
-      });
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/ai-llms-list-editor.gjs
```diff
@@ -112,17 +112,9 @@ export default class AiLlmsListEditor extends Component {
...
-      return i18n("discourse_ai.llms.usage.ai_persona", {
-        persona: usage.name,
-      });
-    } else if (usage.type === "automation") {
-      return i18n("discourse_ai.llms.usage.automation", {
-        name: usage.name,
-      });
-    } else {
-      return i18n("discourse_ai.llms.usage." + usage.type);
-    }
```
jjaffeux, line 125: I don't think we needed all this complexity here, I get that we think name is more accurate than persona for automation, but this forces us to always have a special case for it (and created this bug, as we forgot the special case at another place). Maybe we could also think of a more generic name, like: identifier.

SamSaffron (approved):

PR #1486, opened 2025-07-04
backfill embeddings

xfalcox:
- **FEATURE: Seamless embedding model upgrades**
- **helper method to clear related topics cache**

### DIFF
--- a/app/jobs/scheduled/embeddings_backfill.rb
+++ b/app/jobs/scheduled/embeddings_backfill.rb
@@ -18,105 +18,116 @@ module Jobs
         )
       end
 
-      rebaked = 0
+      production_vector = DiscourseAi::Embeddings::Vector.instance
+      production_vector_def = production_vector.vdef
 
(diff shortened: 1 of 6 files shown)


app/jobs/scheduled/embeddings_backfill.rb
```diff
@@ -18,105 +18,116 @@ def execute(args)
...
-      # embeddings due to edits or new replies. Here we only do 10% of the limit
-      relation =
-        topics
-          .where("#{table_name}.updated_at < ?", 6.hours.ago)
-          .where("#{table_name}.updated_at < topics.updated_at")
+      topic_work_list = []
+      topic_work_list << [production_vector, production_vector_def]
+      topic_work_list << [backfill_vector, backfill_vector_def] if backfill_vector
+
+      topic_work_list.each do |vector, vector_def|
```
romanrizzi, line 37: We could use `vector#vdef` and have a one-dimensional array.

romanrizzi (approved):

PR #1489, opened 2025-07-07
DEV: Add backtrace to logs

nattsw:
Logs don't show backtrace, so this adds it.


Grubba27 (approved):

nattsw:
Maybe, maybe not. I feel it might be optimizing for something that's not worth optimizing for (they're all different models)

